Issue #26, opened 2025-01-29
Rename hoku to recall


joewagner:
Repo name is good, but there's still some spots in the code using "hoku"

PR #28, opened 2025-01-30
chore: rename hoku to recall

joewagner:
fixes: #26

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,31 +1,35 @@
-# shadcn/ui monorepo template
+# `js-recall`
 
-This template is for creating a monorepo with shadcn/ui.
+[![License](https://img.shields.io/github/license/recallnet/js-recall.svg)](./LICENSE)
+[![standard-readme compliant](https://img.shields.io/badge/standard--readme-OK-green.svg)](https://github.com/RichardLitt/standard-readme)
+
+> JS/TS monorepo for Recall
+
+## Background
+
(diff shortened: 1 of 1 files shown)


packages/sdkx/src/chains.ts
```diff
@@ -11,18 +11,18 @@ export const ignition: Chain = {
   testnet: true,
   rpcUrls: {
     [2481632]: {
-      http: ["https://evm-ignition-0.hoku.sh"],
-      webSocket: ["wss://evm-ignition-0.hoku.sh"],
+      http: ["https://evm-ignition-0.recall.network"],
```
dtbuchholz: @joewagner i think these will be `https://evm.node-0.testnet.recall.network`, `wss://evm.node-0.testnet.recall.network`, etc right? or, will the existing subdomain format also work? (for [reference](https://github.com/recallnet/docs/pull/24#discussion_r1934648330))

and cc @asutula because it doesn't look like we have DNS set up yet, so merging this (right now) would probably cause issues with the Studio build

joewagner: Good catch,  I did a global replace, but just the tld name. I was thinking (overly optimistic) that this might land before the next testnet reset, but we might as well switch to the new url here.  I can do that now.  As you mention, we just  need to make sure not to merge till after the subnet update.

PR #29, opened 2025-01-30
feat: update contracts | add build/clean scripts

dtbuchholz:
- Add `clean` script for all packages, deleting the `.turbo`, `node_modulees`, etc.
- Add `.env.example` for the Studio app 
- Add `build` script step for the `@recall/contracts` to generate abis with `@wagmi/cli`
- Alter `@recall/contracts` to use the gateway _manger_ facet, not gateway _getter_
- Alter `@recall/contracts` to include all relevant chain IDs

### DIFF
--- a/packages/contracts/package.json
+++ b/packages/contracts/package.json
@@ -5,7 +5,8 @@
   "private": true,
   "scripts": {
     "dev": "tsc --watch",
-    "build": "wagmi generate && tsc",
+    "generate": "wagmi generate && tsc",
+    "build": "tsc",
     "lint": "eslint . --max-warnings 0",
     "clean": "rm -rf .turbo node_modules dist"
   },
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -14,16 +14,17 @@ jobs:
     name: Check lint, format, and build
     runs-on: ubuntu-latest
     steps:
-      - name: Set up SSH key
-        uses: webfactory/ssh-agent@v0.9.0
-        with:
-          ssh-private-key: ${{ secrets.CONTRACTS_DEPLOY_KEY }}
+      # TMP: disable SSH & submodules since it causes problems with Vercel
+      # - name: Set up SSH key
+      #   uses: webfactory/ssh-agent@v0.9.0
+      #   with:
+      #     ssh-private-key: ${{ secrets.CONTRACTS_DEPLOY_KEY }}
 
       - name: Checkout Repository
         uses: actions/checkout@v4
-        with:
-          submodules: true
-          fetch-depth: 0
+        # with:
+        #   submodules: true
+        #   fetch-depth: 0
 
       - name: Install Node.js
         uses: actions/setup-node@v4


apps/studio/package.json
```diff
@@ -7,7 +7,8 @@
     "dev": "next dev --turbopack",
     "build": "next build",
     "start": "next start",
-    "lint": "next lint"
+    "lint": "next lint",
+    "clean": "rm -rf .turbo .next node_modules dist"
```
dtbuchholz: i added a `clean` step in each package...i was running into some issues with caching when testing things out with my sdk, so this helped

turbo.json
```diff
@@ -17,6 +17,9 @@
     "dev": {
       "cache": false,
       "persistent": true
+    },
+    "clean": {
```
dtbuchholz, line 21: i _think_ this is how you do it w/ turbo, but just wanted to double check

packages/contracts/wagmi.config.ts
```diff
@@ -2,1519 +2,143 @@ import { Config, defineConfig } from "@wagmi/cli";
...
-            name: "blockHash",
-            type: "bytes32",
-            internalType: "bytes32",
-          },
-        ],
-      },
-    ],
-    stateMutability: "view",
-  },
+const testnetSubnetChainId = 2481632;
```
dtbuchholz, line 5: i included all parent and subnet chain IDs, allowing us to use the respective contract on each chain (below)

packages/contracts/wagmi.config.ts
```diff
@@ -2,1519 +2,143 @@ import { Config, defineConfig } from "@wagmi/cli";
...
+const testnetParentChainId = 314159;
+const localnetSubnetChainId = 248163216;
+const localnetParentChainId = 31337;
+
+const gatewayManagerFacetAbi: Abi = [
```
dtbuchholz, line 10: note: different ABI—we need this for the deposit and withdrawal related actions. (I don't think we use the `gatewayGetterFacetAbi` anywhere but just a heads up)

packages/contracts/src/generated.ts
```diff
@@ -3,8 +3,8 @@
 //////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 /**
- *
- */
+
```
dtbuchholz, line 6: (note: idk why the diff is different for these generated comments)

packages/contracts/src/generated.ts
```diff
@@ -1553,1867 +1359,343 @@ export const gatewayGetterFacetParentAbi = [
...
 ] as const
 
 /**
- *
+ * -
+ * - [__View Contract on Filecoin Calibration Filscan__](https://calibration.filscan.io/address/0x63DEDA399100Dc536CD4d98FC564ea4Eaf88479F)
  */
-export const gatewayGetterFacetSubnetAddress = {
-  2481632: '0x77Aa40B105843728088c0132e43FC44348881DA8',
+export const recallErc20Address = {
```
dtbuchholz, line 1689: added the supply source / RECALL erc20

packages/contracts/package.json
```diff
@@ -4,9 +4,9 @@
   "type": "module",
   "private": true,
   "scripts": {
-    "dev": "tsc --watch",
-    "build": "tsc",
-    "lint": "eslint . --max-warnings 0"
+    "build": "wagmi generate && tsc",
```
dtbuchholz: added the `wagmi generate` step. and if the `dev` script is relevant, I can add that back in there

apps/studio/package.json
asutula: There won't be a `dist` folder to remove. Everything gets written inside of `.next`.

package.json
```diff
@@ -6,7 +6,8 @@
     "build": "turbo build",
     "dev": "turbo dev",
     "lint": "turbo lint",
-    "format": "prettier --write \"**/*.{ts,tsx,md}\""
+    "format": "prettier --write \"**/*.{ts,tsx,md}\"",
+    "clean": "turbo run clean && rm -rf node_modules"
```
asutula: I think this should be `turbo clean && ...` just for consistency with the other commands. (no `run` needed)

apps/studio/package.json
asutula: Also there shouldn't ever be a `.turbo`.

packages/address-utils/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "dev": "tsc --watch",
     "build": "tsc",
-    "lint": "eslint . --max-warnings 0"
+    "lint": "eslint . --max-warnings 0",
+    "clean": "rm -rf .turbo node_modules dist"
```
asutula, line 10: No `.turbo` needed. That only exists in the repo root is my understanding.

packages/bigint-utils/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "dev": "tsc --watch",
     "build": "tsc",
-    "lint": "eslint . --max-warnings 0"
+    "lint": "eslint . --max-warnings 0",
+    "clean": "rm -rf .turbo node_modules dist"
```
asutula, line 10: Same, no `.turbo` needed.

packages/contracts/package.json
asutula: Yea we always need a `dev` script for running dev across the whole repo. That way if you change any file in any package, it automatically get re-compiled and hot loaded into any app you're working on (like Studio).

turbo.json
asutula: Yea I don't actually understand this file yet haha. Let's try it!

apps/studio/package.json
dtbuchholz: ah right, no `dist` in `apps/studio`, but there _is_ a `.turbo`. if I run `pnpm clean`, `pnpm i`, and `pnpm build` from root, it produces an `apps/studio/.turbo` dir, and that's true for all other packages. the only file in each package's `.turbo` is `turbo-build.log`.
Screenshot 2025-01-31 at 2 16 36 PM

PR #31, opened 2025-01-31
fix: fix relative import + CI action

asutula:
Bug that was breaking compilation, plus add CI action, plus fixes to get CI passing.

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -27,6 +27,19 @@ jobs:
         with:
           version: 9
 
+      - name: Get pnpm store directory
+        shell: bash
+        run: |
+          echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_ENV
+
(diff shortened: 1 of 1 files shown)


.github/workflows/code-quality.yml
```diff
@@ -0,0 +1,50 @@
...
+    name: Code Quality
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Install Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "18"
```
dtbuchholz: Should we make this version 20 to match the package config here https://github.com/recallnet/js-recall/blob/main/package.json#L22

asutula: Yea, you read my mind. Just set it to 22 which is the version Vercel uses by default. Hoping that fixes the build error because it builds on Vercel just fine and locally as well.

PR #36, opened 2025-02-04
feat: new testnet contract addresses | fix `sdk` buckets errors

dtbuchholz:
- Updates the `contracts` package with the new testnet contract addresses; also, adds the subnet registry contract address
- Fixed `fvm` address parsing logic (it was interpreting a valid `0x00` value as invalid due to a non-explicit undefined check)
- Refactored `sdk` to pull contract addresses from the `contracts` package
- Cleaned up a few `sdk` methods to take optional params

### DIFF
--- a/packages/sdk/tsconfig.json
+++ b/packages/sdk/tsconfig.json
@@ -1,5 +1,5 @@
 {
-  "extends": "@recall/typescript-config/react-library.json",
+  "extends": "@recall/typescript-config/base.json",
   "compilerOptions": {
     "outDir": "dist"
   },


packages/sdk/test/contracts.test.ts
```diff
@@ -21,11 +21,16 @@ import {
   CreditManager,
 } from "../src/entities/index.js";
 
-// TODO: once https://github.com/recallnet/contracts/pull/56 is merged, we can remove this.
-// Currently, `ipc` localnet deploys a credit and blob manager contract that returns a different
-// value in `getAccount`, `getCreditBalance`, and `getStorageUsage` than what this JS lib expects.
-const CREDIT_MANAGER_ADDRESS = "";
+// TODO: these tests only work with `ipc` commit `2cf4f92` or earlier due to a bug that makes it
+// impossible to add objects via the bucket manager wrapper contract:
+// https://github.com/recallnet/ipc/issues/510
```
dtbuchholz, line 26: of note: if you try to use the localnet with the wrapper contracts, make sure you use `ipc` commit `2cf4f92` or earlier. (this matches the testnet's fendermint image sha, so you shouldn't run into any issues with the wrappers on testnet.)

there's a breaking bug that prevents you from adding objects in the wrappers, but ~everything else should work. it happened in this commit https://github.com/recallnet/ipc/commit/e622dead184297bc897db4a90c187c79bc394e95 and is being tracked here: https://github.com/recallnet/ipc/issues/510

packages/sdk/tsconfig.json
```diff
@@ -1,9 +1,8 @@
 {
-  "extends": "@recall/typescript-config/base.json",
+  "extends": "@recall/typescript-config/react-library.json",
```
asutula: Since this isn't a react library, I would leave this the way it was.

dtbuchholz: oops, idk why i changed that. good catch.

PR #42, opened 2025-02-06
feat: sdk web compatibiltiy w/ direct object uploads

dtbuchholz:
Removes iroh as a dependency and uses the direct upload feature, enabling web compatibility

### DIFF
--- a/packages/sdk/src/provider/utils.ts
+++ b/packages/sdk/src/provider/utils.ts
@@ -70,35 +70,14 @@ export interface FileHandler {
   }>;
 }
 
-// TODO: figure out if this is the right pattern for file handling for web vs nodejs
-export const nodeFileHandler: FileHandler = {
-  async readFile(input) {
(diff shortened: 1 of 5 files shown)


packages/sdk/src/provider/utils.ts
```diff
@@ -70,48 +70,53 @@ export interface FileHandler {
   }>;
 }
 
-// TODO: figure out if this is the right pattern for file handling for web vs nodejs
-export const nodeFileHandler: FileHandler = {
-  async readFile(input) {
+export const createFileHandler = (): FileHandler => ({
```
dtbuchholz, line 73: @asutula i'm open to ideas on a multipurpose file handler (both nodejs and web envs). this works...but idk if the logic is a best practice

asutula: Cool, yea I need to learn more about all that, so this is great, at least for now.

Issue #43, opened 2025-02-06
Publish packages on npm


dtbuchholz:
Note: we might want to semver everything with `...-pre.0` or similar

dtbuchholz:
done https://www.npmjs.com/package/@recallnet/sdk

PR #48, opened 2025-02-12
chore: publish packages

dtbuchholz:
- Adds esm and cjs distribution build with `tsup`
- Adds [`changelog`](https://github.com/changesets/changesets/blob/main/docs/prereleases.md) tool for versioning, publishing, and creating automated markdown describing changes
- Adds a [GH action to do the package publishing](https://github.com/changesets/action) (but we need an npm key set up for this repo). (Optionally, we might want to add this bot, which comments on PRs: [here](https://github.com/apps/changeset-bot))

Note: this follows the instructions for the turbo monorepo: [here](https://turbo.build/repo/docs/guides/publishing-libraries). Closes https://github.com/recallnet/js-recall/issues/43.

### DIFF
--- a/packages/fvm/package.json
+++ b/packages/fvm/package.json
@@ -59,6 +59,7 @@
     "chai-as-promised": "^8.0.1",
     "mocha": "^11.0.1",
     "ts-node": "^10.9.2",
+    "tsup": "^8.3.6",
     "typescript": "^5.7.3"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -258,6 +258,9 @@ importers:
       ts-node:
         specifier: ^10.9.2
         version: 10.9.2(@types/node@22.12.0)(typescript@5.7.3)
+      tsup:
+        specifier: ^8.3.6
+        version: 8.3.6(jiti@1.21.6)(postcss@8.4.49)(typescript@5.7.3)(yaml@2.6.1)
       typescript:
         specifier: ^5.7.3
         version: 5.7.3


packages/ui/src/recall/wallet.tsx
```diff
@@ -111,7 +111,17 @@ export const Wallet = ({ className, ...props }: Props) => {
       </Dialog>
       <BuyCreditsDialog open={buyCreditsOpen} setOpen={setBuyCreditsOpen} />
       <ConnectButton.Custom>
-        {({ account, chain, openConnectModal, mounted }) => {
+        {/* eslint-disable @typescript-eslint/ban-ts-comment */}
```
dtbuchholz, line 114: note: not sure why, but i couldn't build w/o doing this, and i couldnt figure out the right type (w/o spending too much time on it)

packages/ui/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# @recall/ui
```
dtbuchholz: changelog gets autocreated. i basically ran:
- `pnpm changeset pre enter rc` (to set a prerelease versoin)
- `pnpm changeset` (to pick which packages need a release for major/minor/patch)
- `pnpm changeset version` (to create the version for the release)

.github/workflows/release.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: pnpm install
+
+      - name: Create Release Pull Request or Publish to npm
+        id: changesets
+        uses: changesets/action@v1
+        with:
+          publish: pnpm publish-packages
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
dtbuchholz, line 45: we can run `pnpm publish-packages` to trigger this manually, or, i set up this GH action to do it on merges to main....we just need to set up these tokens

asutula: Sweet

apps/studio/next.config.mjs
```diff
@@ -1,6 +1,10 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   transpilePackages: ["@recall/ui"],
+   webpack: (config) => {
+    config.externals = [...(config.externals || []), "pino-pretty"];
+    return config;
+  },
```
asutula, line 7: What's this?

packages/address-utils/package.json
```diff
@@ -14,11 +32,5 @@
     "@recall/typescript-config": "workspace:*",
     "typescript": "^5.7.3"
   },
-  "dependencies": {},
-  "exports": {
-    "./display": {
-      "types": "./dist/display.d.ts",
-      "default": "./dist/display.js"
-    }
-  }
```
asutula: Why was this removed? I think this is the way to define exactly how a package exports different files.

packages/address-utils/package.json
```diff
@@ -1,11 +1,29 @@
...
+    }
+  },
+  "files": [
+    "dist"
+  ],
   "scripts": {
-    "dev": "tsc --watch",
-    "build": "tsc",
+    "dev": "tsup --watch",
+    "build": "tsup",
```
asutula, line 19: `tsup` looks cool. I don't know much about it, would like to understand why the change. Also, shouldn't `tsup` be a dev dependency then?

packages/address-utils/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./display.js";
```
asutula: This type of single export is what we're trying to avoid by using `exports` in `package.json`.

packages/address-utils/tsup.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from "tsup";
+
+export default defineConfig({
+  entry: ["src/index.ts", "src/display.ts"],
```
asutula: Seems wrong to have two entrypoints to the same code.

asutula: Also seems like tsup should get this information from package.json exports.

packages/bigint-utils/package.json
```diff
@@ -1,11 +1,39 @@
...
+    "./conversions": {
+      "types": "./dist/conversions.d.ts",
+      "import": "./dist/conversions.mjs",
+      "require": "./dist/conversions.cjs"
+    },
+    "./format-atto-rcl": {
+      "types": "./dist/format-atto-rcl.d.ts",
+      "import": "./dist/format-atto-rcl.mjs",
+      "require": "./dist/format-atto-rcl.cjs"
+    }
```
asutula, line 22: Ok cool you left the exports in this package. But I still think we shouldn't have a top level index.ts that re-exports everything.

packages/bigint-utils/src/index.ts
asutula, line 1: Shouldn't need this file.

packages/typescript-config/package.json
asutula, line 1: Just thinking here... Is it necessary to publish packages that are only used as dev dependencies? I don't think it should be needed. If I'm installing a cli or sdk package, I won't use any of the dev dependencies of those packages. If I'm going to contribute to the js-recall repo, I'll have the repo checked out and will resolve the dev dependencies is the pnpm workspace (monorepo).

packages/ui/src/recall/wallet.tsx
asutula: Yea all good. It's some weird thing with the rainbow kit types. For some reason, they types aren't being resolved here. Not sure why, but this works for now.

apps/studio/next.config.mjs
dtbuchholz: @asutula i always get warnings complaining about this package during the build step. you can see an example [here](https://github.com/recallnet/js-recall/actions/runs/13278515530/job/37072359653#step:11:82). with this change, the error went away [here](https://github.com/recallnet/js-recall/actions/runs/13292474390/job/37116381429?pr=48#step:11:213)

packages/address-utils/package.json
dtbuchholz: @asutula i struggled for quite some time to wire it all up and build successfully, specifically, for the studio/portal build step. i couldnt figure out why it would fail, but this seemed to fix it.

it's possible there were some old `dist` files that were causing issues, not sure. i can test again to be sure tho.

packages/address-utils/package.json
dtbuchholz: i thought the same thing, but after some back/forth with cursor, it convinced me otherwise since dev dependencies are hoisted / shared across packages.

packages/address-utils/src/index.ts
dtbuchholz: yeh same wrt having issue in the studio build step. i'll test it out again

packages/typescript-config/package.json
dtbuchholz: @asutula i'm not sure. the reason i included it was because i tried ignoring the config files in the changeset configuration, and it would not let me.

```
🦋  error The package "@recall/bigint-utils" depends on the ignored package "@recall/eslint-config", but "@recall/bigint-utils" is not being ignored. Please add "@recall/bigint-utils" to the `ignore` option.
```
but, changesets is supposed to respect the package field `"private": true` (see [here](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#access-restricted--public)), so i could add that back in there. i think that would stop it from being published, and you also "select" the packages you want to add to the changeset whenever you run the command.

however, the other reason i included it was because it'd be nice to reuse the same eslint/tsconfigs elsewhere—like in the docs site, or demo repos. otherwise, i end up copying/pasting what you set up.

dtbuchholz: also—we can better tailor this all to our needs over time. some examples of other packages that use it are [here](https://github.com/changesets/changesets?tab=readme-ov-file#cool-projects-already-using-changesets-for-versioning-and-changelogs), including wagmi and pnpm

dtbuchholz: maybe to spell it out a bit more explicitly—we currently are publishing everything except the `studio`, `ui`, and `sdkx` because these all are set to `"private": true` in their `package.json` files. 

if package A depend on package B, and package A needs to be published, then package B needs to be included in the changeset config. but, package B can still be set to `"private": true` and _not_ published.

packages/address-utils/tsup.config.ts
dtbuchholz: if you dont explicitly define multiple entrypoints, then tsup wont include them in `dist`.

i wanted to keep the code compatible with what you already had in the studio. e.g., if we dont define `"src/display.ts"`, then you cant import `@recall/address-utils/display`.

apps/studio/next.config.mjs
asutula: Strange. The repo was certainly building with no problems many times without this change. I imagine something else is causing some weirdness and the error message is a red herring. Can leave it for now and look into it later.

packages/address-utils/package.json
asutula: Turborepo says to install packages where they are used, we should stick to that. I also like it because it makes leaves no mysteries about where a package is coming from. https://turbo.build/repo/docs/crafting-your-repository/managing-dependencies#install-dependencies-where-theyre-used

packages/address-utils/tsup.config.ts
asutula: I googled and found others asking about with with tsup. https://github.com/egoist/tsup/discussions/732 for example. Maybe it's not yet possible, so fine to duplicate that information in package.json and tsup.config.ts for now.

packages/typescript-config/package.json
asutula: Yea maybe keeping internal dev packages as `"private": true` is the way to go. If that works, we should do that now., just to avoid publish a bunch of packages that don't need to be out there and causing confusion in the future. If that is too much of a pain for some reason, good to know we can adjust this over time to suit our needs.

packages/sdk/package.json
```diff
@@ -1,55 +1,87 @@
...
-      "default": "./dist/index.js"
+    "./chains": {
+      "types": "./dist/chains.d.ts",
+      "import": "./dist/chains.mjs",
+      "require": "./dist/chains.cjs"
+    },
+    "./client": {
+      "types": "./dist/client.d.ts",
+      "import": "./dist/client.mjs",
+      "require": "./dist/client.cjs"
```
dtbuchholz, line 25: @mzkrasner note: i removed the single entry point barrel file from the `sdk` package, and we renamed the org. 

instead, you'll import like `@recallnet/sdk/client` or `@recallnet/sdk/chains`

PR #49, opened 2025-02-14
Rename to `@recallnet`

dtbuchholz:
- Renames `@recall` to `@recallnet`
- Fixes release workflow by adding the wallet connect env var

### DIFF
--- a/apps/studio/app/buckets/_components/add-object-dialog.tsx
+++ b/apps/studio/app/buckets/_components/add-object-dialog.tsx
@@ -7,15 +7,15 @@ import { useForm } from "react-hook-form";
 import { Address } from "viem";
 import { z } from "zod";
 
-import { useAddFile } from "@recall/sdkx/react/buckets";
-import { Button } from "@recall/ui/components/button";
+import { useAddFile } from "@recallnet/sdkx/react/buckets";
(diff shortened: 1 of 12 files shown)


.changeset/pre.json
```diff
@@ -0,0 +1,17 @@
+{
+  "mode": "pre",
+  "tag": "rc",
+  "initialVersions": {
+    "studio": "0.0.0",
+    "@recallnet/address-utils": "0.0.0",
```
dtbuchholz: afaik—this just snapshots all of the package versions, even if they're ignored in the `.changeset/config.json`. it only happens if you run `pnpm changset pre enter <name>`. whenever we want to switch to "normal" semver, we run `pnpm changset pre exit`.

if we dont want to use the prerelease mode, that's fine too.

.changeset/pretty-walls-sip.md
```diff
@@ -0,0 +1,7 @@
+---
+"@recallnet/contracts": patch
+"@recallnet/fvm": patch
+"@recallnet/sdk": patch
```
dtbuchholz: when you run `pnpm changeset`, you pick each package you want to include in the release. then, you pick if you want a major, minor, or patch release. i chose patch for all three.

after you select them, you create a short summary.

PR #52, opened 2025-02-17
Constants and chains packages

asutula:
This PR adds two new workspace packages, `network-constants` which lists all things like urls and contract addresses, etc, and `chains` which defines all the wagmi chains and provides helper functions to work with them. These were mostly taken from existing sdk code. This is useful because we can use these packages now without depending on the entire sdk and there is a single place now to update all things when a network upgrade or reset happens.

Once this is merged, it will be nice and easy to make the portal app configurable for which chain to run on.


packages/contracts/wagmi.config.ts
asutula, line 1: No longer need to make updates in here, it's all done in the `network-constants` package.

packages/network-constants/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+export const TESTNET_BUCKET_MANAGER_ADDRESS =
+  "0xfbCF213040240BA86Fed92961BB60625233641a1";
+export const LOCALNET_BUCKET_MANAGER_ADDRESS =
+  "0xf7Cd8fa9b94DB2Aa972023b379c7f72c65E4De9D";
+
+// CreditManager contract addresses
+export const TESTNET_CREDIT_MANAGER_ADDRESS =
+  "0xDB85431B6a016e1652c7E898918d787B6aef7185";
+export const LOCALNET_CREDIT_MANAGER_ADDRESS =
+  "0x82C6D3ed4cD33d8EC1E51d0B5Cc1d822Eaa0c3dC";
```
asutula, line 96: Moved all the contract addresses in here.

packages/network-constants/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+// Objects API URLs
+export const TESTNET_OBJECT_API_URL =
+  "https://objects.node-0.testnet.recall.network";
+export const LOCALNET_OBJECT_API_URL = "http://127.0.0.1:8001";
+export const DEVNET_OBJECT_API_URL = "http://127.0.0.1:8001";
+
+// Explorer URLs
+export const TESTNET_EXPLORER_URL = "https://explorer.testnet.recall.network";
+export const LOCALNET_EXPLORER_URL = "http://127.0.0.1:8000";
+export const DEVNET_EXPLORER_URL = "http://127.0.0.1:8000";
```
asutula, line 40: Added explorer urls. Not sure what the localnet and devnet addresses should be, need to check and update.

packages/network-constants/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+
+// Faucet URLs
+export const TESTNET_FAUCET_URL = "https://faucet.recall.network";
```
asutula, line 43: Added the faucet URL as well.

packages/sdk/src/client.ts
```diff
@@ -15,7 +15,8 @@ import {
 import { privateKeyToAccount } from "viem/accounts";
 import "viem/window";
 
-import { type ChainName, getChain, testnet } from "./chains.js";
+import { type ChainName, getChain, testnet } from "@recallnet/chains";
+
```
asutula, line 19: SDK now relies on the chains package.

packages/sdk/src/entities/blob.ts
```diff
@@ -14,9 +14,9 @@ import {
 } from "viem";
 
 import { blobManagerAbi, blobManagerAddress } from "@recallnet/contracts";
+import { MIN_TTL } from "@recallnet/network-constants";
```
asutula, line 17: It also depends on the network constants package.

apps/studio/lib/wagmi-config.ts
```diff
@@ -1,13 +1,13 @@
 import { getDefaultConfig } from "@rainbow-me/rainbowkit";
 import { http } from "wagmi";
 
-import { ignition } from "@recallnet/sdkx/chains";
+import { testnet } from "@recallnet/chains";
```
asutula, line 4: Studio app now depends on the shared chains package.

PR #54, opened 2025-02-17
Make chain configureable for studio

asutula:
You can set an env var to run on testnet, localnet, devnet, etc. Required a few supporting changes in other packages.

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -8,6 +8,7 @@ on:
 
 env:
   NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID: fe319abf1a1b0d76b39894e5e17aff0c
+  NEXT_PUBLIC_CHAIN_NAME: testnet
 
 jobs:
   quality:


apps/studio/next.config.mjs
```diff
@@ -1,10 +1,6 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   transpilePackages: ["@recallnet/ui"],
-  webpack: (config) => {
-    config.externals = [...(config.externals || []), "pino-pretty"];
-    return config;
-  },
```
asutula, line 7: NextJS was giving errors about how "webpack was enabled and turbopack was not, which may cause problems". Basically the version of NextJS we use expects to use Turbopack and this config change forced it to use Webpack. I reverted this and get only a warning about pino-pretty, which I think we can live with for now. Seems better than deviating from what NextJS expects.

packages/chains/src/index.ts
```diff
@@ -127,3 +131,25 @@ export function checkChainName(chainName: ChainName): boolean {
...
+}
+
+export function getRegistrarUrl(chain: Chain): string {
+  switch (chain.id) {
+    case testnet.id:
+      return TESTNET_REGISTRAR_URL;
+    default:
+      throw new Error(`Registrar URL not found for chain ${chain.name}`);
+  }
+}
```
asutula, line 155: Added these helpers to the chain package to return some useful urls. This kinda replaces the type of functionality that the `Network` class in the SDK provides, but is arguably more simple. Not sure which is the better approach.

dtbuchholz: nice, i think that works...and with the other `chains` methods for getting parent chain info, i should be able to fully migrate the `sdk` to use this.

Issue #51, opened 2025-02-17
Refactor wrapper methods with new `from` param


dtbuchholz:
See https://github.com/recallnet/ipc/pull/524 / https://github.com/recallnet/contracts/issues/80 for details

PR #55, opened 2025-02-17
feat: updates for testnet v0.1.3

dtbuchholz:
## Summary

Adds the new required `from` field in bucket methods for adding or deleting objects via https://github.com/recallnet/contracts/pull/81. Also updates contracts and RPCs for the latest testnet. Closes https://github.com/recallnet/js-recall/issues/51.

## Details
- Updates the `contracts` package with the latest changes, which alter the ABI for `addObject` and `deleteObject` and add a new `IMachineFacade` abi.
- Updates `network-constants` with new contracts and RPCs
- Includes the new `from` param in `sdk`, `sdkx`, and `studio`
- Removes events for wrappers, and only uses `IMachineFacade` abi
- Fix tests for `fvm`, `sdk`


packages/contracts/src/index.ts
```diff
@@ -866,6 +892,26 @@ export const bucketManagerAbi = [
...
+      },
+      {
+        name: "bucket",
+        internalType: "address",
+        type: "address",
+        indexed: true,
+      },
+      { name: "key", internalType: "string", type: "string", indexed: false },
+    ],
+    name: "UpdateObjectMetadata",
```
dtbuchholz: new method for updating an existing object's metadata

packages/sdk/src/entities/bucket.ts
```diff
@@ -383,6 +384,7 @@ export class BucketManager {
       size,
       contentType,
     );
+    const from = this.client.walletClient.account.address;
```
dtbuchholz, line 367: in the `sdk` package: when adding an object—instead of taking `from` as a param, i decided to just assume the it's the signer's address. idk if that's ideal, or we should take it as a param. i change that if needed.

packages/sdkx/src/react/buckets.ts
```diff
@@ -139,6 +139,7 @@ export function useCreateBucket() {
 
 export interface AddFileArgs {
   bucket: Address;
+  from: Address;
```
dtbuchholz, line 142: a new `from` field in the `sdkx`'s args for adding a file. the portal uses the `useAccount` hook to get the connected wallet's address and passes it as this value.

packages/sdkx/src/react/buckets.ts
```diff
@@ -250,19 +252,19 @@ export function useDeleteObject() {
   );
 
   const deleteObject = useCallback(
-    (bucket: Address, key: string) =>
+    (bucket: Address, from: Address, key: string) =>
       writeContract({
         ...baseConfig,
-        args: [bucket, key],
+        args: [bucket, key, from],
```
dtbuchholz, line 258: note: i realized the parameter order in the actual solidity contract is a bit weird because it's more intuitive to pass `from` before you pass `key`—or even have `from` as the first param instead of `bucket`

packages/sdkx/src/react/buckets.ts
```diff
@@ -283,8 +285,8 @@ async function uploadFile(variables: {
   onProgress?: (progress: number) => void;
 }) {
   const f = new FormData();
-  f.append("data", variables.file);
   f.append("size", variables.file.size.toString());
+  f.append("data", variables.file);
```
dtbuchholz, line 289: note: i ran into this issue with the `sdk` package too. if you _dont_ put `size` first, the api will error with `missing size in form`. not sure why or when that changed.

asutula (approved):

PR #58, opened 2025-02-18
chore: release workflow changes

dtbuchholz:
- We use `version.yml` to auto-create version PRs whenever the `.changesets` directory is updated upon merging to main.
   - This only happens if you stage a new release via the `pnpm changeset` command locally, which will create markdown file(s) in `.changesets` that describe the semver change and relevant packages.
   - When you merge these `.changesets` markdown files to main, a PR is opened, which deletes these changeset files, creates/updates the `CHANGELOG.md` files with their contents, and updates each package's `package.json` file's semver accordingly.
   - If an existing version PR is open, and you merge more staged changesets into main, it'll automatically keep updating the PR.
   - An example of what this looks like is https://github.com/recallnet/js-recall/pull/57. 
- Upon merging the PR from `version.yml`, the `publish.yml` workflow _should_ auto-trigger since it only runs when changes to `CHANGELOG.md` occur.
   - This will publish directly to npm. It [assumes the previous commit](https://github.com/changesets/changesets/blob/main/packages/cli/README.md#publish) is the output from `changeset version`.
   - I'm assuming this should work—the PR linked below only tested out the versioning step, and the only way to really test the publishing is to "do it in prod"
- Optionally, we can [add this bot](https://github.com/apps/changeset-bot) to comment on PRs to remind you to run the `pnpm changeset` command. This is _probably_ useful, but also might get annoying.

See [this PR](https://github.com/recallnet/js-recall/pull/56#issuecomment-2664495538) for details on how some of this works in practice, and the [changeset docs](https://github.com/changesets/changesets/blob/main/docs/automating-changesets.md#how-do-i-run-the-version-and-publish-commands) have more information. There are probably more optimal ways to do this, too, so i'm open to ideas. Alternatively, we could handle all of this manually/locally.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,53 @@
+# Publishes packages to npm *directly after* merging a version PR via `version.yml`
+name: Publish
+
+on:
+  push:
+    paths:
+      - "**/CHANGELOG.md" # Only run when changelog is updated
+    branches:
+      - main
+
+env:
+  NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID: fe319abf1a1b0d76b39894e5e17aff0c
+  NEXT_PUBLIC_CHAIN_NAME: testnet
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  publish:
+    name: Publish
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+
+      - name: Install Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@v2
+        with:
+          version: 9
+
+      - name: Install Dependencies
+        run: pnpm install
+
+      - name: Build
+        run: pnpm build
+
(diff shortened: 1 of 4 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,53 @@
...
+      # See the automated release feature for more details: https://github.com/changesets/action#with-publishing
+      - name: Publish to npm
+        id: changesets
+        uses: changesets/action@v1
+        with:
+          publish: pnpm publish-packages
+          commit: "chore: publish packages"
+          title: "chore: publish packages"
+          setupGitUser: true
+          createGithubReleases: ${{ github.ref == 'refs/heads/main' }}
```
dtbuchholz, line 50: i tried to base some of this off of wagmi's flow, but theirs is way more complicated: https://github.com/wevm/wagmi/blob/main/.github/workflows/changesets.yml

for ref: [chakra](https://github.com/chakra-ui/chakra-ui/blob/main/.github/workflows/release.yml), [astro](https://github.com/withastro/astro/blob/main/.github/workflows/release.yml), and [some others](https://github.com/changesets/changesets/tree/main?tab=readme-ov-file#cool-projects-already-using-changesets-for-versioning-and-changelogs)

asutula (approved):
Very cool stuff, nice work!

PR #61, opened 2025-02-19
chore: fix release flow, 'version' cmd

dtbuchholz:
1. The root `pnpm version` command was (incorrectly) being used...since that will use the default pnpm command, and not our custom one. This has been renamed to `pnpm version-packages`
2. We simplify the release workflow into a single action, based on [Cloudflare's](https://github.com/cloudflare/workers-sdk/blob/main/.github/workflows/changesets.yml) and [ShadCN's](https://github.com/shadcn-ui/ui/blob/main/.github/workflows/release.yml) pattern. 
   - We had split a single `release.yml` into two separate steps because we observed packages getting published unexpectedly. But, it doesn't look like other orgs do this, and everyone seems to use the single release workflow.
   - Both of these noted an important gap with `changesets`, which is that it doesn't update the lockfile after a version changes, so we've added a script to make sure the `changeset version` is followed by updating the lockfile (see reference [here](https://github.com/cloudflare/workers-sdk/blob/main/.github/changeset-version.js) / [here](https://github.com/shadcn-ui/ui/blob/main/.github/changeset-version.js))


mzkrasner (commented):
If execSync fails, I believe it will cause an immediate failure without helpful error messages

dtbuchholz:
> If execSync fails, I believe it will cause an immediate failure without helpful error messages

@mzkrasner true, kk i added a try-catch setup

mzkrasner (approved):
nice - looks good

PR #76, opened 2025-02-24
enable vercel analytics in studio

asutula:


dtbuchholz (approved):
lgtm. also, if we want to enable Google analytics, here's what I set up in the docs. we should be able to use the same google ID https://github.com/recallnet/docs/blob/main/app/layout.tsx#L59

asutula:
> lgtm. also, if we want to enable Google analytics, here's what I set up in the docs. we should be able to use the same google ID https://github.com/recallnet/docs/blob/main/app/layout.tsx#L59

Cool. Not sure what the benefit is for Vercel vs. Google, but we should consider both I suppose.

dtbuchholz:
yeah idk the differences but I know the marketing team looks at the GA stuff

Issue #79, opened 2025-02-24
Refactor SDK `entities` to allow for similar instantiation as `RecallClient`


dtbuchholz:
The `entities` (buckets, blobs, etc) currently take a `RecallClient` in their constructor. But, `RecallClient` instantiates each entity, which means the entity itself is sort of self-referencing. It's not an ideal pattern and should, instead, have some per-entity config that nicely picks from the global `RecallClient`'s config.

Issue #77, opened 2025-02-24
Add docstrings to public packages


dtbuchholz:
Primarily the `sdk`, but also `contracts`, `chains`, and `network-constants`

PR #80, opened 2025-02-24
feat: remove `sdk` network, add more exports

dtbuchholz:
- Removes the `network.ts`, which was legacy code that tried to emulate `rust-recall` patterns
- Removes `entities/index.ts` barrel file, move `utils` and `errors` into root, and export them so they're usable elsewhere
- Moves the objects `provider.ts` into root
- Fixes contract override logic
- Alters the `GatewayManager` pattern to take a public client, wallet client, and contract overrides—instead of taking the `RecallClient` that posses these values. This required `AccountManager` to go through some refactoring to account for this.


packages/sdk/src/client.ts
```diff
@@ -34,19 +30,6 @@ export const createPublicClientForChain: (
     transport: http(),
   });
 
-// Creates a wallet client for the given chain with a browser wallet provider
-export const walletClientFromBrowser = (
```
dtbuchholz, line 38: i realized web-based wallets dont really work with the `sdk` wrt executing transactions, so i removed this. 

we could either choose to try and make web-based wallet flows work in the `sdk`, or just (eventually) publish the `sdkx` package—and i think the latter was already the plan. maybe we call `sdkx` the `@recallnet/web` package or something.

packages/sdk/src/client.ts
```diff
@@ -84,7 +82,8 @@ export class RecallClient {
     }
     const chain = this.publicClient.chain;
     if (!chain) throw new Error("missing chain in provided client");
-    this.network = config.network ?? Network.fromChain(chain);
+    this.subnetId = SubnetId.fromChain(chain);
```
dtbuchholz, line 85: need for cross-chain flows like deposits and withdrawals. (this used to be part of the `Network`, which we've removed.)

packages/sdk/src/entities/account.ts
```diff
@@ -61,22 +68,28 @@ export class AccountManager {
...
-    if (!deployedSupplySourceAddress) {
+    const overrideConfig =
+      this.client.contractOverrides.accountManager?.recallErc20;
+    const overrideSupplySourceAddress = overrideConfig
+      ? overrideConfig[chainId]
+      : undefined;
+    const override =
+      contractAddress ??
+      overrideSupplySourceAddress ??
+      deployedSupplySourceAddress;
```
dtbuchholz, line 86: note: this pattern appears in a couple of places. i didn't want to overhaul the `AccountManager` etc too much, but it's a little messy imo. i'll take another pass in a different PR on refactoring a few other pieces in #79

packages/sdk/src/entities/account.ts
```diff
@@ -119,36 +132,31 @@ export class AccountManager {
...
     ).read.balanceOf(args);
     await reset();
     return {
       result: { address: addr, nonce, balance: balance.result, parentBalance },
     };
   }
 
-  // Approve a spender to transfer funds from the account
-  async approve(spender: Address, amount: bigint): Promise<Result> {
+  // Approve a spender to transfer funds from the account. Assumes the wallet client is on the
```
dtbuchholz, line 145: and note: i'll also handle adding docstrings to our public packages in #77

packages/sdk/src/ipc/gateway.ts
```diff
@@ -74,55 +69,50 @@ function releaseParamsToTyped(address: Address): ReleaseParams {
 // Gateway manager facet for managing gateway funding and releasing
 export class GatewayManager {
   getContract(
-    client: RecallClient,
-    contractAddress?: Address,
+    publicClient: PublicClient,
+    walletClient: WalletClient,
+    contractAddress: Address,
```
dtbuchholz, line 74: Note: this pattern is probably needed in the `entities` (buckets, blobs, etc) since `RecallClient` instantiates each entity, but then the entity itself takes a `RecallClient` in its constructor. It's self-referencing and not ideal, but I wanted to keep the PR small and can handle that later. will handle in #79

packages/sdk/src/utils.ts
```diff
@@ -120,3 +132,65 @@ export const createFileHandler = (): FileHandler => ({
     };
   },
 });
+
```
dtbuchholz, line 135: these arent "new"—we simply merged a couple of `utils` files into a single one

dtbuchholz:
@asutula i started to refactor some pieces from the `sdk` package (backward compatible). i started with removing that weird `network.ts` logic which tried to recreate `rust-recall` patterns, and that led to some cascading changes with contract overrides and some subnet-related definitions.

i didnt want this PR to get too complicated, so i plan to do another refactor pass in the future. the tests all pass locally, so this all works for now. 

once we merge this, i can merge https://github.com/recallnet/js-recall/pull/75, which will create a new release that accounts for the RPC changes in https://github.com/recallnet/js-recall/commit/a989499b498f059598f65f38c6c3a09a3769e004 as well as these SDK changes

packages/sdk/src/client.ts
asutula: Yea sounds like a good plan to eventually publish the react sdk.

PR #81, opened 2025-02-24
Import the faucet app into the monorepo

asutula:
Also updated the faucet code to read a chain name from the env, then look up the chain, registrar url, and explorer url from the `chains` package.


dtbuchholz (approved):
🤘

PR #84, opened 2025-02-25
feat: update public RPCs

dtbuchholz:


dtbuchholz:
note: the only remaining API that directly hits our nodes is [the faucet URL](https://github.com/recallnet/js-recall/pull/84/files#diff-1a42ef25a49c1dc57276387d2c64de2c98e2f751a120bb805216be559b68a836R38). i dont think there's a way around doing this due to how the faucet works—just calling it out.

dtbuchholz:
@asutula is there anything else we want to include in a release? considering we want ppl to use our public RPCs asap, i'll publish a new release after i merge this.

asutula (commented):
This looks good, but I'm a bit surprised about the URLs considering this is a testnet. What will the chain.love URLs look like for mainnet? Maybe I'm missing something about how they differentiate between testnet and mainnet?

asutula:
> note: the only remaining API that directly hits our nodes is [the faucet URL](https://github.com/recallnet/js-recall/pull/84/files#diff-1a42ef25a49c1dc57276387d2c64de2c98e2f751a120bb805216be559b68a836R38). i dont think there's a way around doing this due to how the faucet works—just calling it out.

Nothing else I can think of.

dtbuchholz:
> This looks good, but I'm a bit surprised about the URLs considering this is a testnet. What will the chain.love URLs look like for mainnet? Maybe I'm missing something about how they differentiate between testnet and mainnet?

@asutula hm, yeah these feel like mainnet RPCs; it's what we set up with them tho: https://recall.chain.love. once we go to mainnet, i'm guessing we'd want `https://evm.recall.chain.love` to be the mainnet RPC...which is a terrible UX b/c people will have to migrate—and likely, accidentally start using mainnet instead of testnet.

e.g., i'd expect us to use something like `https://evm.recall.testnet.chain.love` for testnet, and `https://evm.recall.chain.love` for mainnet. filecoin's a good example: https://filecoin.chain.love/

dtbuchholz:
sander noted: _we can alias for now, and then after they switch, we can change values in the SDK and docs_

so i think we're good to publish these changes now—mainly because we have ETHDenver hackathons starting tomorrow, and the load balancing etc with chain.love will be useful.

asutula:
> sander noted we'll request chain.love to make a change and use a `testnet` subdomain: _we can alias for now, and then after they switch, we can change values in the SDK and docs_
> 
> so i think we're good to publish these changes now—mainly because we have ETHDenver hackathons starting tomorrow, and the load balancing etc with chain.love will be useful.

I don't know what "alias" means in this context, but I believe you and @sanderpick. Cool!

sanderpick:
@dtbuchholz fyi, they have added the new URLs, but the old ones do NOT work anymore... PR here: https://github.com/recallnet/js-recall/pull/87

PR #87, opened 2025-02-26
chore: update chain.love testnet urls

sanderpick:


packages/network-constants/src/index.ts
```diff
@@ -47,7 +47,7 @@ export const TESTNET_FAUCET_URL = "https://faucet.recall.network";
 
 // Miscellaneous
 export const RPC_TIMEOUT = 60_000;
-export const MIN_TTL = 3600n; // one hour
```
sanderpick: was this a typo?

dtbuchholz: The `n` suffix is for big ints. I originally used them through the SDK, but I can't remember if I need to use them or can leverage a normal number. 

Anyways, probably fine to leave as 3600n

sanderpick: cool, reverted :+1:

dtbuchholz (approved):
🤘

PR #89, opened 2025-02-27
Buckets improvements + DX improvements

asutula:
- Use NextJS `Link` component in breadcrumb so we get partial page loads on breadcrumb navigation (vs reloading entire page)
- Better pending state when creating a bucket, properly detect no buckets to display
- Allow bucket metadata to be undefined
- Allow object metadata to be undefined
- Use `tsc` in dev mode for better IDE support of click to navigate to source. 
- Use `.js` for esm build file extensions to be consistent with above bullet when in dev mode.

### DIFF
--- a/packages/address-utils/package.json
+++ b/packages/address-utils/package.json
@@ -7,7 +7,7 @@
   "exports": {
     "./display": {
       "types": "./dist/display.d.ts",
-      "import": "./dist/display.mjs",
+      "import": "./dist/display.js",
       "require": "./dist/display.cjs"
     }
   },
@@ -15,7 +15,7 @@
     "dist"
   ],
   "scripts": {
-    "dev": "tsup --watch",
+    "dev": "tsc --watch",
     "build": "tsup",
     "lint": "eslint . --max-warnings 0",
     "clean": "rm -rf .turbo node_modules dist"
--- a/packages/address-utils/tsup.config.ts
+++ b/packages/address-utils/tsup.config.ts
@@ -6,6 +6,6 @@ export default defineConfig({
   dts: true,
   clean: true,
   outExtension: ({ format }) => ({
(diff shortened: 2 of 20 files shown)


packages/address-utils/package.json
```diff
@@ -7,15 +7,15 @@
...
+      "import": "./dist/display.js",
       "require": "./dist/display.cjs"
     }
   },
   "files": [
     "dist"
   ],
   "scripts": {
-    "dev": "tsup --watch",
+    "dev": "tsc --watch",
```
asutula, line 18: `tsc` simply compiles the typescript files and outputs `.js` esm modules. I don't know why, but VSCode plays way better with this configuration. It allows you to get proper types and also click on a symbol to go the the typescript source, even if that source is in another monorepo package. We had lost that ability with the previous configuration that used `tsup` in dev mode.

packages/address-utils/tsup.config.ts
```diff
@@ -6,6 +6,6 @@ export default defineConfig({
   dts: true,
   clean: true,
   outExtension: ({ format }) => ({
-    js: format === "esm" ? ".mjs" : ".cjs",
+    js: format === "esm" ? ".js" : ".cjs",
```
asutula, line 9: Since we use tsc in dev mode and it outputs `.js` files, and `package.json` lists those `.js` files for the `import` option of a module, I updated the production build to output `.js` files for esm format.

packages/address-utils/package.json
```diff
@@ -7,15 +7,15 @@
   "exports": {
     "./display": {
       "types": "./dist/display.d.ts",
-      "import": "./dist/display.mjs",
+      "import": "./dist/display.js",
```
asutula, line 10: For some reason, VSCode likes this better.

PR #90, opened 2025-02-28
Correct use of package.json exports and tsc/tsup output

asutula:
Something seemed off about our package exports and ts types. I noticed that `tsup` was outputting special ts types for cjs that we weren't referencing. Some research and this tool Are the Types Wrong (https://arethetypeswrong.github.io/?p=@recallnet/sdk@0.0.4 and the cli command `pnpm dlx @arethetypeswrong/cli --pack`) helped me understand what was going on and what the correct configuration should be.

### DIFF
--- a/.zed/settings.json
+++ b/.zed/settings.json
@@ -1,0 +1,18 @@
+// Folder-specific settings
+//
+// For a full list of overridable settings, and general information on folder-specific settings,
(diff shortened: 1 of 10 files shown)


dtbuchholz (approved):
makes sense...I never dug too deep into the exports + cjs stuff but had wondered about the cts export

PR #112, opened 2025-03-17
feat: agent toolkit

dtbuchholz:
## Summary

This adds a new `@recallnet/agent-toolkit`, which brings framework agnostic agentic flows to Recall. 

## Details

Heavily inspired by the [Stripe agent toolkit](https://github.com/stripe/agent-toolkit), it provides:
- A set of `shared` tools, prompts, and API to the Recall network
- Framework adapters for OpenAI, AI SDK, LangChain, and MCP—all which use the same shared tools but have slightly different dependencies and instantiation logic etc
- Builds (but doesnt export) a set of examples for each framework so that devs can run them
   - Note: I'm not sure if this is the right pattern, or if we should just have separate repos/docs and remove the `examples`. Counterpoint: having them in the package helped with testing the adapters work properly. 

The best way to see how these work is to review the README, or install/build and run `node dist/examples/...` for one of the packages. The MCP example is probably the easiest to inspect via: `SERVER_PORT=3001 npx @modelcontextprotocol/inspector node dist/examples/mcp.js --tools=all --network=localnet`


dtbuchholz:
@mzkrasner you're probably the best one to review this from a logic perspective. if you run `pnpm build`, the `dist/examples` have some prebuilt programs you can run, including MCP.

@asutula from a packaging perspective, maybe you could tal? the one callout is that i ended up with a barrel file export for the `shared` module (`@recallnet/agent-toolkit/shared`). i can split out the paths to be more explicit, but the way i see this package being used is in two ways:
1. youre working with a framework/tools like LangChain, OAI, AI SDK, or MCP—and you want to easily import prebuilt Recall tools that agents can use w/o much effort. e.g., `@recallnet/agent-toolkit/mcp`
2. you want to have greater control and implement things on your own—which is you want to use the `shared` deps.  e.g., `@recallnet/agent-toolkit/shared`

since i think most ppl will opt for the flow in (1), i decided the barrel file export might be easier so that there aren't too many different paths to choose from? idk or maybe it's fine to replace `@recallnet/agent-toolkit/shared` with:
- `@recallnet/agent-toolkit/api`
- `@recallnet/agent-toolkit/configuration`
- `@recallnet/agent-toolkit/functions`
- `@recallnet/agent-toolkit/parameters`
- `@recallnet/agent-toolkit/prompts`
- `@recallnet/agent-toolkit/tools`
- `@recallnet/agent-toolkit/util`

dtbuchholz:
@asutula heads up that if you bump zod from `^3.24.1` to `^3.24.2`, it causes cascading updates to viem/wagmi/rainbowkit, which then somehow causes build failures like these:
- https://github.com/recallnet/js-recall/actions/runs/13906497133/job/38910686151#step:11:499
- (this one is a bit more descriptive about rainbowkit complaints) https://github.com/recallnet/js-recall/actions/runs/13905343876/job/38906964305#step:11:490

not sure why, but the zod bump will pull in new wagmi/rainbowkit versions that break the build. after reverting back to `^3.24.1`, the build worked fine. weird.

PR #113, opened 2025-03-18
Use query params for bucket nav and delimiter

asutula:
We were previously just using URL pathname to represent buckets and their contained data as you browse. I realized simple URL pathnames can't represent everything that is supported by bucket keys, so the next best solution is using query parameters. The new links will look similar to:

```https://portal.recall.network/buckets/0xFF00000000000000000000000000000000000251?path=a/b/c/```

We can now support weird stuff like the prefix `///a/b///c` or anything else really.

I included a middleware function that redirects the old style URLs to the new style.

I also added support for a `delimiter` query param that allows you to specify the delimiter to use when querying the bucket. The default is `/` if you don't specify one. There is not yet some UI to allow the user to easily change the delimiter, but that will come in another PR soon.

I updated the bucket breadcrumb UI to use the actual delimiter to separate prefix parts.


dtbuchholz (approved):
nice, lgtm!

PR #116, opened 2025-03-20
feat: contract/abi changes for testnet v0.1.5

dtbuchholz:
- update contracts values for the new testnet.
- update the contracts submodule, which has minor ABI changes.
   - there are no breaking changes for the bucket or credit wrappers. the blob manager had some minor adjustments, but it's not used in the Portal or anywhere else really
   - the main callout is that the `queryObjects` method's return value now includes a new field: `expiry`
- handle ABI changes in `sdk` package and update tests accordingly.
   - note: the blob hash has some new behavior, so the `sdk` tests had to account for this.
      > _sander_: the top level hash for a blob is now a hash sequence. the first element in the sequence is the actual user blob.
- also, closes https://github.com/recallnet/js-recall/issues/117 with a small fix wrt Portal/sdkx and the blob recovery hash


dtbuchholz:
note: we'll want the chain.love RPCs to update and use the new testnet before merging this

asutula (approved):
Nice!

Issue #115, opened 2025-03-19
Unified Agent Toolkit for Recall


carsonfarmer:
# Unified Agent Toolkit for Recall

## Description: What is it?
A unified toolkit that standardizes and streamlines the integration of Recall’s agent functionality across multiple frameworks and SDKs. It consolidates core logic and tool definitions into a single, reusable codebase, and provides framework-specific adapters (for OpenAI SDK, AI SDK, MCP, LangChain, etc.) along with a Recall-centric MCP server CLI. This ensures consistency in developer experience (DX) and reduces duplication of effort.

## Problem: What problem is this solving?
- **Duplication of Code:** Currently, similar logic is implemented separately across different frameworks and agentic tools. This leads to maintenance overhead and inconsistent behavior.
- **Inconsistent Patterns:** Interaction with Recall lacks standardization—tools, prompts, return types, and error handling vary across implementations, complicating testing and refinements.
- **Inefficient Integrations:** Developers have to write custom integrations for each framework, slowing down development and increasing the potential for bugs.
- **Complex MCP Deployment:** Setting up and running MCP servers requires cloning and building from source, which is cumbersome and error-prone.

## Why: How do we know this is a real problem and worth solving?
- **Developer Feedback:** Engineers have expressed frustration over repeated work and inconsistent integration patterns when building with multiple frameworks.
- **Industry Best Practices:** Standardization and reuse of core logic are proven methods for increasing productivity and reducing error rates.
- **Improved DX:** By offering a unified API and standardized tooling, we can significantly lower the onboarding barrier and improve the maintainability of integrations.
- **Operational Efficiency:** A ready-to-run MCP server via a simple CLI command (e.g., `npx @recallnet/modelcontextprotocol`) can dramatically simplify deployment and adoption.

## Success: How do we know if we’ve solved this problem?
- **Reduced Code Rework:** Measurable decrease in duplicated code across different SDKs and frameworks.
- **Standardized Interactions:** Consistent tool definitions, prompt formats, return types, and error handling observed across integrations.
- **Developer Satisfaction:** Positive feedback from developers regarding ease of integration and reduced setup time.
- **Adoption Metrics:** Increased usage of the RecallAPI and RecallAgentToolkit adapters, along with higher engagement with the MCP server CLI.
- **Faster Deployment:** Reduced time-to-deploy for MCP servers in external products.

## Audience: Who are we building for?
[...]

PR #130, opened 2025-03-21
feat: agent toolkit library

dtbuchholz:
## Summary

This adds a new `@recallnet/agent-toolkit`, which brings framework agnostic agentic flows to Recall. Closes https://github.com/recallnet/js-recall/issues/115

## Details

Heavily inspired by the [Stripe agent toolkit](https://github.com/stripe/agent-toolkit), it provides:
- A set of `shared` tools, prompts, and API to the Recall network
- Framework adapters for OpenAI, AI SDK, LangChain, and MCP—all which use the same shared tools but have slightly different dependencies and instantiation logic etc
- Builds (but doesnt export) a set of examples for each framework so that devs can run them
   - Note: I'm not sure if this is the right pattern, or if we should just have separate repos/docs and remove the `examples`. Counterpoint: having them in the package helped with testing the adapters work properly. 

The best way to see how these work is to review the README, or install/build and run `node dist/examples/...` for one of the packages. The MCP example is probably the easiest to inspect via: `SERVER_PORT=3001 npx @modelcontextprotocol/inspector node dist/examples/mcp.js --tools=all --network=localnet`

## Notes

if you run `pnpm build`, the `dist/examples` have some prebuilt programs you can run, including MCP. we might want to move the related deps to dev deps.

from a packaging perspective, the one callout is that i ended up with a barrel file export for the `shared` module (`@recallnet/agent-toolkit/shared`). i can split out the paths to be more explicit, but the way i see this package being used is in two ways:
1. youre working with a framework/tools like LangChain, OAI, AI SDK, or MCP—and you want to easily import prebuilt Recall tools that agents can use w/o much effort. e.g., `@recallnet/agent-toolkit/mcp`
2. you want to have greater control and implement things on your own—which is you want to use the `shared` deps.  e.g., `@recallnet/agent-toolkit/shared`

since i think most ppl will opt for the flow in (1), i decided the barrel file export might be easier so that there aren't too many different paths to choose from? idk or maybe it's fine to replace `@recallnet/agent-toolkit/shared` with:
- `@recallnet/agent-toolkit/api`
- `@recallnet/agent-toolkit/configuration`
- `@recallnet/agent-toolkit/functions`
- `@recallnet/agent-toolkit/parameters`
- `@recallnet/agent-toolkit/prompts`
- `@recallnet/agent-toolkit/tools`
- `@recallnet/agent-toolkit/util`


dtbuchholz:
i have to make some tweaks with the new requirements in https://github.com/recallnet/js-recall/issues/115. mainly:
1. split out the `examples/mcp.ts` into its own package at `packages/mcp`, and combine with https://github.com/recallnet/recall-mcp/. this probably makes sense to do in a separate PR, though.
2. general refactoring—but i think it'd be nice to make any additional tweaks based on feedback, like do the APIs make or the resource/permission patterns make sense? (since it copied Stripe, i actually quite like the way it's handled.)

dtbuchholz:
@carsonfarmer just wanted to flag this PR for a proper review. i can do a bit more refactoring before i officially open it, but before doing so, i wanted to get a gut check on the current design. 

i added some notes in the PR description as well as in the comment above.

dtbuchholz:
note: for whatever feedback you have, plus the general refactoring & tests—i'll open a secondary PR that points to this one, making it easier to review any final changes. this can also include the `packages/mcp` setup that i mentioned above.

packages/agent-toolkit/README.md
```diff
@@ -0,0 +1,231 @@
...
+| [AI SDK](https://github.com/vercel/ai)                                  | `@recallnet/agent-toolkit/ai-sdk`    |
+
+## Usage
+
+### Installation
+
+Install the package:
+
+```bash
+npm install @recallnet/agent-toolkit
```
carsonfarmer, line 43: Do we want to also have pnpm install steps? They aren't really all that different. Additionally, does this work with deno? Probably would via `deno add npm:@recallnet/agent-toolkit`..?

packages/agent-toolkit/README.md
```diff
@@ -0,0 +1,231 @@
...
+The `@recallnet/agent-toolkit` package provides a set of tools that can be used to interact with Recall. Each tool has a set of resources and permissions that are used to determine what actions can be performed. It allows you to refine which tools are available by specifying the resources and permissions for each tool.
+
+The resources are:
+
+- `account`: Account operations, like reading balances or buying credit
+- `bucket`: Bucket operations, like creating buckets or adding objects
+
+The permissions are:
+
+- `read`: Read-only operations (i.e., do not cost any tokens/credit)
```
carsonfarmer:
```suggestion
- `read`: Read-only operations (i.e., operations do not cost any tokens/credit)
```

packages/agent-toolkit/README.md
```diff
@@ -0,0 +1,231 @@
...
+- `write`: Write operations, which modify state (i.e., submit transactions)
```
carsonfarmer:
```suggestion
- `write`: Write operations, which modify state (i.e., operations may submit transactions)
```

packages/agent-toolkit/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+- Resource: `bucket`
+- Permission: `write`
+
+Upon instantiating the `RecallAgentToolkit`, you can define the resources and permissions you want to be available to the agent.
+
+```ts
+import { Configuration } from "@recallnet/agent-toolkit/shared";
+
+const config: Configuration = {
```
carsonfarmer, line 125: Is this the full set of config options at the moment?

packages/agent-toolkit/package.json
```diff
@@ -0,0 +1,87 @@
...
+  "dependencies": {
+    "@ai-sdk/openai": "^1.2.5",
+    "@langchain/core": "^0.3.42",
+    "@langchain/openai": "^0.4.4",
+    "@modelcontextprotocol/sdk": "^1.7.0",
+    "@recallnet/chains": "workspace:*",
+    "@recallnet/sdk": "workspace:*",
+    "ai": "^4.1.61",
+    "dotenv": "^16.4.7",
+    "langchain": "^0.3.19",
```
carsonfarmer: Is there any way we can move some of these to peer dependencies? Or is code from these packages actually used directly within our package itself?

https://nodejs.org/en/blog/npm/peer-dependencies

packages/agent-toolkit/src/ai-sdk/tool.ts
```diff
@@ -0,0 +1,27 @@
...
+ * const tool = new RecallTool(recallAPI, "get_account_info", "Get account info", getAccountInfoParameters);
+ * ```
+ */
+export default function RecallTool(
+  recallAPI: RecallAPI,
+  method: string,
+  description: string,
+  schema: z.ZodObject<any, any, any, any, { [x: string]: any }>,
+): CoreTool {
+  return tool({
```
carsonfarmer, line 21: ah ok, so this probably answers my previous question

carsonfarmer: oh no wait, this would still work right? to mark ai-ask as a peer dep?

packages/agent-toolkit/src/ai-sdk/toolkit.ts
```diff
@@ -0,0 +1,89 @@
...
+      this.tools[tool.method] = RecallTool(
+        this._recall,
+        tool.method,
+        tool.description,
+        tool.parameters,
+      );
+    });
+  }
+
+  middleware(): LanguageModelV1Middleware {
```
carsonfarmer: can you explain why we need this? just because of the framework agnostic factor? because in this case, we're not _actually_ providing any "middleware" functionality, we're just passing things right along...

packages/agent-toolkit/src/examples/ai-sdk.ts
```diff
@@ -0,0 +1,72 @@
...
+        write: true,
+      },
+    },
+    context: {
+      network: RECALL_NETWORK,
+    },
+  },
+});
+
+const model = wrapLanguageModel({
```
carsonfarmer: so, this doesn't _actually_ do anything, because our toolkit doesn't actually provide any middleware functionality, right?

packages/agent-toolkit/src/examples/langchain.ts
```diff
@@ -0,0 +1,86 @@
+import type { ChatPromptTemplate } from "@langchain/core/prompts";
+import { ChatOpenAI } from "@langchain/openai";
+import { config } from "dotenv";
+import { existsSync } from "fs";
+import { AgentExecutor, createStructuredChatAgent } from "langchain/agents";
+import { pull } from "langchain/hub";
+import { dirname, resolve } from "path";
+import { fileURLToPath } from "url";
+
+import { RecallAgentToolkit } from "../langchain/index.js";
```
carsonfarmer: can we set things up so that the import is actually the public api import? I think we can do that with aliases right? so it'd be `import { RecallAgentToolkit } from "@recallnet/agent-toolkit/langchain"` here?

packages/agent-toolkit/src/examples/mcp.ts
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env node
```
carsonfarmer: This is great, though as you point it, it'd be nice if we just made this its own sub-package so folks can just use it not as an example, but rather as a standalone mcp implementation, which _also_ acts as an example!

Note also that this (https://github.com/wong2/litemcp/) is a pretty nice framework for building mcp services very easily.

packages/agent-toolkit/src/shared/parameters.ts
```diff
@@ -0,0 +1,172 @@
+import { z } from "zod";
+
+/**
+ * Parameters for the getAccountInfo function
+ * @param address - The address of the account to get account info for (optional)
+ */
+export const getAccountInfoParameters = z.object({
+  address: z
+    .string()
+    .regex(/^0x[a-fA-F0-9]{40}$/)
```
carsonfarmer: should probably make this into a const

packages/agent-toolkit/src/shared/parameters.ts
```diff
@@ -0,0 +1,172 @@
+import { z } from "zod";
```
carsonfarmer, line 1: lovely module!

carsonfarmer: I didn't double check your zod defs here against any official source of truth... do we _have_ an official source of truth for these from which we can derive the zod schemas? Like json schema docs?

packages/agent-toolkit/src/shared/prompts.ts
```diff
@@ -0,0 +1,72 @@
+export const getAccountInfoPrompt = `
```
carsonfarmer, line 1: nice, again, wondering if we should have a single source of truth for these (maybe this is it), but that is language agnostic? like just adding it to the docs string for a json schema definition?

carsonfarmer: what is best practices here?

packages/agent-toolkit/tsup.config.ts
```diff
@@ -0,0 +1,87 @@
...
+    outDir: "dist/shared",
+    format: ["cjs", "esm"],
+    dts: true,
+    sourcemap: true,
+    clean: true,
+    outExtension: ({ format }) => ({
+      js: format === "esm" ? ".js" : ".cjs",
+    }),
+  },
+  // Also, publish examples so that they can be used as standalone processes (e.g., MCP server)
```
carsonfarmer: For the mcp server, this makes sense. Maybe not for the others though?

carsonfarmer (approved):
This is looking really solid to me @dtbuchholz. I'm giving this a 👍 to finalize and move forward. Then we'll get a proper review from myself and maybe @mzkrasner.

packages/agent-toolkit/README.md
dtbuchholz: yeah, i kept it simple in case we want to enhance this in the future. if you expand it, it looks like this. the config lets you define:
- the "action" groups and permissions you want to enable (like `bucket` related scopes with `read` only functionality)
- any additional context that can be used by the actual tool calls. 
```ts
export type Resource = "account" | "bucket";
export type Permission = "read" | "write";
export type Configuration = {
  actions: {
    [K in Resource]?: {
      [K in Permission]?: boolean;
    };
  };
  context: {
    network?: string;
    [key: string]: any;
  };
};
```
we dont _really_ use any special context in actual tool calls, but i wanted to give us the option to do so. the existing `context` has a `network` field, which only is used upon the `RecallAPI` instantiation.
```ts
export default class RecallAPI {
  constructor(
    privateKey: string,
    context?: Context,
    serializer: (data: any) => string = jsonStringify,
  ) {
    // Use the `context.network` string to set up the connected chain
    const chain =
      context?.network !== undefined &&
      checkChainName(context.network as ChainName)
        ? getChain(context.network as ChainName)
        : testnet;
    const walletClient = walletClientFromPrivateKey(privateKey as Hex, chain);
    const recallClient = new RecallClient({ walletClient });

    this._recall = recallClient;
    this._context = context || {};
    this._serialize = serializer;
  }
```
so, how can we better use the `Configuration`? the `context` is passed to every tool function. e.g., a few methods use a `params.bucket` parameter and pass that address to the JS SDK method call. instead, we could choose to alter that flow such that the `context` provides the param as a "global" variable. here's what the current state looks like—you can see that `context` isnt used at all:
```ts
export const addObject = async (
  recall: RecallClient,
  context: Context,
  params: z.infer<typeof addObjectParameters>,
): Promise<Result<{ txHash: string }>> => {
  try {
    const metadata = params.metadata ?? {};
    const dataToStore =
      typeof params.data === "string"
        ? new TextEncoder().encode(params.data)
        : params.data;

    const { meta } = await recall
      .bucketManager()
      .add(params.bucket as Address, params.key, dataToStore, {
        overwrite: params.overwrite ?? false,
        metadata: { ...metadata },
      });
    if (!meta?.tx) {
      return { success: false, error: "Transaction not found" };
    }
    return { success: true, result: { txHash: meta.tx.transactionHash } };
  } catch (error: any) {
    return { success: false, error: `Failed to add object: ${error.message}` };
  }
};
```
e.g., we could enhance the config to let you:
1. you set a global bucket
2. when you call `addObject`, we make the `params.bucket` optional; if you set a global bucket, we'll use that unless overridden by `params.bucket`
```ts
export const addObject = async (
  recall: RecallClient,
  context: Context,
  params: z.infer<typeof addObjectParameters>,
): Promise<Result<{ txHash: string }>> => {
    let bucket;
    if (context.bucket) bucket = context.bucket;
    if (params.bucket) bucket = params.bucket;
    if (bucket === undefined) throw new Error("No bucket set in context or passed to method");
    // ...
}

File: packages/agent-toolkit/package.json (hunk 5 above)
Comment (dtbuchholz) [reply packages/agent-toolkit/package.json]: yeh, i'll move them—a handful arent actually dependences and are here because they're used in the "examples" module. instead, we should split them out into dev deps, i think, since they're not actually exported and are here for demo purposes

File: packages/agent-toolkit/src/examples/ai-sdk.ts (hunk 8 above)
Comment (dtbuchholz) [reply packages/agent-toolkit/src/examples/ai-sdk.ts]: true. i sorta just copied the Stripe toolkit, which had a bunch of custom middleware logic that tracked Stipe account billing. let me see if i can just remove this middleware altogether.

File: packages/agent-toolkit/src/examples/langchain.ts (hunk 9 above)
Comment (dtbuchholz) [reply packages/agent-toolkit/src/examples/langchain.ts]: umm, if we want to allow people to run the examples (e.g., clone the repo, build, and run), no. the `@recallnet/agent-toolkit/langchain` package would be a circular, self-referencing dep; we can't list it as a dependency.

if/when we put this example in the docs site, then yes.

but, i'll move the `src/examples` to just `examples` in root, since i think that's the better practice.

File: packages/agent-toolkit/src/shared/parameters.ts (hunk 12 above)
Comment (dtbuchholz) [reply packages/agent-toolkit/src/shared/parameters.ts:1]: umm, no, i dont think we have an official source of truth. these definitions loosely match what we use in our JS SDK. 

for example, our SDK might require a viem `Address` type (like a user or bucket address). with zod, i define this more generically as a string with a regex hex pattern requirement.

however, maybe we need to create a new package as the source of truth—or, add this to the JS SDK? i made a note to do this in https://github.com/recallnet/js-recall/issues/91

File: packages/agent-toolkit/src/shared/prompts.ts (hunk 13 above)
Comment (dtbuchholz) [reply packages/agent-toolkit/src/shared/prompts.ts:1]: yeah i guess we'd need a language-agnostic spec. perhaps the scope for the agent memory specs https://github.com/recallnet/competition-v0/issues/1 could be expanded to incorporate this idea?

or, since the MCP spec uses JSON RPC 2.0, we could create a JSON RPC spec for all of these prompts etc—and then import that spec into the agent toolkit https://spec.modelcontextprotocol.io/specification/2024-11-05/server/prompts/#getting-a-prompt.

File: packages/agent-toolkit/tsup.config.ts (hunk 14 above)
Comment (dtbuchholz) [reply packages/agent-toolkit/tsup.config.ts]: yeah true—i'll remove this, and since the MCP server will be published on its own, i'll remove it as well. instead, these will just live as a standalone examples dir, and docs will better showcase their usage.

File: packages/agent-toolkit/package.json (hunk 5 above)
Comment (dtbuchholz) [reply packages/agent-toolkit/package.json]: ah i forgot that _most_ of these are used in the actual adapters, and not the examples. your peer deps idea makes sense, i think. maybe something like:
```json
   {
     "name": "@recallnet/agent-toolkit",
     "exports": {
       "./langchain": "./dist/langchain/index.js",
       "./mcp": "./dist/mcp/index.js",
       "./openai": "./dist/openai/index.js",
       "./ai-sdk": "./dist/ai-sdk/index.js"
     },
     "peerDependencies": {
       "@langchain/core": "^0.3.42",
       "@modelcontextprotocol/sdk": "^1.7.0",
       "openai": "^4.87.3",
       "ai": "^4.1.61"
     },
     "peerDependenciesMeta": {
       "@langchain/core": {
         "optional": true
       },
       "@modelcontextprotocol/sdk": {
         "optional": true
       },
       "openai": {
         "optional": true
       },
       "ai": {
         "optional": true
       }
     }
   }
```

dtbuchholz:
okay, i think i addressed your feedback @carsonfarmer:
- clean up the package exports and dependencies:
   - move things into an `examples` dir outside of `src`
   - set up optional peer or dev deps accordingly
- linting fixes for `any` types (mostly, ignoring them)
- clean up the ai-sdk adapter/example (removed the middleware) and MCP example (default to all `--tools` if not provided)
- make the zod address/regex type a constant so that it's reused across the different method parameter schemas
- docs cleanups for the install steps and examples

packages/agent-toolkit/src/shared/prompts.ts
carsonfarmer: Can you create a ticket for this in the competition repo pls?

carsonfarmer (approved):
Yup, agreed. It looks like you've addressed my comments. Let's get this merged and ready to use asap. Can we publish pre-release versions of these libs so that we can use the in other "internal" projects for now?

carsonfarmer:
Also, let's do a fast follow (as time permits) on the standalone mcp service with proper "bin" tooling, so we can do `npx @reacallnet/agent-toolkit/mcp` in mcp configs etc.

dtbuchholz:
> Can we publish pre-release versions of these libs so that we can use the in other "internal" projects for now?

We use `changesets` for versioning/publishing, which does have a prerelease versioning feature. I'd have to see how that works with the monorepo, though, since it might conflict with the other packages that don't need it. Worst case, I'll just publish it as `0.0.1`. 

I'll merge this PR as-is. Then, I'll set up the separate MCP package/binary, and within that PR, I'll include what we need to version/release both the agent toolkit and the MCP server. I should be able to get that done before tmrw.

PR #135, opened 2025-03-26
feat: mcp server

dtbuchholz:
## Summary

Add an official `@recallnet/mcp` binary that can be ran with `npx`.

## Details

- The design is, basically, unchanged from the [agent toolkit example](https://github.com/recallnet/js-recall/blob/main/packages/agent-toolkit/examples/mcp.ts) MCP server
- You can test the tool out by building and then running `node dist/index.js --private-key=0x...` (or by setting the RECALL_PRIVATE_KEY env var). Just make sure you're explicit since it won't be read from any `.env.` file.
- The docs are pretty light and just point to the docs site, which we can update later

### DIFF
--- a/packages/mcp/README.md
+++ b/packages/mcp/README.md
@@ -1,0 +1,102 @@
+# Recall MCP Server
+
+> Official MCP server for interacting with the Recall network.
+
+## Table of Contents
+
+- [Table of Contents](#table-of-contents)
+- [Background](#background)
+- [Usage](#usage)
+  - [Available tools](#available-tools)
+- [Development](#development)
+- [Contributing](#contributing)
+- [License](#license)
+
+## Background
+
+The `@recallnet/mcp` server allows agents to interact with the Recall network using the [Model Context Protocol (MCP)](https://github.com/modelcontextprotocol). See the [Recall MCP documentation](https://docs.recall.network/tools/mcp) for more information.
+
+## Usage
+
+Run the package with npx. You'll need to provide a Recall private key and optionally a network (e.g., `testnet` or `localnet`).
+
+```bash
+npx @recallnet/mcp --private-key=0x... --network=testnet
+```
+
+Or, set the `RECALL_PRIVATE_KEY` and `RECALL_NETWORK` environment variables:
+
+```bash
+RECALL_PRIVATE_KEY=0x... RECALL_NETWORK=testnet npx @recallnet/mcp
+```
+
+Optionally, you can specify the tools you want to enable. By default, all tools are enabled.
+
+```bash
+npx @recallnet/mcp --private-key=0x... --network=testnet --tools=bucket.read,bucket.write
+```
+
+### Available tools
+
+The server exposes the following MCP tools:
+
+| Tool name              | Tool scope      | Description                                                            |
+| ---------------------- | --------------- | ---------------------------------------------------------------------- |
+| `get_account_info`     | `account.read`  | Get Recall account information (e.g., address, balance)                |
+| `get_credit_info`      | `account.read`  | Get Recall account credit information (e.g., credit available or used) |
+| `buy_credit`           | `account.write` | Buy credit for Recall account                                          |
+| `list_buckets`         | `bucket.read`   | List all buckets owned by an address                                   |
+| `create_bucket`        | `bucket.write`  | Create a bucket in Recall                                              |
+| `get_or_create_bucket` | `bucket.write`  | Get or create a bucket in Recall (using alias)                         |
+| `add_object`           | `bucket.write`  | Add an object to a Recall bucket                                       |
+| `get_object`           | `bucket.read`   | Get an object from a Recall bucket                                     |
+| `query_objects`        | `bucket.read`   | Query objects in a Recall bucket                                       |
+
+## Development
+
+Clone the repository:
+
+```bash
+git clone https://github.com/recallnet/js-recall.git
+```
+
+And change into the `packages/mcp` directory:
+
+```bash
+cd js-recall/packages/mcp
+```
+
+Install dependencies and build the binary:
+
+```bash
+pnpm install
+pnpm build
+```
+
+Run the server directly from the `dist` directory:
+
+```bash
+node dist/index.js --private-key=0x... --network=testnet
+```
+
+The following `pnpm` commands are available:
+
+| Command             | Description                                     |
+| ------------------- | ----------------------------------------------- |
+| `pnpm build`        | Build the binary                                |
+| `pnpm dev`          | Run in development mode                         |
+| `pnpm lint`         | Lint the project with ESLint                    |
+| `pnpm lint:fix`     | Lint the project and fix linting errors         |
+| `pnpm format:check` | Check if the project is formatted with Prettier |
+| `pnpm format`       | Format the project (writes files)               |
+
+## Contributing
+
+PRs accepted.
+
+Small note: If editing the README, please conform to
+the [standard-readme](https://github.com/RichardLitt/standard-readme) specification.
+
+## License
+
+MIT OR Apache-2.0, © 2025 Recall Network Corporation
--- a/packages/mcp/README.md
+++ b/packages/mcp/README.md
@@ -7,6 +7,9 @@
 - [Table of Contents](#table-of-contents)
 - [Background](#background)
 - [Usage](#usage)
+  - [Configure for Cursor or Claude Desktop](#configure-for-cursor-or-claude-desktop)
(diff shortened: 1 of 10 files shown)


dtbuchholz:
Re: prerelease versioning—the way `changesets` works is:
- If a package has _never_ been published, it'll automatircally get published with whatever version it has set in its package file. Usually, you run `pnpm changeset` to trigger npm publishing, but it's a special case where that isn't a prerequisite.
- If a package _has_ been published, it's only published if you run `pnpm changeset` to bump the version.

So, the changesets prerelease functionality will only work for existing packages. If needed, there are ways where you can "ignore" a package and ensure changesets won't publish it. Or, we could see what happens if we set the versions manually to something like `0.0.1-pre.0`...I think this will work like (1) above.

Fwiw, all of our other packages are released as patch bumps (there are no majors/minors, yet), with the intent they're prereleases/alpha software.

packages/mcp/README.md
```diff
@@ -0,0 +1,102 @@
...
+Or, set the `RECALL_PRIVATE_KEY` and `RECALL_NETWORK` environment variables:
+
+```bash
+RECALL_PRIVATE_KEY=0x... RECALL_NETWORK=testnet npx @recallnet/mcp
+```
+
+Optionally, you can specify the tools you want to enable. By default, all tools are enabled.
+
+```bash
+npx @recallnet/mcp --private-key=0x... --network=testnet --tools=bucket.read,bucket.write
```
carsonfarmer, line 39: Let's add a usage section for using Cursor or Claude Desktop.

packages/mcp/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "dev": "tsc --watch",
+    "build": "tsup && node -e \"require('fs').chmodSync('dist/index.js', '755')\"",
+    "lint": "eslint .",
+    "lint:fix": "eslint . --fix",
+    "format": "prettier --write .",
+    "format:check": "prettier --check ."
+  },
+  "dependencies": {
+    "@modelcontextprotocol/sdk": "^1.7.0",
+    "@recallnet/agent-toolkit": "workspace:*",
```
carsonfarmer, line 28: so nice!

packages/mcp/src/index.ts
```diff
@@ -0,0 +1,174 @@
...
+ * @param privateKey - The private key to use
+ * @param network - The Recall network to use: `testnet` or `localnet` (defaults to `testnet`)
+ */
+type Options = {
+  tools?: string[];
+  privateKey?: string;
+  network?: string;
+};
+
+// The server will accept flags for a `--private-key` (hex string), `--tools` (comma separated),
```
carsonfarmer: Since this is strictly a stdio server, do you think we need this? The client (be it Claude Desktop, or whatever) is probably always going to use npx and pass an env, right? Correct me if I'm wrong here though!

packages/mcp/src/index.ts
```diff
@@ -0,0 +1,174 @@
...
+
+// Note: since we require a private key, we only expose both read and write tools
+const ACCEPTED_TOOLS = [
+  "account.read",
+  "account.write",
+  "bucket.read",
+  "bucket.write",
+];
+
+// eslint-disable-next-line turbo/no-undeclared-env-vars
```
carsonfarmer: You _could_ extend process.env so that you get type checking here if you wanted. No problem if not though.

packages/mcp/src/index.ts
```diff
@@ -0,0 +1,174 @@
...
+const { RECALL_PRIVATE_KEY, RECALL_NETWORK } = process.env;
+
+/**
+ * Parse the command line arguments
+ * @param args - The command line arguments—one or more of:
+ *  `--tools=all` (or `--tools=resource.action`, like `--tools=account.read`),
+ *  `--private-key=$KEY`, or `--network=$NETWORK`
+ * @returns The parsed options
+ */
+export function parseArgs(args: string[]): Options {
```
carsonfarmer: I don't _love_ that we have a custom parseArgs function here. There _must_ be small libraries that we could use here instead? I guess probs fine so long as it's easy to maintain.

packages/mcp/src/index.ts
```diff
@@ -0,0 +1,174 @@
...
+  const server = new RecallAgentToolkit({
+    privateKey: options.privateKey!,
+    configuration: configuration,
+  });
+
+  const transport = new StdioServerTransport();
+  await server.connect(transport);
+  // We use console.error instead of console.log since console.log will output to stdio, which
+  // will confuse the MCP server
+  console.error(chalk.green("Recall MCP Server running on stdio"));
```
carsonfarmer, line 39: This is the correct thing to do!

packages/mcp/src/index.ts
```diff
@@ -0,0 +1,174 @@
...
+}
+
+main().catch((error) => {
```
carsonfarmer: If possible, I'd rather use a top-level await within try/catch. But that might not work depending on the node version?

carsonfarmer (approved):
Right on! Made a few comments, but nothing blocking.

packages/mcp/src/index.ts
dtbuchholz: kk i'll try that. the other option is to just add them to the [turbo config](https://github.com/recallnet/js-recall/blob/ee0fa0f26060a522a09c2df8dff7525c556cc3ed/turbo.json#L4), but those seem to be application specific (e.g., Portal, Faucet) and not meant for what we're doing there.

packages/mcp/src/index.ts
dtbuchholz: i'll tal. i didnt wanna change too much from the example MCP impl, but might as well do it right.

packages/mcp/src/index.ts
dtbuchholz: hm so our root package file defines node >=20 as the supported version. we [inherit the `module` definition](https://github.com/recallnet/js-recall/blob/b5833f9c24d8ca184c4d318bede8708247aba402/packages/typescript-config/base.json#L11) as NodeNext. so, it's probably fine? or, if we _do_ want to support older version, we could target Node16 or something—and then alter the tsup config to compile to cjs.

packages/mcp/src/index.ts
dtbuchholz: i dont have a strong preference. i kinda like having both options, so you could choose one or the other: 
```json
{
  "mcpServers": {
    "recall-mcp-server": {
      "name": "Recall MCP",
      "type": "command",
      "command": "npx",
      "args": [
        "-y",
        "@recallnet/mcp",
        "--private-key",
        "0x12345...."
      ],
      "env": {
        "RECALL_PRIVATE_KEY": "0x12345..."
      }
    }
  }
}
```
i debated also adding a config file like a true CLI, but i didnt wanna go too overboard.

dtbuchholz:
made a few changes:
- use `minimist` for CLI arg parsing, and split out the config/setup logic into a separate file.
- use `shx` in the build script instead of inline nodejs chmod. (this is what[ anthropic does](https://github.com/modelcontextprotocol/servers/blob/202f6c5e587b9530bf04e34e80d693a65f5b5bd2/src/sequentialthinking/package.json#L17) in their example servers.)
- remove the inline eslint-disable line about env vars + turbo, and just put it in the eslint config.
- update docs with the claude/cursor setup.
- use `zod` for env var and CLI flag validation—which helps with nice error logging:
```
Error initializing Recall MCP server:

 Configuration validation failed:
  - tools: Invalid tool(s). Each tool must be one of: account.read, account.write, bucket.read, bucket.write or "all". Tools provided: test.asdf

 Configuration validation failed:
  - privateKey: Private key is required. Set RECALL_PRIVATE_KEY env var or pass --private-key

 Configuration validation failed:
  - network: Network must be one of: testnet, localnet
```

packages/mcp/src/index.ts
carsonfarmer: naw, don't change anything. anyone writing agents is probably using pretty new versions of stuff

packages/mcp/src/index.ts
carsonfarmer: oh yeh, makes sense actually. also, if we did decide to expose this using the new mcp streaming transport, then you would likely want to be able to run it independently.

carsonfarmer (approved):

dtbuchholz:
ooh, this is what we should do as a follow up (inspiration: [here](https://www.mcpt.com/server/trieve)):
```
npx @recallnet/mcp add --client <cursor|claude|windsurf>
```
ticket: https://github.com/recallnet/js-recall/issues/153

PR #189, opened 2025-04-02
Refactor Object View Layout

andrewxhill:
**Goal:** Modify the existing object view page layout (`apps/portal`) to allocate more prominent space for the upcoming file preview area, while ensuring existing metadata display remains accessible. Implement GitHub-inspired UI patterns for improved usability.

Screenshot 2025-04-02 at 2 34 18 PM

### Layout Analysis
- The current object viewer (`apps/portal/app/buckets/[address]/_components/object.tsx`) uses a `Card` with a simple grid layout:
  - `CardHeader`: displays object name and actions (delete, download)
  - `CardContent`: 2-column grid (1 column on mobile) displaying:
    - Blob Hash, Recovery Hash, Size, Expiry metrics
    - Metadata JSON display (spans both columns)
- The layout prioritizes metadata display but lacks any content preview area.
- Data is fetched using `useGetObject` hook from the SDK.

### New Design
- Implemented a GitHub-inspired file viewer layout:
  - Filename displayed in breadcrumbs with copy button (not duplicated in content area)
  - Toolbar above file preview showing file size on left and actions on right
  - Added "Share" button to copy object URL to clipboard
  - File preview area as the main focus
  - Collapsible metadata panel for detailed information

### Implementation Details
- Created a new component structure with:
  - A placeholder file preview component (`FilePreviewPlaceholder.tsx`) that will be replaced in Phase 2
  - A collapsible metadata panel component (`MetadataPanel.tsx`) that displays all object metadata in a more compact format
  - Modified the main `object.tsx` to use a vertical layout with:
    - The file preview area at the top (primary focus)
    - The collapsible metadata panel at the bottom (collapsed by default)
- Created utility components for copy and share functionality
- Enhanced breadcrumbs to include the filename with a copy button
- The metadata panel preserves all existing metadata fields (Blob Hash, Recovery Hash, Size, Expiry, and Properties)
- Fixed type issues in the components to ensure compatibility with SDK data formats


apps/portal/app/buckets/[address]/_components/bucket.tsx
```diff
@@ -128,7 +131,18 @@ export default function Bucket({ bucketAddress }: { bucketAddress: Address }) {
...
+                      {part || "\u00A0\u00A0"}
+                      {isObject && (
+                        <CopyButton
+                          value={part}
+                          tooltip="Copy filename"
+                          successMessage="Filename copied to clipboard"
+                          iconClassName="opacity-40 hover:opacity-100"
+                        />
+                      )}
+                    </div>
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/bucket.tsx]: Just using the UI here, seems like there should always be a copy button to the right of the breadcrumb. In the root of the bucket it could copy the bucket id. If we are at least one level deep in the bucket it would copy the path. Seems like the copy button should copy the entire path (or prefix, whatever you want to call it), not just the file name. That would make the display logic more simple.

apps/portal/app/buckets/[address]/_components/bucket.tsx
```diff
@@ -128,7 +131,18 @@ export default function Bucket({ bucketAddress }: { bucketAddress: Address }) {
               <Fragment key={`${index}-${part}`}>
                 <BreadcrumbItem>
                   {index === pathParts.length - 1 ? (
-                    part || "\u00A0\u00A0"
+                    <div className="text-foreground flex items-center gap-2 font-semibold">
+                      <File className="text-primary mr-1 size-4" />
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/bucket.tsx]: I personally don't like the look of this file icon in the last breadcrumb component. It looks too busy. It's also not correct that it displays the `File` icon even if the current object is a folder. Anyways, I'd rather just remove it.

apps/portal/app/buckets/[address]/_components/object.tsx
```diff
@@ -14,19 +14,17 @@ import {
...
-import {
-  Card,
-  CardContent,
-  CardHeader,
-  CardTitle,
-} from "@recallnet/ui/components/card";
+import { Card } from "@recallnet/ui/components/card";
 import { useToast } from "@recallnet/ui/hooks/use-toast";
-import CollapsedStringDisplay from "@recallnet/ui/recall/collapsed-string-display";
+import { cn } from "@recallnet/ui/lib/utils";
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/object.tsx]: This isn't used.

apps/portal/app/buckets/[address]/_components/object.tsx
```diff
@@ -14,19 +14,17 @@ import {
...
 
-import Metric from "@/components/metric";
-import { arrayToDisplay } from "@/lib/convert-matadata";
+import { CopyButton } from "@/components/copy-button";
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/object.tsx:20]: This also isn't used.

apps/portal/app/buckets/[address]/_components/object.tsx
```diff
@@ -104,6 +104,12 @@ export default function Object({
 
   const objectApiUrl = getObjectApiUrl(getChain(chainId));
 
+  // Construct the full portal URL for sharing
+  const getPortalShareUrl = () => {
+    const origin = typeof window !== "undefined" ? window.location.origin : "";
+    return `${origin}${pathname}?${searchParams.toString()}`;
+  };
+
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/object.tsx]: You can remove all this and the associated hooks that were added and simply use `window.location.toString()` as the share url.

apps/portal/app/buckets/[address]/_components/object.tsx
```diff
@@ -122,70 +128,59 @@ export default function Object({
...
-          <CardTitle className="flex items-center gap-4">
-            <File />
-            {name}
+      <Card className="flex flex-col rounded-none">
+        {/* Toolbar - Contains file size and actions */}
+        <div className="flex items-center justify-between border-b p-3">
+          <div className="text-muted-foreground flex items-center gap-4 text-sm">
+            <span>
+              {objectSize.formatted ||
+                `${objectSize.val} ${objectSize.unit || ""}`}
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/object.tsx]: `objectSize.formatted` is always a string so you can just display it and remove the fallback logic.

apps/portal/app/buckets/[address]/_components/object.tsx
```diff
@@ -122,70 +128,59 @@ export default function Object({
...
+            </span>
+            <span>
+              Expires {objectBlockDiff && objectBlockDiff < 0n ? "d" : ""}{" "}
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/object.tsx]: This looks wrong. Should be more like:
```
Expire{objectBlockDiff && objectBlockDiff < 0n ? "d" : "s"}{" "}
```

apps/portal/app/buckets/[address]/_components/object.tsx
```diff
@@ -122,70 +128,59 @@ export default function Object({
...
-          <div className="flex flex-col gap-2 sm:col-span-2">
-            <span className="text-muted-foreground text-xs">Metadata</span>
-            <pre className="text-muted-foreground min-h-12 border p-4 font-mono">
-              {arrayToDisplay(object.metadata)}
-            </pre>
+            <ShareButton
+              url={getPortalShareUrl()}
+              tooltip="Copy portal link to share"
+              successMessage="Portal link copied to clipboard"
+            />
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/object.tsx]: The icon in this button is small compared to the trash and download icons next to it. Will look into that component and leave more comments there.

apps/portal/components/copy-button.tsx
asutula, line 1: `CopyButton` and `ShareButton` are basically identical other than the icon displayed inside. I'd make this a single `CopyButton` component that has two `type`s: `copy` and `share`. Based on the `type` prop value, the icon displayed changes.

apps/portal/components/copy-button.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return (
+    <div
+      title={tooltip}
+      onClick={handleCopy}
+      className={cn("cursor-pointer", className)}
+    >
+      <Copy
+        className={cn("size-4 opacity-20 hover:opacity-100", iconClassName)}
+      />
+    </div>
```
asutula: The props and view hierarchy here could be simplified. The outer div can really do nothing except for provide the `title` prop so the tooltip displays. The real content of this component is the `Copy` component. You should make the props for the whole component extend the prop type of `Copy` and allow all `className` modifications to happen just on the `Copy` element. The props would look like:

```
type CopyButtonProps = {
  value: string;
  tooltip?: string;
  successMessage?: string;
} & HTMLAttributes<SVGElement>;
```
Then users can style the svg any way they want.

asutula (commented):
Looking pretty good. Need to run, but left a handful of comments to get started with. Will continue with another review again asap.

apps/portal/app/buckets/[address]/_components/bucket.tsx
Comment (andrewxhill) [reply apps/portal/app/buckets/[address]/_components/bucket.tsx]: cool, makes sense to always show on the breadcrumb
for what it copies, i was leveraging the github ui logic there. which is only the node...

apps/portal/app/buckets/[address]/_components/object.tsx
Comment (andrewxhill) [reply apps/portal/app/buckets/[address]/_components/object.tsx]: Should that be a rule in prettier / lint?

Comment (asutula) [reply apps/portal/app/buckets/[address]/_components/object.tsx]: It is actually. We just need to update the portal package.json `lint` script to be `next lint --max-warnings 0` and that will cause lint to return exit code 1 if there are any warnings. This is the way all our other packages are configured.

Comment (andrewxhill) [reply apps/portal/app/buckets/[address]/_components/object.tsx]: added the flag

apps/portal/components/copy-button.tsx
andrewxhill: okay, i think i got all of 'em

apps/portal/components/copy-button.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { cn } from "@recallnet/ui/lib/utils";
+
+// Extend SVGProps for direct SVG element attributes
+type CopyButtonProps = {
+  value: string;
+  type?: "copy" | "share";
+  tooltip?: string;
+  successMessage?: string;
+  // We still need a wrapperClassName for the tooltip span
+  wrapperClassName?: string;
```
asutula: There is no reason to let the caller style the wrapper, just remove this prop, we don't use it anyway.

apps/portal/components/copy-button.tsx
```diff
@@ -0,0 +1,65 @@
...
+      title={tooltip}
+      onClick={handleCopy}
+      className={cn(
+        "inline-flex cursor-pointer items-center justify-center",
+        wrapperClassName,
+      )}
+    >
+      <Icon
+        className={cn(
+          "size-4 opacity-20 hover:opacity-100", // Base styles for the icon
```
asutula: I would just let the icon be its natural size by default and remove this `size-4`. Caller can always customize it if needed.

apps/portal/app/buckets/[address]/_components/bucket.tsx
Comment (asutula) [reply apps/portal/app/buckets/[address]/_components/bucket.tsx]: See your thumbs up here, but the file icon persists. Get rid of it?

apps/portal/app/buckets/[address]/_components/bucket.tsx
```diff
@@ -120,15 +119,34 @@ export default function Bucket({ bucketAddress }: { bucketAddress: Address }) {
...
             {pathParts.map((part, index) => (
               <Fragment key={`${index}-${part}`}>
                 <BreadcrumbItem>
                   {index === pathParts.length - 1 ? (
-                    part || "\u00A0\u00A0"
+                    <div className="text-foreground flex items-center gap-2 font-semibold">
+                      <File className="text-primary mr-1 size-4" />
+                      {part || "\u00A0\u00A0"}
+                      <CopyButton
+                        value={part}
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/bucket.tsx]: Was thinking about this more and I still think the copy button should copy the whole path. The primary use case of copying would be to use the value in the CLI. The CLI needs the entire path. Is there some other use case I'm not thinking of where just the last part of the path is needed?

apps/portal/app/buckets/[address]/_components/bucket.tsx
```diff
@@ -120,15 +119,34 @@ export default function Bucket({ bucketAddress }: { bucketAddress: Address }) {
...
+                        tooltip={
+                          isObject ? "Copy filename" : "Copy directory name"
+                        }
+                        successMessage={`${isObject ? "Filename" : "Directory name"} copied to clipboard`}
+                        className="opacity-40 hover:opacity-100"
+                      />
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/bucket.tsx:144]: Was still imaging the copy button would live after the `Breadcrumb` component down below here. You could remove this `CopyButton` and the one above, and just have a single `CopyButton` down below after the closure of the `Breadcrumb` component. The value passed into `CopyButton` would depend on if you're viewing the bucket root (the bucket id) or a folder/file (the path). If you do that, you an also remove the `<div>` currently surrounding each `CopyButton`.

apps/portal/app/buckets/[address]/_components/metadata-panel.tsx
```diff
@@ -0,0 +1,107 @@
...
+    recoveryHash: string;
+    size: bigint;
+    expiry: bigint;
+    metadata: readonly {
+      key: string;
+      value: string;
+    }[];
+  };
+  objectSize: {
+    val: string | number;
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/metadata-panel.tsx]: This is always a `number`.

apps/portal/app/buckets/[address]/_components/metadata-panel.tsx
```diff
@@ -0,0 +1,107 @@
...
+    unit: string | undefined;
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/metadata-panel.tsx]: This should always be a `string`. We can pass in `objectSize.unit || ""` into here from the parent component. Upstream of there, we need to update the `formatBytes` function to always return a string for `unit`, but that can be done later.

apps/portal/app/buckets/[address]/_components/metadata-panel.tsx
```diff
@@ -0,0 +1,107 @@
...
+    formatted?: string;
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/metadata-panel.tsx]: This is never used, can be removed.

apps/portal/app/buckets/[address]/_components/metadata-panel.tsx
```diff
@@ -0,0 +1,107 @@
...
+  objectExpiryIso,
+  objectBlockDiff,
+}: MetadataPanelProps) {
+  const [expanded, setExpanded] = useState(false);
+
+  // Convert number val to string if needed
+  const sizeVal =
+    typeof objectSize.val === "number"
+      ? objectSize.val.toString()
+      : objectSize.val;
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/metadata-panel.tsx]: If you make the changes suggested above, you can remove this and access `objectSize.val` directly.

apps/portal/app/buckets/[address]/_components/metadata-panel.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  // Ensure unit is not undefined
+  const sizeUnit = objectSize.unit || "";
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/metadata-panel.tsx]: If you make the changes suggested above, you can remove this and access `objectSize.unit` directly.

apps/portal/app/buckets/[address]/_components/metadata-panel.tsx
```diff
@@ -0,0 +1,107 @@
...
+        className="hover:bg-muted/50 flex cursor-pointer items-center justify-between p-3"
+        onClick={() => setExpanded(!expanded)}
+      >
+        <h3 className="text-sm font-medium">Object Metadata</h3>
+        {expanded ? <ChevronUp size={16} /> : <ChevronDown size={16} />}
+      </div>
+
+      {expanded && (
+        <div className="space-y-4 p-4">
+          <div className="grid grid-cols-2 gap-4">
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/metadata-panel.tsx]: This grid was previously responsive so it would be a single column by default, then use two cols for larger screen sizes.

asutula (commented):
Looking good again. I think this is all my feedback, address these comments and we're good to go.

apps/portal/app/buckets/[address]/_components/bucket.tsx
Comment (andrewxhill) [reply apps/portal/app/buckets/[address]/_components/bucket.tsx]: sure, was just going by the github norm. will switch

Comment (andrewxhill) [reply apps/portal/app/buckets/[address]/_components/bucket.tsx]: there is a funny little quirk herein the UI, since you'd never want the bucket id copied in that path. so then if you get down to just the bucket it will flip to letting you copy the id

apps/portal/app/buckets/[address]/_components/bucket.tsx
Comment (andrewxhill) [reply apps/portal/app/buckets/[address]/_components/bucket.tsx:144]: okay. played with this. struggled with line wrapping. if on its own it has its own wrapping and linebreak behavior. i could fix the linebreak behavior but wrapping remained an issue. ideally it would only wrap _with_ the last element of the breadcrumb, that's just not what was happening. perhaps improving the breadcrumb could be a follow-up ticket

apps/portal/app/buckets/[address]/_components/bucket.tsx
Comment (andrewxhill) [reply apps/portal/app/buckets/[address]/_components/bucket.tsx]: funny, did another prompt after removing it and it added it back a second time. cursor really wanted the file icon

apps/portal/app/buckets/[address]/_components/metadata-panel.tsx
Comment (andrewxhill) [reply apps/portal/app/buckets/[address]/_components/metadata-panel.tsx]: okay, got them both simple. i think i interpreted your idea right, but lmk

apps/portal/app/buckets/[address]/_components/bucket.tsx
Comment (andrewxhill) [reply apps/portal/app/buckets/[address]/_components/bucket.tsx:144]: besides if this is okay, i think the rest is gtg

asutula (approved):

PR #197, opened 2025-04-10
feat: agent toolkit logic tweaks

dtbuchholz:
- alter the add/get object methods to _only_ work with string values input/output values. 
   - e.g., the types were too flexible when adding or getting an object (Uint8Array, File, or string/path), and agents were getting confused on how to write or read data.
- move function call definitions into the `tools` vs handling logic in the `RecallAPI` (this also makes each tool's function more easily).
- include `getTools` on the Recall API so that agent toolkit instances can use it directly when setting up tools.
- allow `prompts` and `parameters` to accept the `context` arg. 
   - we don't do anything with this yet, but it'll set us up for future features, like injecting bucket aliases or competition / agent config data into an agent's prompt.
    ```ts
    export const getOrCreateBucketPrompt = (_context: Context = {}) => `
    Gets an existing bucket by alias, or creates a new one if it doesn't exist. Use this to ensure you have a bucket for storage.
    
    // add some logic to inject the alias as a predefined param, or tell the agent to create their own if one isnt provided
    ...
    ```

- update prompts with some language tweaks, which includes describing the return type shape.
- allow the `context` to be optional. currently, the only thing we use here is the `network`, and we already handle setting a default if the value isnt present. now, the main difference is that you can just omit the `context` key:
    ```ts
    const toolkit = new RecallAgentToolkit({
      privateKey: "0x...",
      configuration: {
        actions: {
          account: {
            read: true,
            write: true,
          },
        },
        context: {} // this is no longer required, so you can remove `context` and just pass `actions`
      },
    });
    ```
- fixup some examples
- update ai-related deps to the the latest
- add more docstrings

### DIFF
--- a/packages/agent-toolkit/src/shared/functions.ts
+++ b/packages/agent-toolkit/src/shared/functions.ts
@@ -27,7 +27,7 @@ import { Result } from "./util.js";
 export const getAccountInfo = async (
   recall: RecallClient,
   context: Context,
-  params: z.infer<typeof getAccountInfoParameters>,
+  params: z.infer<ReturnType<typeof getAccountInfoParameters>>,
 ): Promise<Result<AccountInfo>> => {
   try {
     const address = params.address ? (params.address as Address) : undefined;
@@ -57,7 +57,7 @@ export const getAccountInfo = async (
 export const listBuckets = async (
   recall: RecallClient,
   context: Context,
-  params: z.infer<typeof listBucketsParameters>,
+  params: z.infer<ReturnType<typeof listBucketsParameters>>,
 ): Promise<Result<ListResult>> => {
   try {
     const address = params.address ? (params.address as Address) : undefined;
@@ -87,7 +87,7 @@ export const listBuckets = async (
 export const getCreditInfo = async (
   recall: RecallClient,
   context: Context,
-  params: z.infer<typeof getCreditInfoParameters>,
+  params: z.infer<ReturnType<typeof getCreditInfoParameters>>,
 ): Promise<Result<CreditAccount>> => {
   try {
     const address = params.address ? (params.address as Address) : undefined;
@@ -117,7 +117,7 @@ export const getCreditInfo = async (
 export const buyCredit = async (
   recall: RecallClient,
   context: Context,
-  params: z.infer<typeof buyCreditParameters>,
+  params: z.infer<ReturnType<typeof buyCreditParameters>>,
 ): Promise<Result<string>> => {
   try {
     const to = params.to ? (params.to as Address) : undefined;
@@ -146,7 +146,7 @@ export const buyCredit = async (
 export const createBucket = async (
   recall: RecallClient,
   context: Context,
-  params: z.infer<typeof createBucketParameters>,
+  params: z.infer<ReturnType<typeof createBucketParameters>>,
 ): Promise<Result<{ bucket: Address; txHash: string }>> => {
   try {
     const metadata = params.metadata ?? {};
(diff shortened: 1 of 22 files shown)


packages/agent-toolkit/src/shared/functions.ts
```diff
@@ -260,25 +278,26 @@ export const addObject = async (
...
 
-    if (!result) {
+    if (!object) {
       return { success: false, error: "Object not found" };
     }
-
-    return params.outputType === "string" || params.outputType === undefined
-      ? { success: true, result: new TextDecoder().decode(result) }
-      : { success: true, result };
+    const result = new TextDecoder().decode(object);
```
dtbuchholz, line 299: note: we just assume data is always a string...since that's what agents/llms should be writing to their buckets, anyways

packages/agent-toolkit/src/shared/parameters.ts
```diff
@@ -9,123 +11,127 @@ const addressSchema = z.string().regex(/^0x[a-fA-F0-9]{40}$/);
...
-  bucketAlias: z.string().min(1).describe("The alias of the bucket to create"),
-  metadata: z
-    .record(
-      z.string().min(1).describe("The key of the metadata"),
-      z.string().min(1).describe("The value of the metadata"),
-    )
-    .optional()
-    .describe("The metadata to store with the bucket"),
-});
+export const createBucketParameters = (_context: Context = {}) =>
```
dtbuchholz, line 65: allow `Context` to get passed...tbd on if/what we use, but "global" bucket aliases are a good example.

packages/agent-toolkit/src/shared/tools.ts
```diff
@@ -39,114 +53,130 @@ export type Tool = {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   parameters: z.ZodObject<any, any, any, any>;
   actions: Actions;
+  execute: (
```
dtbuchholz, line 56: added this `execute` method so that you can grab all `tools` as an export and more easily build your own adapter with everything included.

packages/agent-toolkit/src/shared/parameters.ts
```diff
@@ -9,123 +11,127 @@ const addressSchema = z.string().regex(/^0x[a-fA-F0-9]{40}$/);
...
-  overwrite: z
-    .boolean()
-    .optional()
-    .describe("Whether to overwrite existing data at that key"),
-});
+export const addObjectParameters = (_context: Context = {}) =>
+  z.object({
+    bucket: addressSchema.describe("The address of the bucket"),
+    key: z.string().min(1).describe("The key under which to store the object"),
+    data: z.string().describe("The data to store as a string value"),
```
dtbuchholz, line 111: now, we just assume that when you add an object, it's a string (instead of allowing uint8arrays or files, which confused the model)

packages/agent-toolkit/src/shared/functions.ts
mzkrasner: makes sense for cot for sure! though i wonder long term if we might want to allow llms to write images/files or other forms of media to buckets as well in the future

dtbuchholz: forsure. i'll open a separate PR...i guess the way we'll handle it is just with separate methods like `addFile` (uint8array). or maybe also `addImage` (base64 encoded string? idk)

mzkrasner: sure that works. was going to say maybe we should add a note about the current support limitations in the interim, but that will be quickly outdated once the separate pr is merged

mzkrasner (approved):
Super slick. Tested these changes locally with claude and works seamlessly. Only thing I'd add is maybe adding a note in the README about the string-only handling, but if we plan on submitting another PR for other data types, then it's probably not necessary

PR #199, opened 2025-04-10
UI upgrade

asutula:
Getting things prepared for Protofire and the new competitions app. This PR upgrades our UI code and package to Tailwind v4, and the updates Shadcn/ui components to their Tailwind v4 versions. I also refactored the folder structure of the ui package so it more easily supports shadcn components as well as our own components that don't use shadcn (I suspect new components that we build for the competitions app won't use shadcn). Lots of changes in the diff here, but it's mostly moving files around, plus a few fixes for the new version of Tailwind.


dtbuchholz (approved):

asutula:
Cool looking into those @dtbuchholz

asutula:
Fixed all those issues @dtbuchholz, thanks.

Issue #202, opened 2025-04-11
Fix gas estimation and query logic


dtbuchholz:
We noticed two problems:
1. Gas estimation was wayyyy too high—at least 3000x the actual value. For example, adding an object had an effective gas price of 25000 gwei, while the Rust client always uses 8 gwei.
2. As discovered in #201, the wrappers cant handle queries with over 50 objects at a time. Thus, we need to have logic to cap query results and then batch them into multilpe requests, as needed.

PR #203, opened 2025-04-11
feat: fix gas estimation & query limit issues

dtbuchholz:
## Summary

closes https://github.com/recallnet/js-recall/issues/202 https://github.com/recallnet/recall-alpha-agent/issues/7. 

## Details

for the issue where we're burning through RECALL tokens, all we need to do is manually make the `getGasPrice` call before every write operation, and then pass it as an option to the method call:
```ts
const gasPrice = await this.client.publicClient.getGasPrice()

const hash = await this.client.walletClient?.sendTransaction({
  account: this.client.walletClient.account,
  chain: this.client.walletClient.chain,
  to: recipient,
  value: amount,
  gasPrice,
});
```

for the query objects issue, we do the following:
1. set a MAX_QUERY_LIMIT to 50 objects because if you try to make a query larger than that, there's a high likelihood it'll fail (out of gas issues due to onchain processing)
2. we call the bucket manager's query object method one time per 50 objects. e.g., if you want to query for 100 objects in a bucket, we'll automatically make 2 requests in 50 object increments.

note: when we move from wrappers to the facades, we should revisit what the effective query limit is since it'll be much cheaper.

### Gas issues

The observed effective gas rate per tx dropped from `25000` to `7`. This more closely aligns with what we see in the rust clients where the gas price is `8`.
```
// before changes, you'd use up a minimum of 0.0001 RECALL, as high as 0.6
Balance before 3.016125347138601077
Gas used 158711044n
Gas price 250011n // sometimes as high as 3750220406n
ETH used 0.000039679506821484 // sometimes as high as 0.6 RECALL
Balance after 3.016085667574953021

// after, you use a negligible amount
Balance before 0.328464769145263858
Gas used 176823074n
Gas price 7n
ETH used 0.000000001237761518
Balance after 0.328464767861086285
```

### Query objects

- The existing `sdk` tests pass. 
- It takes too long to wait to add >50 objects, and then query them. We _can_ add this as a test, though.
- Fixes my trading eval scaper, which is where we uncovered this bug.
- Also, I manually ran some experiments. For a bucket with 82 objects:
   1. query for <=50 objects (default)
   2. query for 80 objects (i.e., 2 "leftover")
   4. query for 100 objects (all of them)
- The main difference in the scenarios above is that (1) and (2) will show a value like `nextKey: 'competition/v0/cot/4-uw99f00yqca'`, which is collected during the batch processing—and (3) does not (since there's nothing left).

### DIFF
--- a/packages/network-constants/src/index.ts
+++ b/packages/network-constants/src/index.ts
@@ -163,6 +163,11 @@ export const MIN_TTL = 3600n; // one hour
 export const MAX_OBJECT_SIZE = 5_000_000_000; // 5GB
 
 /**
+ * Maximum number of objects that can be returned in a single query (to avoid hitting gas limits).
+ */
+export const MAX_QUERY_LIMIT = 50;
+
+/**
  * Gateway Manager Facet contract address for the Testnet parent chain.
  */
 export const TESTNET_PARENT_GATEWAY_MANAGER_FACET_ADDRESS =
--- a/packages/sdk/src/entities/bucket.ts
+++ b/packages/sdk/src/entities/bucket.ts
@@ -18,7 +18,11 @@ import {
   bucketManagerAddress,
   iMachineFacadeAbi,
 } from "@recallnet/contracts";
-import { MAX_OBJECT_SIZE, MIN_TTL } from "@recallnet/network-constants";
+import {
+  MAX_OBJECT_SIZE,
+  MAX_QUERY_LIMIT,
+  MIN_TTL,
+} from "@recallnet/network-constants";
 
 import { RecallClient } from "../client.js";
 import {
@@ -28,6 +32,7 @@ import {
   CreateBucketError,
   InvalidValue,
   ObjectNotFound,
+  OutOfGasError,
   UnhandledBucketError,
   isActorNotFoundError,
 } from "../errors.js";
@@ -96,14 +101,32 @@ export type QueryResultRaw = ContractFunctionReturnType<
   ]
 >;
 
+// Object value in query response (converted `metadata` to normal javascript object)
+export type QueryObjectValue = Omit<
+  QueryResultRaw["objects"][number]["state"],
+  "metadata"
+> & {
+  metadata: Record<string, unknown>;
+};
+
+// Object in query response
+export type QueryResultObject = {
+  key: QueryResultRaw["objects"][number]["key"];
+  state: QueryObjectValue;
+};
+
 // Converts metadata Solidity struct to normal javascript object
 export type QueryResult = Omit<QueryResultRaw, "objects"> & {
-  objects: {
-    key: QueryResultRaw["objects"][number]["key"];
-    state: Omit<QueryResultRaw["objects"][number]["state"], "metadata"> & {
-      metadata: Record<string, unknown>;
-    };
-  }[];
+  objects: QueryResultObject[];
+};
+
+// Options for query()
+export type QueryOptions = {
+  prefix?: string;
+  delimiter?: string;
+  startKey?: string;
+  limit?: number;
+  blockNumber?: bigint;
 };
 
 // Note: this event stems from the underlying machine facade contract; creating a bucket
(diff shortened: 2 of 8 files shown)


dtbuchholz:
@asutula i didnt look at the Portal/frontend code, but it might need to do something similar as what i did for the sdk:
- paginate objects (it looks like you _might_ do this already) when you query objects in a bucket
- explicitly set the gas price when you execute txs

also cc @mzkrasner this should fix the RECALL token issues we see

packages/sdk/src/entities/bucket.ts
```diff
@@ -566,10 +648,18 @@ export class BucketManager {
       return { result };
     } catch (error: unknown) {
       if (error instanceof ContractFunctionExecutionError) {
-        // TODO: We're optimistically assuming an error means the bucket doesn't exist
-        // 00: t0134 (method 3844450837) -- contract reverted (33)
-        // 01: t0134 (method 6) -- contract reverted (33)
-        throw new BucketNotFound(bucket);
+        if (error.message.includes("contract reverted")) {
+          const isOutOfGasError =
+            error.message.includes("wasm `unreachable` instruction executed") ||
```
dtbuchholz, line 653: meta observation: the fvm/wasm errors can be a bit too ambiguous. we'll see something like this if you query a bucket and it doesnt exist on the network, which is non-descriptive:
```
00: t0134 (method 3844450837) -- contract reverted (33)
01: t0134 (method 6) -- contract reverted (33)
```
but an out of gas issue will have an extra line that at least helps a bit:
```
02: t0134 (method 6) -- out of gas (7)
```
i.e., the js client has to make some assumption to help handle these errors nicely for the dev.

packages/sdk/src/entities/bucket.ts
```diff
@@ -530,21 +561,72 @@ export class BucketManager {
...
+
+    // If within MAX_QUERY_LIMIT, use a single query
+    if (requestedLimit <= MAX_QUERY_LIMIT) {
+      return this.executeQuery(bucket, options);
+    }
+
+    // Otherwise, paginate and collect results. Note: this is needed to avoid
+    // hitting gas limits and failing to retrieve all objects.
+    try {
+      const allObjects: Array<QueryResultObject> = [];
```
dtbuchholz, line 576: for reference, here's where we batch `query` calls automatically. since the query limit is so low (50 objects), i figured it makes sense to do this automatically vs adding some flag or another method for batching multiple queries.

asutula:
> i didnt look at the Portal/frontend code, but it might need to do something similar as what i did for the sdk:

Hey @dtbuchholz have you ever seen gas problems with the portal? I haven't, but maybe haven't been pushing it in the same ways you have. 

The portal does paginate objects. It does not explicitly set gas price when sending txns.

packages/network-constants/src/index.ts
```diff
@@ -162,6 +162,11 @@ export const MIN_TTL = 3600n; // one hour
  */
 export const MAX_OBJECT_SIZE = 5_000_000_000; // 5GB
 
+/**
+ * Maximum number of objects that can be returned in a single query (to avoid hitting gas limits).
+ */
+export const MAX_QUERY_LIMIT = 50;
+
```
asutula, line 169: We can incorporate this into our query react hook.

asutula (approved):
With the caveat that I'm not super familiar with the gas issues within the sdk, looks good to me. Not sure if changes are needed in the portal since I've never experienced gas troubles there.

dtbuchholz:
> > i didnt look at the Portal/frontend code, but it might need to do something similar as what i did for the sdk:
> 
> Hey @dtbuchholz have you ever seen gas problems with the portal? I haven't, but maybe haven't been pushing it in the same ways you have.
> 
> The portal does paginate objects. It does not explicitly set gas price when sending txns.

oddly, no. i've only seen it with the programmatic tooling. maybe it happens when an account quickly adds objects to a bucket one after another...which isnt really possible to do fast enough in the frontend?

also, somewhat related, check out [this issue](https://github.com/recallnet/rust-recall/issues/264) re: how query results are returned to a user. TL;DR—queried objects are returned based on the HAMT's representation (e.g., not lexicographical / chronological). to improve the DX, _i think_ we'll have to alter the SDK to query _all_ objects + sort/filter so that data returned to users isn't confusing. that'll reuse some of this PRs logic, too.

PR #210, opened 2025-04-16
Competitions app + ui2 package

asutula:
Creating a place for the new web app.


dtbuchholz (approved):

asutula:
FYI @dtbuchholz after your review, I added a new package called `ui2`. It's just like our existing `ui` package, but we can use it to build the new component library for the competitions app or any new UI work based on Wednesday's designs.

PR #224, opened 2025-04-23
Placeholder comps-db package

asutula:
This PR creates a simple `comps-db` package that we can use as a starting point for further working in defining our data model and eventually our APIs.


ramirotw (approved):
looks good

Issue #225, opened 2025-04-23
Wallet connection and backend auth


asutula:
In the `comps` app, let's get a first version of wallet connection and auth with the backend going. I think we will learn a bit during this process, it will prompt some discussions, and we can adjust things if needed. I'm open to discussion about what choices we make, but I'll just suggest that we:

- [ ] Integrate  [Coinbase Smart Wallet](https://docs.base.org/identity/smart-wallet/quickstart/nextjs-project) into the project
- [ ] Create a simple UI that allows us to create/connect our Coinbase smart wallet (this doesn't need to look good at all)
- [ ] Crate a `/login` API route that we can send a SIWE payload to for login. This endpoint should just verify the SIWE payload for now. We can save user data in the database at a later time.
- [ ] Set up the client code to, upon successful wallet connection, create and sign the SIWE payload and then call the `/login` endpoint.

PR #227, opened 2025-04-23
Rafael/wallet connect auth

Rcontre360:
#225 

Just added some basic auth with the tools requested. Currently the coinbaseWallet config already handles the wallet creation with the window used for signing.

I think next steps with this would be to add JWT and validate the nonce properly. For this we will need the db and db schema. Also added an endpoint to generate a nonce.


Rcontre360:
Closing it. need to see why so many files are being changed on the pr

PR #228, opened 2025-04-23
Rafael/wallet auth

Rcontre360:
https://github.com/recallnet/js-recall/issues/225

Just added some basic auth with the tools requested. Currently the coinbaseWallet config already handles the wallet creation with the window used for signing.

I think next steps with this would be to add JWT and validate the nonce properly. For this we will need the db and db schema. Also added an endpoint to generate a nonce.


apps/comps/package.json
```diff
@@ -16,24 +16,32 @@
...
     "@recallnet/ui2": "workspace:*",
+    "@tanstack/react-query": "^5.64.2",
     "@vercel/analytics": "^1.5.0",
+    "@wagmi/core": "^2.17.0",
+    "axios": "^1.7.9",
     "next": "^15.1.0",
     "react": "^19.0.0",
-    "react-dom": "^19.0.0"
+    "react-dom": "^19.0.0",
+    "siwe": "^3.0.0",
```
ramirotw: do we need siwe? can't we just use https://viem.sh/docs/siwe/utilities/createSiweMessage ?

apps/comps/package.json
```diff
@@ -16,24 +16,32 @@
     "docs:serve": "typedoc --watch --serve"
   },
   "dependencies": {
+    "@coinbase/wallet-sdk": "^4.3.2",
     "@recallnet/comps-db": "workspace:*",
+    "@coinbase/wallet-sdk": "^4.3.2",
```
ramirotw: @coinbase/wallet-sdk is duplicated

apps/comps/app/api/login/route.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  if (msg.domain != host) {
+    throw new Error("Invalid domain")
+  }
+  if (msg.chainId != 84532) {
+    throw new Error("Invalid chain id")
+  }
+  if (msg.nonce != '' && false) { // we should get nonce from some db
+    throw new Error("Invalid nonce")
+  }
```
ramirotw, line 23: about the nonce, what we can do for now is something similar to https://github.com/spruceid/siwe-quickstart/blob/main/03_complete_app/backend/src/index.js which saves the generated nonce in a cookie and then we verify the signed message against that nonce (https://viem.sh/docs/siwe/actions/verifySiweMessage#nonce-optional)

ramirotw (changes requested):
looking good, a few comments:

* need to run prettier to the files
* need to discuss about sessions

apps/comps/package.json
Rcontre360: thanks!

apps/comps/app/api/login/route.ts
asutula: So just to clarify, in the api we will need a `nonce` endpoint that is called by the client before constructing the SIWE payload on the client side. The endpoint stores the nonce encrypted in a cookie and also returns the nonce to the client. The client uses that nonce in the payload, then calls the `login` endpoint. The `login` endpoint decrypts the cookie to get the nonce, and verifies the SIWE payload with that nonce.

asutula: Ah cool, I see the nonce enddpoint below now, cool.

apps/comps/app/api/nonce/route.ts
```diff
@@ -0,0 +1,27 @@
...
+  try {
+    const nonce = generateSiweNonce();
+
+    const cookie = serialize("session", nonce, {
+      httpOnly: true,
+      secure: true,
+      maxAge: 60 * 60 * 24 * 7, // One week
+      path: "/",
+    });
+
```
ramirotw, line 15: why not `import { cookies } from 'next/headers'` instead of a new dep?
https://nextjs.org/docs/app/api-reference/functions/cookies

apps/comps/app/api/login/route.ts
```diff
@@ -0,0 +1,62 @@
...
+      );
+    }
+
+    const siweMessage = parseSiweMessage(message) as SiweMessage;
+    await validateMessage(req, siweMessage);
+
+    const result = await verifyMessage(config, {
+      address: siweMessage.address as `0x${string}`,
+      message: message,
+      signature,
```
ramirotw, line 50: shouldn't we also validate the nonce from the cookie?

ramirotw (changes requested):

asutula (changes requested):
Ok this is a good start that taught us a few things about how auth will work. I suggest we resolve the lint problems and there is also a problem with the `pnpm-lock.yml` that will be resolved by running `pnpn install` in the root of the repo. Then we can merge that. Next step will be to build the navigation bar that contains the "login" button and uses the client side code you now have to connect the user's wallet. Joe or I can take what you've prototyped here for the server side and move it over to the API package. Sound good?

Rcontre360:
> Not sure what editor you use, but if you use VSCode or Cursor, install the ESLint plugin and you will see all the errors. You can also just run `pnpm lint` in the reop root and see the errors.

I use vim. Let me check with vscode to see the issues. Vim is sometimes annoying with these things, if it keeps failing I'll switch to vscode

Rcontre360:
> Ok this is a good start that taught us a few things about how auth will work. I suggest we resolve the lint problems and there is also a problem with the `pnpm-lock.yml` that will be resolved by running `pnpn install` in the root of the repo. Then we can merge that. Next step will be to build the navigation bar that contains the "login" button and uses the client side code you now have to connect the user's wallet. Joe or I can take what you've prototyped here for the server side and move it over to the API package. Sound good?

Sounds good! If needed, I can take care of moving it to the API package

packages/agent-toolkit/src/mcp/toolkit.ts
```diff
@@ -64,10 +63,7 @@ export default class RecallAgentToolkit extends McpServer {
         tool.method,
         tool.description,
         tool.parameters.shape,
-        async (
-          arg: z.infer<typeof tool.parameters>,
-          _extra: RequestHandlerExtra,
-        ) => {
+        async (arg: z.infer<typeof tool.parameters>) => {
```
asutula: cc @mzkrasner just to make sure this is an ok change.

asutula (approved):

packages/agent-toolkit/src/mcp/toolkit.ts
mzkrasner: Even though the RequestHandlerExtra var was unutilized, I'd think including it in the function signature might still be a good idea as it would make the toolkit more flexible (for example, future updates might make use of it)

Rcontre360: Added types again here https://github.com/recallnet/js-recall/pull/228/files#diff-2f63303dc14ff33e9e50e43fc2fcf75f38f71a59a7643bf30372354354b6143f

apps/comps/tsconfig.json
```diff
@@ -19,5 +26,8 @@
     "**/*.tsx",
     ".next/types/**/*.ts"
   ],
-  "exclude": ["node_modules"]
+  "exclude": [
+    "node_modules",
+    "wagmi-config.ts"
```
asutula: Why are you excluding this file?

Rcontre360: I was getting weird typescript errors before. Now it seems to be resolved (pushing changes)

PR #229, opened 2025-04-24
Integrate api package into monorepo, step 1

asutula:
- Uses shared ts and eslint configs
- Removes dependencies provided by those shared configs
- Fixes all eslint warnings and errors introduces by the new config
- Updates `package.json` to be more consistent with other repo packages


.prettierignore
```diff
@@ -8,4 +8,5 @@ packages/ui/src/styles
 packages/ui2/src/components/shadcn
 packages/ui2/src/hooks
 packages/ui2/src/lib
-packages/ui2/src/styles
\ No newline at end of file
+packages/ui2/src/styles
+apps/api
```
asutula, line 12: Ignoring the api package for now, but will remove this line and let prettier do it's thing in the next pr. Just want to consolidate pure formatting changes to a single PR.

apps/api/.prettierignore
asutula, line 1: Prettier is simply configured and managed at the root of the repo.

apps/api/eslint.config.mjs
asutula, line 1: Extend the shared esling config to have similar settings to the rest of the repo.

apps/api/jest.config.ts
asutula, line 1: This as typescript plays better with the updated typescript settings that are inherited from the repo shared config.

apps/api/package.json
```diff
@@ -1,14 +1,13 @@
 {
-  "name": "trading-simulator",
+  "name": "@recallnet/api",
```
asutula, line 2: New package name.

apps/api/package.json
```diff
@@ -1,14 +1,13 @@
...
   "version": "1.0.0",
   "description": "Solana Trading Simulator Server for hosting trading competitions",
   "main": "dist/src/index.js",
   "scripts": {
-    "lint": "eslint .",
+    "lint": "eslint . --max-warnings 0",
```
asutula, line 7: Don't allow warnings. Eslint will now exit with code 1 if there are any warnings. All our packages are configured this way.

apps/api/package.json
```diff
@@ -1,14 +1,13 @@
...
     "lint:fix": "eslint . --fix",
-    "format": "prettier --write .",
-    "format:check": "prettier --check .",
     "start": "node dist/src/index.js",
+    "clean": "rm -rf .turbo node_modules dist",
```
asutula, line 10: Useful script that exists in all other packages.

apps/api/package.json
```diff
@@ -62,31 +61,24 @@
...
     "swagger-ui-express": "^5.0.1",
     "ts-node": "^10.9.1",
-    "typescript-eslint": "^8.23.0",
     "uuid": "^9.0.0"
   },
   "devDependencies": {
     "@elizaos/core": "^0.25.9",
-    "@eslint/js": "^9.19.0",
+    "@recallnet/eslint-config": "workspace:^",
+    "@recallnet/typescript-config": "workspace:*",
```
asutula, line 69: Depend on shared ts and eslint configs.

apps/api/package.json
```diff
@@ -62,31 +61,24 @@
...
     "axios": "^1.6.0",
-    "eslint": "^9.19.0",
-    "eslint-config-prettier": "^10.0.1",
-    "eslint-plugin-prettier": "^5.2.3",
     "jest": "^29.5.0",
     "openapi-markdown": "^1.2.3",
-    "prettier": "^3.4.2",
     "ts-jest": "^29.1.0",
     "ts-node-dev": "^2.0.0",
-    "typescript": "^5.0.0"
```
asutula, line 90: Remove deps that are provided by the shared configs we now depend on, or things that are managed elsewhere like prettier.

apps/api/scripts/check-db.ts
asutula, line 1: From this file on to the ones below until you see another comment from me, all the changes are lint fixes required by the new eslint settings that are shared in the repo. The rules are a little stricter, but I think they are good.

apps/api/tsconfig.json
```diff
@@ -1,18 +1,15 @@
 {
+  "extends": "@recallnet/typescript-config/base.json",
```
asutula, line 2: Extend the shared config.

apps/api/tsconfig.json
```diff
@@ -1,18 +1,15 @@
...
   "compilerOptions": {
-    "target": "ES2020",
-    "module": "commonjs",
-    "lib": ["es2020"],
```
asutula, line 5: Went ahead and allowed these to come from the shared config. This uses more modern settings, but it didn't cause many problems since you're already writing modern Typescript.

apps/api/tsconfig.json
```diff
@@ -1,18 +1,15 @@
...
-    "strict": true,
-    "esModuleInterop": true,
-    "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
-    "resolveJsonModule": true,
     "experimentalDecorators": true,
     "emitDecoratorMetadata": true,
-    "sourceMap": true
+    "sourceMap": true,
+    "noUncheckedIndexedAccess": false
```
asutula, line 12: This is set to `true` in the shared config, which is the better setting I think. I like that when you access an array by index, the result can be `undefined`. Forces you to deal with that case. So, removing this line will cause us to deal with all that, which is better left for a separate PR.

apps/api/turbo.json
asutula, line 1: This app has lots of env vars! Turbo needs to know about them so it uses their values as input for cache key calculation. Need to spend a bit more time with this file to make sure things are optimal, but this is ok for now.

.prettierignore
mzkrasner: Cool - makes sense

apps/api/package.json
mzkrasner: 100%

mzkrasner (approved):
Everything makes sense - appreciate the detailed comments!

PR #238, opened 2025-04-24
Enable prettier on api package, format fixes

asutula:
Just using our repo rules for now, we can always discuss and update them later if there are any serious objections to this.


mzkrasner (approved):
lgtm

PR #239, opened 2025-04-24
fix: remove html error responses

mzkrasner:
# Summary

- Somehow with the turbo-related changes some of my error responses were defaulting to html
- Fixed by standardizing in error handler

### DIFF
--- a/apps/api/src/middleware/errorHandler.ts
+++ b/apps/api/src/middleware/errorHandler.ts
@@ -1,4 +1,4 @@
-import { Request, Response } from "express";
+import { NextFunction, Request, Response } from "express";
 
 /**
  * Custom error class with HTTP status code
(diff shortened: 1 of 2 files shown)


apps/api/src/middleware/errorHandler.ts
```diff
@@ -17,7 +17,13 @@ export class ApiError extends Error {
 /**
  * Global error handler middleware
  */
-const errorHandler = (err: Error | ApiError, req: Request, res: Response) => {
+const errorHandler = (
+  err: Error | ApiError,
+  req: Request,
+  res: Response,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  next: NextFunction,
```
asutula, line 25: Oh yea I had removed this last arg because it wasn't used and just assumed it would work. Guess not. Sorry.

asutula:
Don't worry about that vercel deploy failure for now. It's erroring out because the comps app is using the free plan on Neon for a db.

PR #242, opened 2025-04-24
Mzk/update workflow

mzkrasner:
# Summary

- Moves original api CI workflow into api-ci in correct directory
- Updates test format with wrappers and path resolution to enable execution and exit


asutula (approved):
Perfect, basically exactly what I had haha.

Issue #223, opened 2025-04-23
Drizzle table definitions


asutula:
We have an existing database schema defined in another repo [here](https://github.com/recallnet/trading-simulator/blob/main/src/database/init.sql). 

To begin with, it would be useful to move that definition into a package within the monorepo and create the definitions with Drizzle (I prefer this tool because it is programmatic, type safe, and has great support for migrations).

Some of the table definitions there are related to teams, agents, and competitions, while others are focused on the trading competition specifically. I suggest we separate those two things by using separate postgres schemas for each one. We can also keep the drizzle table definitions separate/modular in order to not have too much code in one place.

Once this work is done, we can decide exactly how we want to build api endpoints on top of this database layer. 

I started on a `comps-db` package in [this pr](https://github.com/recallnet/js-recall/pull/224). Let's get that merged and you can use it as a starting point @ramirotw.

ramirotw:
@asutula this is the final ER diagram with the drizzle schema from the comps-db package and the merged trading-sim api schema combined.

I'm trying to understand the modeling of users and agents. In the current schema everything is modeled under the `teams` entity, why not a team->agents relationship? a few more questions:
* apiKey: shouldn't the API key be set for each agent?
* walletAddress: wallet address here is confusing. It can represent the user's wallet address for authentication, and also the agent's wallet address for the trading competition. A user should be able to use their own wallet to manage all their agents, and (this is the part I'm not 100% sure), each agent could have it's own wallet address. An agent having its own wallet address makes sense for a trading agent, but does it for other types of agents?

![Image](https://github.com/user-attachments/assets/e869deb2-ad7a-474f-afef-159356536f45)

asutula:
Nice, very helpful diagram! These are good questions, and I think @mzkrasner should chime in as well, but here are my thoughts:

- Yea it seems there are some confusing terminology things happening here - "Team" is actually a single agent or a single non-agent user (like in the case of an admin user using an admin api key for managing competitions). There really is no concept of a team in the way that work is normally used (a group of users collaborating). Is that correct @@mzkrasner?
- That confusion means that, yes, there should be a separate "agents" table. Seems renaming "teams" to "users" would be useful as well. One user can have many agents.
- Assuming the above change is made, it does make sense for users to have a wallet address for auth.
- The core agents model shouldn't need wallet address.
- If the existing trading type of competition requires a wallet address for agents, the trading competition module/logic can introduce a separate table to store wallet addresses for agents.

[...]

PR #244, opened 2025-04-25
[WIP] Update comps-db schema

ramirotw:
https://github.com/recallnet/js-recall/issues/223

This is a work in progress PR which will be updated upon having more definitions over a stable schema. It basically includes the Drizzle changes based on the trading API schema.


ramirotw:
PR changes were included in https://github.com/recallnet/js-recall/pull/289

Issue #219, opened 2025-04-22
Browse/join competitions flow


asutula:
- [ ] Display a list of competitions categorized by active/inactive, upcoming, past
- [ ] Link to competition page

PR #251, opened 2025-04-29
feat: add competitions basic layout and components

ramirotw:
https://github.com/recallnet/js-recall/issues/219

This PR aims to add the basic components used on the Competitions page. Keep in mind that:

1. The displayed data is hardcoded. Once progress is made with the data model, we can start mocking with more realistic values.
2. There is no behavior yet in any component, nor are there any hooks.
3. Several things still need to be polished, it’s not pixel-perfect according to the Figma design


packages/ui2/src/components/button.tsx
```diff
@@ -1,30 +1,54 @@
 import { type VariantProps, cva } from "class-variance-authority";
 
+import { cn } from "../lib/utils.js";
```
asutula: This is causing ts compilation error and should be imported like this:
```
import { cn } from "@recallnet/ui2/lib/utils"
```
This is a little surprising to me that what you have doesn't work, but I think it has to do with how we are transpiling the ui2 code from the web app and how we have the tsconfig path remapping setup in ui2. I copied this setup from shadui, so it must not be too bad :)

packages/ui2/src/components/input.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from "react";
+
+import { cn } from "../lib/utils.js";
```
asutula: Same: `import { cn } from "@recallnet/ui2/lib/utils"`

packages/ui2/src/components/table.tsx
```diff
@@ -0,0 +1,106 @@
+import * as React from "react";
+
+import { cn } from "../lib/utils.js";
```
asutula: Same here

```import { cn } from "@recallnet/ui2/lib/utils"```

File: packages/ui2/src/components/tabs.tsx [hunk 4]
```diff
@@ -0,0 +1,43 @@
+import * as TabsPrimitive from "@radix-ui/react-tabs";
+import * as React from "react";
+
+import { cn } from "../lib/utils.js";
```
asutula: And here:

```
import { cn } from "@recallnet/ui2/lib/utils"
```

packages/ui2/src/components/input.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export interface InputProps
+  extends React.InputHTMLAttributes<HTMLInputElement> {}
+
+const Input = React.forwardRef<HTMLInputElement, InputProps>(
```
asutula: In all the components you created in this PR, no need to use `React.forwardRef` now because we are using React 19:

https://react.dev/blog/2024/12/05/react-19#improvements-in-react-19

asutula (commented):
Looking good so far. Left some comments to get the build working and a note about `forwardRef`.

packages/ui2/src/components/input.tsx
asutula: The button component I created and you modified in this PR doesn't use `forwardRef`, for example.

ramirotw: updated, thanks

packages/ui2/src/components/button.tsx
ramirotw: yeah, the build wasn't working but the dev env was. My editor setup auto imports files without taking into account the tsconfig mapping

apps/comps/app/competitions/page.tsx
```diff
@@ -1,7 +1,42 @@
-export default function Competitions() {
+import React from "react";
+
+import { AgentSpotlightSection } from "../../components/agent-spotlight-section";
```
asutula: Not a big deal, but I prefer to use the `@/` alias we set up in order to make imports more portable and readable.

packages/ui2/src/components/input.tsx
```diff
@@ -0,0 +1,23 @@
...
+        "placeholder:text-muted-foreground flex h-10 w-full rounded-md border px-3 py-2 text-xs font-semibold focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-offset-2 disabled:cursor-not-allowed disabled:opacity-50",
+        className,
+      )}
+      ref={ref}
+      {...props}
+    />
+  );
+};
+
+Input.displayName = "Input";
```
asutula: Maybe these type of property settings aren't needed since we aren't using `forwardRef` now? I'm not sure.

asutula (approved):
Left a couple little comments with some optional things, but this is good to go.

packages/ui2/src/components/input.tsx
ramirotw: yeah good point, let me remove it

PR #256, opened 2025-04-29
fix: deduplicate bearer token security header

joewagner:
In attempting to generated SDKs with [fern](https://buildwithfern.com/learn/sdks/guides/generate-your-first-sdk) I discovered that we have what amounts to a duplicate entry in our definition of the headers for some of our paths in the openapi spec.

I'm not sure that we want to make the changes here, but I'm opening this for easier discussion.

### DIFF
--- a/apps/api/openapi/openapi.json
+++ b/apps/api/openapi/openapi.json
@@ -17,10 +17,6 @@
     {
       "url": "http://localhost:3000",
       "description": "Local development server"
-    },
-    {
-      "url": "https://api.example.com",
-      "description": "Production server"
     }
   ],
   "components": {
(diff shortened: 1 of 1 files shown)


apps/api/openapi/openapi.json
```diff
@@ -17,10 +17,6 @@
     {
       "url": "http://localhost:3000",
       "description": "Local development server"
-    },
-    {
-      "url": "https://api.example.com",
-      "description": "Production server"
     }
   ],
   "components": {
```
joewagner, line 22: See below in `securitySchemas.BearerAuth`.  This entry acts as the specification for the header that contains the bearer token.  When we also add header in the `parameters` of some routes, it is treated as a duplicate during Fern's SDK generation, which results in an error building the typescript SDK.

apps/api/openapi/openapi.json
```diff
@@ -17,10 +17,6 @@
     {
       "url": "http://localhost:3000",
       "description": "Local development server"
-    },
-    {
-      "url": "https://api.example.com",
```
joewagner, line 22: Looks like I didn't have to remove this, but I took it out for simplicity while investigating.

dtbuchholz: Note: if we end up merging this PR, we should add this public URL back but change it to the competitions API URL: api.competitions.recall.network

apps/api/openapi/openapi.json
mzkrasner: these docs are auto-generated by swagger - is this something we instead need to change in ./src/config/swagger?

joewagner:
closing in favor of #265

PR #257, opened 2025-04-29
Make api an esmodule, update imports

asutula:
In order to depend on other monorepo packages which are esmodules, the `api` package itself needs to be an esmodule. This is because at dev time (`pnpm dev`) all of our packages simply run `tsc` and output only esmodule js and types (vs using `tsup` and outputting a production build that contains cjs as well). This also makes the dev experience better in the `api` package since esmodules are nice and modern.

After converting the module type, I had to update all the import statements. To make relative imports nicer, I created a few remapped source paths in `tsconfig.json`.


apps/api/e2e/debug-test.ts
```diff
@@ -12,7 +12,7 @@ import { config } from "dotenv";
 import fs from "fs";
 import path from "path";
 
-import { killExistingServers } from "./utils/server";
+import { killExistingServers } from "./utils/server.js";
```
asutula: A pattern I tend to use is to use relative imports if the imported file is at a sibling or child location and use the remapped `@/...` path if it is elsewhere.

asutula: So this is at a child location.

apps/api/e2e/setup.ts
```diff
@@ -9,9 +9,14 @@ import fs from "fs";
 import { Server } from "http";
 import path from "path";
 
-import { SchedulerService } from "../src/services/scheduler.service";
-import { dbManager } from "./utils/db-manager";
-import { killExistingServers, startServer, stopServer } from "./utils/server";
+import { SchedulerService } from "@/services/scheduler.service.js";
```
asutula, line 12: This is elsewhere (see the deleted `../`).

apps/api/tsconfig.json
```diff
@@ -5,6 +5,12 @@
     "declarationMap": false,
     "outDir": "./dist",
     "rootDir": ".",
+    "baseUrl": ".",
+    "paths": {
+      "@/*": ["./src/*"],
+      "@/scripts/*": ["./scripts/*"],
+      "@/e2e/*": ["./e2e/*"]
+    },
```
asutula, line 13: Made `src` the primary/short/easy `@/...` path with `scripts` and `e2e` with their own paths.

apps/api/e2e/teardown.ts
asutula, line 1: I'm not currently running this file because there is no Vitest way to run a file after all tests finish. You can only specify the setup file which can export a teardown function or return a teardown function. The `setup.ts` file does this, so that is what is being run. Everything seems to work, not sure if this file is needed or if we need to move some of its logic to the teardown function returned in `setup.ts`.

apps/api/e2e/tsconfig.json
asutula, line 1: The `e2e` folder is just part of the api project and uses the project tsconfig.

apps/api/package.json
```diff
@@ -1,40 +1,41 @@
...
   "private": true,
+  "description": "Solana Trading Simulator Server for hosting trading competitions",
   "main": "dist/src/index.js",
   "scripts": {
     "lint": "eslint . --max-warnings 0",
     "lint:fix": "eslint . --fix",
     "start": "node dist/src/index.js",
     "clean": "rm -rf .turbo node_modules dist",
-    "dev": "ts-node-dev --transpile-only src/index.ts",
+    "dev": "tsx watch src/index.ts",
```
asutula, line 13: Moved to `tsx` because `ts-node-dev` was throwing weird errors now that we're in esmodule territory. Plus, `ts-node-dev` is an ancient project that hasn't been updated in like 5 or more years.

apps/api/e2e/utils/log-reporter.ts
asutula, line 1: Deleted a lot of the functionality of our custom Jest reporter. Have some bare minimum functionality here, but we may want to look at adding in more callbacks when we want to take the time to do that.

apps/api/vitest.config.ts
```diff
@@ -0,0 +1,43 @@
+import tsconfigPaths from "vite-tsconfig-paths";
+import { defineConfig } from "vitest/config";
+
+import LogReporter from "./e2e/utils/log-reporter.js";
+
+export default defineConfig({
+  plugins: [tsconfigPaths()],
```
asutula, line 5: This plugin allows Vitest to work with TS path aliases like we use in the project.

apps/api/vitest.config.ts
```diff
@@ -0,0 +1,43 @@
...
+          maxConcurrency: 1,
+          pool: "threads",
+          poolOptions: {
+            threads: {
+              singleThread: true,
+            },
+          },
+        },
+      },
+    ],
```
asutula, line 39: We're using the Vitest "workspace" feature to split up unit tests from the end to end tests. This allows us to run each workspace "project" individually and with different settings.

asutula:
All the code in `e2e` is now esmodules. Had to update from Jest to Vitest to get esmodule tests working. Vitest is wonderful.

apps/api/package.json
```diff
@@ -1,40 +1,41 @@
...
-    "test:e2e:debug": "ts-node e2e/debug-test.ts",
-    "test:multi-chain": "jest src/services/providers/__tests__/multi-chain.provider.test.ts --forceExit",
-    "test:multi-chain:e2e": "jest -c e2e/jest.config.js e2e/tests/price/multi-chain-provider.test.ts --forceExit",
-    "db:init": "ts-node scripts/setup-db.ts",
-    "db:check": "ts-node scripts/check-db.ts",
-    "db:clean": "NODE_ENV=development ts-node scripts/clean-db.ts",
+    "test": "vitest run --project unit",
+    "test:watch": "vitest watch --project unit",
+    "test:e2e": "vitest run --project e2e",
+    "test:e2e:watch": "vitest watch --project e2e",
```
asutula, line 18: Check out Vitetest and running workspace projects. Pretty slick.

apps/api/e2e/utils/log-reporter.ts
mzkrasner: The PR i created just removes this altogether

mzkrasner (approved):
Great work! I'll fix the custom logger at a later time

PR #265, opened 2025-05-01
fix(api): de-duplicate routes openapi auth header

joewagner:


mzkrasner (approved):
This is great! Tried it locally with the UI both as an admin and participant. Thanks for fixing!

PR #267, opened 2025-05-01
fix: implement reasonable cross-chain trading fees

mzkrasner:
# Summary

- Adds a reasonable and conservative estimation for what cross-chain trading fees to apply in the event of the cross-chain trade
- Streamlines getQuote and executeTrade usage of slippage fee calculation in service layer
- Updates openapi notation with new crossChainFee response 
- Updates e2e test suite types to extend response payload and adds new test cases to affirm expected functionality


asutula (commented):
I have no context here, so will leave it to @carsonfarmer

PR #271, opened 2025-05-02
fix: updates log-reporter to output results in log file

mzkrasner:
# Summary

- Reverts json log reporting (sorry)
- Updates log-reporter to write results to log file

**Note**: Does not output individual test cases between each test file like previous version, but I think that's fine for now

### DIFF
--- a/apps/api/vitest.config.ts
+++ b/apps/api/vitest.config.ts
@@ -4,15 +4,7 @@ import { defineConfig } from "vitest/config";
 export default defineConfig({
   plugins: [tsconfigPaths()],
   test: {
-    reporters: [
-      "verbose",
-      [
-        "json",
-        {
-          outputFile: "./e2e/e2e-server.json",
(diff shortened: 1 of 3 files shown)


apps/api/vitest.config.ts
```diff
@@ -4,15 +4,7 @@ import { defineConfig } from "vitest/config";
...
-    reporters: [
-      "verbose",
-      [
-        "json",
-        {
-          outputFile: "./e2e/e2e-server.json",
-        },
-      ],
-    ],
+    reporters: ["verbose", "e2e/utils/log-reporter.ts"],
```
asutula, line 7: Cool if this works, but I thought you need to pass in an instance of your custom reporter.

asutula: Ah i see the docs say it can be a path.

asutula (approved):

PR #272, opened 2025-05-02
feat: add Speakeasy to generate sdk, docs, and mcp

joewagner:
This PR does a few things:
 - Adds the generated Speakeasy SDK, SDK docs, and an MCP server.  
 - Migrates one test file and the associated helpers to using the SDK instead of making http requests with Axios.
 - Some fixes to CI in order to get tests passing.
 
 See inline comments for details


joewagner:
@asutula something weird here with the monrepo deps again i think.  I'll keep working on it, but it you see anything obvious let me know.

.github/workflows/code-quality.yml
```diff
@@ -58,6 +58,9 @@ jobs:
       - name: Install Dependencies
         run: pnpm install
 
+      - name: Install eslint cli
+        run: pnpm install -w eslint
+
```
joewagner, line 63: I am not sure why, but linting was failing with `eslint command not found`.  This is the easiest way I could find to get CI passing.

.npmrc
```diff
@@ -0,0 +1 @@
+shamefully-hoist=false
```
joewagner, line 1: There were a few different issues with incompatible dependencies, and Speakeasy suggests that if it is used within a monorepo this should be set. 
ref: https://www.speakeasy.com/guides/sdks/typescript-monorepo-tips

apps/api/e2e/tests/api-key-retrieval.test.ts
```diff
@@ -42,7 +42,7 @@ describe("API Key Retrieval", () => {
     await adminClient.loginAsAdmin(adminApiKey);
 
     // Register a new team
-    const { team, apiKey } = await registerTeamAndGetClient(adminClient);
+    const { team, apiKey } = await registerTeamAndGetClient(adminApiKey);
```
joewagner, line 45: I updated this helper to use the SDK.  Because of this you will see changes to all of the test files, but all of these changes are just updating the argument

apps/api/e2e/tests/price-fetching.test.ts
```diff
@@ -32,10 +31,9 @@ const BASE_TOKENS = {
 };
 
 describe("Price Fetching", () => {
-  // Create variables for authenticated clients
-  let adminClient: ApiClient;
-  let client: ApiClient;
+  // Create variables for api keys
   let adminApiKey: string;
+  let clientApiKey: string;
```
joewagner, line 36: This is the test file I updated to use the SDK.  Maybe spend a bit more time looking here to see how the SDK works.

apps/api/e2e/utils/test-helpers.ts
```diff
@@ -37,29 +39,43 @@ export function createTestClient(baseUrl?: string): ApiClient {
...
+  if (!result.success || !result.team || typeof result.team.id !== "string") {
     throw new Error("Failed to register team");
   }
 
   // Create a client with the team's API key
   const client = new ApiClient(result.team.apiKey);
-
-  return { client, team: result.team, apiKey: result.team.apiKey };
+  // TODO: work on replacing `ApiClient` instances with SDK instances everywhere
+  // TODO: the return for this function can be cleaned up when we use the sdk everywhere
```
joewagner, line 70: I wanted to use the sdk in this helper, but to keep this PR from getting even bigger I kept the return type the same.

apps/api/e2e/utils/test-helpers.ts
```diff
@@ -176,3 +192,26 @@ export function generateRandomString(length: number): string {
   }
   return result;
 }
+
+/**
+ * Generate random hex string of specific length
+ */
+export function generateRandomEthAddress(): string {
```
joewagner, line 218: Had to move this here since the `ApiClient` class is no longer being passed into the above helper function.
In the PR to fully migrate to the SDK, we can go back to using a client class instance if desired.

apps/api/openapi/openapi.json
```diff
@@ -14,13 +14,17 @@
     }
   },
   "servers": [
+    {
```
joewagner, line 17: I had to make some fixes to the spec. Since the spec now defines the SDK, it **must** match the actual api behavior.

apps/api/src/config/index.ts
```diff
@@ -139,6 +139,8 @@ const parseEvmChains = (): SpecificChain[] => {
 export const config = {
   server: {
     port: parseInt(process.env.PORT || "3000", 10),
+    // TODO: these ports are going to be put into the openapi json spec, so they can't really be set at runtime, wtd?
+    testPort: 3001,
```
joewagner, line 143: Since the spec defines the port, we can't set it based on an env at runtime.  There's a few options here, so open to ideas and discussion.  I'm thinking we find a resolution to this in the PR to update all the tests to use the SDK.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "license": "MIT AND Apache-2.0",
   "scripts": {
-    "build": "turbo build",
+    "build": "turbo build --filter='!@recallnet/api-sdk'",
```
joewagner, line 6: We do **not** want to build the sdk during `pnpm build`.  The SDK should be built using the bash script you'll see below.  Whoever does the build, will need a Speakeasy account.  
In a future PR, we will want to setup a github action to build and publish the SDK to npm.

packages/api-sdk/.devcontainer/README.md
```diff
@@ -0,0 +1,29 @@
+> **Remember to shutdown a GitHub Codespace when it is not in use!**
+
+# Dev Containers Quick Start
```
joewagner, line 3: This is the beginning of the generated SDK.  Maybe don't worry to much about a proper review here, but if curious take a look to see what the generated code looks like.

packages/eslint-config/package.json
```diff
@@ -15,6 +15,7 @@
     "react-internal.js"
   ],
   "devDependencies": {
+    "@eslint/js": "^9.19.0",
```
joewagner, line 18: This was needed to get CI passing.

.github/workflows/code-quality.yml
dtbuchholz: created a follow up to make sure we properly fix the eslint problem: https://github.com/recallnet/js-recall/issues/318

package.json
dtbuchholz: created the follow up issue: https://github.com/recallnet/js-recall/issues/319

now that the `api` tests use the speakeasy SDK—i think that means anyone trying to make OpenAPI changes will be unable to update tests, correct?

for example, if i add a new route / alter OpenAPI params in the `api` directory, but i dont have a speakeasy account, i cannot generate the new SDK from spec changes. thus, i wont be able to update the tests with new params, so they'll fail.

packages/api-sdk/package-lock.json
dtbuchholz, line 1: is there a way where the build process works with our monorepo / `pnpm`? not a huge issue, but ideally, subdirs shouldnt have a lockfile

packages/api-sdk/package.json
```diff
@@ -0,0 +1,140 @@
...
+      "./models/operations": "./src/models/operations/index.ts",
+      "./*.js": "./src/*.ts",
+      "./*": "./src/*.ts"
+    }
+  },
+  "sideEffects": false,
+  "scripts": {
+    "lint": "eslint --cache --max-warnings=0 src",
+    "build:mcp": "bun src/mcp-server/build.mts",
+    "build": "npm run build:mcp && tshy",
```
dtbuchholz, line 28: same question here re: `npm` usage...is that a speakeasy thing, or should we be using `pnpm` here?

packages/api-sdk/package.json
```diff
@@ -0,0 +1,140 @@
+{
+  "name": "@recallnet/api-sdk",
+  "version": "0.0.4",
```
dtbuchholz: version should be `0.0.0`, right?

also, if we _dont_ want this package to be published upon merging to main, we'll need to add the field `"private": true` ([example](https://github.com/recallnet/js-recall/blob/main/packages/eslint-config/package.json#L5)). and since it sounds like this package might not even use the changesets flow, we also might want to ignore it [here](https://github.com/recallnet/js-recall/blob/20a1ee300643c3c7676805d2d783835fcb37818c/.changeset/config.json#L13).

it's a weird thing with the changesets publishing action. the first time you merge a new package to main, the changesets action will auto-publish it to npm (unless you mark it as private). but, once a package is exists on npm, changesets will only publish a new release if you explicitly "stage" it with [`pnpm changeset`](https://github.com/recallnet/js-recall/blob/20a1ee300643c3c7676805d2d783835fcb37818c/package.json#L12).

package.json
dtbuchholz: btw, idk if this `build` change is needed? since the `api-sdk`'s script is `"build": "npm run build:mcp && tshy"`, it technically doesnt rely on the speakeasy CLI, right? and the `api` package needs the `dist` output anyways so that it can run tests. (plus, i'm able to run it w/o issue, so idk if the `--filter` flag is working.)

but to your point—maybe we need a `generate` script in the `api-sdk`'s `package.json` file. someone with speakeasy credentials can run it manually, and it will execute the script and update the SDK locally (instead of purely relying on a CI/CD speakeasy action).

dtbuchholz:
@joewagner left a couple of comments. i think the only major concern is about internal DX with the `api` tests depending on the speakeasy SDK, but lmk if my understanding is incorrect.

joewagner:
> @joewagner left a couple of comments. i think the only major concern is about internal DX with the `api` tests depending on the speakeasy SDK, but lmk if my understanding is incorrect.

So, this is a really good point.  I tried a few different options yesterday, I also messaged the Speakeasy folks. So far we are only generating one SDK, so the account is free.  We can potentially all share the login for this account.  This doesn't seem like the best option, but once you login you seem to stay logged in indefinitely.
I'll see what the Speakeasy folks say regarding sharing a login.

packages/api-sdk/package.json
joewagner: This is generated by Speakeasy

packages/api-sdk/package-lock.json
joewagner: I just pushed up a rebase which included a bash script [`build-api-sdk.sh`](https://github.com/recallnet/js-recall/pull/272/files#diff-caf8c0a4d710229182088955d7351bf861a45b7fa2b76e2410ab088a9aca0029).   
I does three things: 
 - build the openapi spec and docs
 - build the speakeasy sdk based on the docs
 - run the our format fix script (because the generated openapi and sdk don't pass linting)

RE: the lockfile.  I agree, good call.  The build of the dist happens via the `speakeasy` cli.  
As a workaround, how about if we  use `.gitignore` to make sure that isn't committed, and add to the build script so that the lockfile is `rm`ed?

package.json
joewagner: I sort of discuss this below, but mentioning here for continuity.
I had made a script to run `api-sdk` build process, but I put it in the speakeasy files, so it was deleted during a re-gen.  I made another commit earlier today that re adds the script to the repo root.

RE: devex
If devs are willing to download the speakeasy cli, anyone can generate the sdk as many times as they want.  Speakeasy does not have the concept of  "developer seats"

packages/api-sdk/package-lock.json
dtbuchholz: yeah, ignoring the file makes sense to me

package.json
dtbuchholz: got it. i just created an account and was able to run things w/o issue, so it seems like it should be fine.

packages/api-sdk/package.json
dtbuchholz: @joewagner just reiterating: we need to add `@recallnet/api-sdk` into this [changesets array](https://github.com/recallnet/js-recall/blob/919d49d85e550188fc119561f4366f15fd02ffdd/.changeset/config.json#L13). else, the package will get published when we merge this PR. it looks like speakeasy lets us set custom `package.json` fields with [`gen.yaml`'s `additionalPackageJSON`](https://www.speakeasy.com/docs/speakeasy-reference/generation/ts-config#additional-json-package) field, but i'm not sure if setting it to private impacts speakeasy or not:
```yaml
typescript:
...
  additionalPackageJSON:
    private: true
```

aside from that, i think we should just get this PR merged so you dont have to keep dealing with OpenAPI changes. i have some nonce/auth work that i'm working on so can test out the internal DX wrt spec changes + speakeasy SDK generation.

joewagner: I missed the changesets comment.  Sorry about that.  I"ll get that update made and push a fix to the open api spec.

joewagner: Ok, I just pushed these changes and corrected the version.

joewagner:
I had to make some fixes to the test helper to get the imageUrl working with the sdk, but ci looks good now.

dtbuchholz (approved):
nice, lgtm

PR #274, opened 2025-05-02
fix: ensure specificChain is ALWAYS recorded in balances

mzkrasner:
# Summary

- Trade simulator service relies on the price provider to give prices before executing a trade which already always records the specificChain
- Trade simulator service then utilizes the balance manager service (and subsequently the balance repository) to update the balances to reflect that trade
- This update brings over the specific chain value already available from the trade execution into the argument set used to engage the balance manager, and also updates the database helper methods used by the balance repository to ensure specificChain is always recorded in the database when trades are made


joewagner (approved):
This looks good.
Is there a plan for updating the schema of the `balances` table?

mzkrasner:
> This looks good. Is there a plan for updating the schema of the `balances` table?

Ah sorry - saw this after I had merged it.

Do you mean making those fields required in the db?

joewagner:
Or put another way, I saw that the `Balances` interface was modified, which made me think we might want to migrate the table schema in the database.
Potentially no an issue once Aaron gets Drizzle integrated, so if doing nothing is easy right now that sounds good.

PR #279, opened 2025-05-05
fix(api): cleanup openapi spec

joewagner:
#265 missed a few place were the openapi spec needs to be updated for code generators to consume the spec.
This PR fixes those, and enables sdk code generation tools to work with our spec.


mzkrasner:
This looks good - made a PR into this branch to resolve some of the script issues I realized were broken given the update to ESM. Also left a note that we will still need to fix the compiled javascript issue

mzkrasner (approved):
LGTM!

PR #280, opened 2025-05-05
fix: update price caching and remove problematic fallback method

mzkrasner:
# Summary

## Implemented Chain-Aware Caching in Providers:

- Added a composite key structure (tokenAddress:specificChain) to both DexScreenerProvider and MultiChainProvider
- Ensures prices are cached and retrieved correctly for each token-chain combination
- Prevents cross-chain price confusion when the same token exists on multiple chains

## Improved Chain Detection Logic:

- Updated getPriceForSpecificEVMChain method to strictly honor the requested chain
- Enhanced the logic to return null when a token doesn't exist on a specified chain
- Added proper chain detection to reliably identify which chain a token belongs to

## Eliminated Cache Redundancy:

- Removed the redundant cache layer in PriceTracker service
- Now relies exclusively on the provider-level caching system
- Simplified the codebase and eliminated potential cache inconsistencies

## Enhanced Test Case Reliability:

- Fixed the test case to properly verify that trades are rejected when a token doesn't exist on a specified chain
- Added additional chain validation to correctly detect when a token doesn't belong to a specific chain

Also removes `getLatestPrice` fallback method


packages/comps-db/src/schema.ts
```diff
@@ -116,7 +116,7 @@ export const balances = pgTable(
     updatedAt: timestamp("updated_at", {
       withTimezone: true,
     }).defaultNow(),
```
asutula, line 118: You'll have to generate a migration to get this change to show up in the db. If we apply that migration to the existing db, you might have to edit the migration SQL to set some value for rows that have a `null` specific chain. Does that sound reasonable? We should talk about this more, but if we don't want to ever migrate the existing database because we just don't need to or because it's too complicated, then we can just ignore migrations for now and make whatever changes to the schema we want.

asutula (approved):

apps/api/openapi/openapi.json
```diff
@@ -787,6 +690,114 @@
           "500": {
             "description": "Server error"
           }
+        },
+        "parameters": [
```
joewagner, line 694: Not sure how this is working it's way into the spec, but we shouldn't be adding the bearer token header like this.  Instead we should be using the BearerAuth security directive.

e.g.
```
"security": [
  {
     "BearerAuth": []
  }
],
```
 Here are docs describing this idea https://swagger.io/docs/specification/v3_0/authentication/bearer-authentication/

joewagner: It looks like there are two duplicate routes being defined in `account.routes.ts`.  Must be a rebase or merge conflict resolution error.  

Checkout line 393 and 553
https://github.com/recallnet/js-recall/blob/main/apps/api/src/routes/account.routes.ts#L393
https://github.com/recallnet/js-recall/blob/main/apps/api/src/routes/account.routes.ts#L553

PR #282, opened 2025-05-05
Mzk/update scripts

mzkrasner:
# Summary

- Noticed when testing the updates from your branch that the scripts needed to be updated to work with esmodule
- This PR brings over the latest from main and also updates the scripts to work with esm

**NOTE**: anything using npm run start is not working as path aliases are not properly being resolved in the compiled javascript - this we will still need to fix


joewagner (approved):
Awesome!

Issue #268, opened 2025-05-01
Remove deactivated agents from the ranking


KGmajor:
When an agent is deactivated from a competition, it should not be ranked in the leaderboard endpoint. 

For example, if they're deactivated and ranked 15, they still hold that leaderboard spot even though they're not active. They should be moved off the ranking and into a separate array.

mzkrasner:
Ready for review: https://github.com/recallnet/js-recall/pull/285

PR #285, opened 2025-05-06
fix: add tsc-alias and change leaderboard return type

mzkrasner:
# TSC-Alias build-related changes

- Compiled JavaScript was previously not reflecting the resolved alias paths used in TS code
- Resolved by adding tsc-alias to dependencies and use in build script

## Leaderboard changes

Aligns with requested change outlined in issue: https://github.com/recallnet/js-recall/issues/268

### Modified the CompetitionController.getLeaderboard method to:

- Separate active and inactive teams into different arrays
- Only assign ranks to active teams
- Put inactive teams in a separate inactiveTeams array without ranks

### Updated the API types:

- Created a new InactiveTeamEntry interface without the rank property
- Updated the LeaderboardResponse interface to include the new inactiveTeams array
- Made hasInactiveTeams a required property
- Updated the OpenAPI documentation to reflect the new response structure


mzkrasner:
Tagging @joewagner as this change alters one of the openapi notation files (want to make sure it doesn't break anything for you)

joewagner (approved):
This looks good to me.  Only callout/question is that we are making a breaking change to the schema of a response.  Are there planned changes to clients that we need to make before we deploy this?

mzkrasner:
> This looks good to me. Only callout/question is that we are making a breaking change to the schema of a response. Are there planned changes to clients that we need to make before we deploy this?

Yes - so the https://github.com/recallnet/trading-simulator-mcp is massively behind even the current main branch of js-recall (it's frozen at the version being used by the active competition).

Once @asutula has the drizzle PR merged, this will be a good time to bring over this service into js-recall so it will automatically be kept up-to-date with the latest changes from the API.

We should also assume that there will probably always be a difference between the versions being used by an active competition and the updates concurrently being made to js-recall, so I anticipate the situation we're currently in (code-frozen versions being used for competitions while improvements continue to be made to the competitions API) being a recurring trend that we will have to deal with

joewagner:
> > This looks good to me. Only callout/question is that we are making a breaking change to the schema of a response. Are there planned changes to clients that we need to make before we deploy this?
> 
> Yes - so the https://github.com/recallnet/trading-simulator-mcp is massively behind even the current main branch of js-recall (it's frozen at the version being used by the active competition).
> 
> Once @asutula has the drizzle PR merged, this will be a good time to bring over this service into js-recall so it will automatically be kept up-to-date with the latest changes from the API.
> 
> We should also assume that there will probably always be a difference between the versions being used by an active competition and the updates concurrently being made to js-recall, so I anticipate the situation we're currently in (code-frozen versions being used for competitions while improvements continue to be made to the competitions API) being a recurring trend that we will have to deal with

Awesome!

PR #289, opened 2025-05-06
Integrate Drizzle

asutula:
Big update.

- Updates the `comps-db` package to reflect the latest db schema, defined with Drizzle.
- Integrates that Drizzle schema and Drizzle queries into the `api` app.
- Integrates Drizzle tooling for migrations, data seeding, studio, etc into the `api` app.
- Makes some simplifications and improvements around routers, controllers, and repositories.
- Cleans up how dependencies on services are set up, allowing the app to apply any needed db migrations before services are instantiated.


.github/workflows/api-ci.yml
```diff
@@ -7,50 +7,13 @@ on:
...
-          path: ${{ env.STORE_PATH }}
-          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
-          restore-keys: |
-            ${{ runner.os }}-pnpm-store-
-
-      - name: Install Dependencies
-        run: pnpm install
-
-      - name: Build project
-        run: pnpm run build
```
asutula, line 47: We're already doing a build test in the other action file.

.github/workflows/api-ci.yml
```diff
@@ -7,50 +7,13 @@ on:
...
+env:
+  NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID: fe319abf1a1b0d76b39894e5e17aff0c
+  NEXT_PUBLIC_CHAIN_NAME: testnet
 
+jobs:
   unit-tests:
     runs-on: ubuntu-latest
-    defaults:
-      run:
-        working-directory: apps/api
```
asutula, line 53: Since the api package now depends on other packages in the repo, we need to build the entire repo. I now set the working directory to the api package only in the steps that need to run there (running the tests).

.github/workflows/api-ci.yml
```diff
@@ -81,7 +44,11 @@ jobs:
       - name: Install Dependencies
         run: pnpm install
 
+      - name: Build everything
+        run: pnpm build
+
```
asutula, line 49: Build the whole repo.

.github/workflows/api-ci.yml
```diff
@@ -81,7 +44,11 @@ jobs:
...
       - name: Run unit tests
+        working-directory: apps/api
```
asutula, line 51: Set the working directory for running tests.

apps/api/.env.example
```diff
@@ -2,13 +2,8 @@
 PORT=3000
 NODE_ENV=development
 
-# Database Configuration
-DB_HOST=localhost
-DB_PORT=5432
-DB_USERNAME=postgres
-DB_PASSWORD=postgres
+DATABASE_URL=postgresql://postgres:postgres@localhost:5432/trading_simulator
```
asutula, line 5: I couldn't resist simplifying certain things I came across. Supporting both URL connection string and individual parameters was hurting my head since it allows us to use one style in some places and another style in other places. Too many things to keep straight and debug when things aren't working. Just simpler and easier to support url string, especially since it is functionally equivalent to the individual parameters.

apps/api/e2e/tests/chain-specific.test.ts
```diff
@@ -165,24 +170,19 @@ describe("Specific Chains", () => {
...
-    const pool = getPool();
-
     // Query the trades table to check if specificChain fields were correctly populated
-    const tradesResult = await pool.query(
-      "SELECT from_token, to_token, from_specific_chain, to_specific_chain FROM trades WHERE team_id = $1",
-      [team.id],
-    );
+    const trade = await db.query.trades.findFirst({
+      where: eq(trades.teamId, team.id),
+    });
```
asutula, line 176: If there were places where raw sql was run that could simply be replaced by querying with drizzle, I did that.

apps/api/e2e/tests/chain-specific.test.ts
```diff
@@ -216,31 +216,22 @@ describe("Specific Chains", () => {
...
-      "SELECT token_address, specific_chain FROM portfolio_token_values WHERE portfolio_snapshot_id = $1",
-      [snapshotId],
-    );
+    const snapshotResult = await db.query.portfolioSnapshots.findFirst({
+      where: eq(portfolioSnapshots.teamId, team.id),
+      orderBy: desc(portfolioSnapshots.timestamp),
+      with: {
+        portfolioTokenValues: true,
+      },
+    });
```
asutula, line 226: Check out this Drizzle feature that includes related data by using `with`. No need to run a second query.

apps/api/e2e/utils/api-types.ts
```diff
@@ -80,7 +80,7 @@ export interface TeamProfileResponse extends ApiResponse {
 
 // Token balance type
 export interface TokenBalance {
-  token: string;
+  tokenAddress: string;
```
asutula, line 83: This is the only change as far as I know to the api types. I chose not to integrate the drizzle types into the api client yet, because I think it can be another PR and we might have a cleaner solution based on the open api spec very soon.

apps/api/package.json
```diff
@@ -19,11 +19,14 @@
     "test:e2e:runner": "tsx e2e/run-tests.ts",
     "test:multi-chain": "vitest run src/services/providers/__tests__/multi-chain.provider.test.ts",
     "test:multi-chain:e2e": "vitest run e2e/tests/price/multi-chain-provider.test.ts",
-    "db:init": "tsx scripts/setup-db.ts",
-    "db:check": "tsx scripts/check-db.ts",
-    "db:clean": "NODE_ENV=development tsx scripts/clean-db.ts",
-    "db:reset": "npm run db:clean && npm run db:init",
-    "db:drop-all": "NODE_ENV=development tsx scripts/drop-all-tables.ts",
+    "db:gen-migrations": "drizzle-kit generate",
```
asutula, line 22: Compares the current Drizzle schema to the current state of the database and creates any migration SQL files that are needed.

apps/api/package.json
```diff
@@ -19,11 +19,14 @@
...
+    "db:check": "drizzle-kit check",
```
asutula, line 23: Checks if there are any migrations that need to be applied to the db.

apps/api/package.json
```diff
@@ -19,11 +19,14 @@
...
+    "db:migrate": "drizzle-kit migrate",
```
asutula, line 24: Runs any needed migrations

apps/api/package.json
```diff
@@ -19,11 +19,14 @@
...
+    "db:clean": "tsx scripts/clean-db.ts",
```
asutula, line 25: Deletes all data from tabled defined in the drizzle schema and resets sequences, etc.

apps/api/package.json
```diff
@@ -19,11 +19,14 @@
     "test:e2e:runner": "tsx e2e/run-tests.ts",
     "test:multi-chain": "vitest run src/services/providers/__tests__/multi-chain.provider.test.ts",
     "test:multi-chain:e2e": "vitest run e2e/tests/price/multi-chain-provider.test.ts",
-    "db:init": "tsx scripts/setup-db.ts",
-    "db:check": "tsx scripts/check-db.ts",
-    "db:clean": "NODE_ENV=development tsx scripts/clean-db.ts",
-    "db:reset": "npm run db:clean && npm run db:init",
```
asutula, line 25: I don't really understand the point of this one.  Seems like cleaning the db (deleting data) be all you need. Cleaning and then running `init` doesn't do anything additional because all the tables already exist.

apps/api/package.json
```diff
@@ -19,11 +19,14 @@
...
-    "db:drop-all": "NODE_ENV=development tsx scripts/drop-all-tables.ts",
+    "db:gen-migrations": "drizzle-kit generate",
+    "db:check": "drizzle-kit check",
+    "db:migrate": "drizzle-kit migrate",
+    "db:clean": "tsx scripts/clean-db.ts",
+    "db:reset": "pnpm db:clean && pnpm db:migrate",
```
asutula, line 26: See comment to the left. I think maybe we don't need this script at all.

apps/api/package.json
```diff
@@ -19,11 +19,14 @@
...
+    "db:drop-all": "NODE_ENV=development tsx scripts/drop-all.ts",
```
asutula, line 27: Drop everything.

apps/api/package.json
```diff
@@ -19,11 +19,14 @@
...
+    "db:seed": "tsx scripts/seed-db.ts",
+    "db:studio": "drizzle-kit studio",
```
asutula, line 29: Exciting new stuff here. Drizzle allows you to seed the database with fake data. This process can be very customized, but I'm just doing the simplest thing for now.

Studio is a web interface to browse the db and run queries.

apps/api/scripts/initialize-db.ts
asutula, line 1: This is now achieved by migrating the db using Drizzle.

apps/api/src/controllers/account.controller.ts
```diff
@@ -1,354 +1,364 @@
...
-          updatedAt: team.updatedAt,
-        },
-      });
-    } catch (error) {
-      next(error);
-    }
-  }
+import { ServiceRegistry } from "@/services/index.js";
 
+export function makeAccountController(services: ServiceRegistry) {
```
asutula, line 12: Changed all the controllers to use this pattern of a "make..." function that returns an object with all the route handlers on it. This allows us to easily pass in the dependency on the service registry when we wire up the app in `src/index.ts`.

asutula: All the logic within the route handler functions should be unchanged.

apps/api/src/database/base-repository.ts
asutula, line 1: Repositories are simple stateless groups of functions that just interact with the database. Drizzle makes that very simple and easy, so the complexity of using classes and inheritance is just overkill now.

apps/api/src/database/connection.ts
asutula, line 1: Drizzle also simplifies how we interact with the database. We don't need to think about the connection other than when we originally create it and pass it into Drizzle. Everywhere else, we simply use the Drizzle `db` instance. So the methods and functionality this class provided aren't needed.

apps/api/src/database/db.ts
asutula, line 1: This is where we build and export our Drizzle db instance.

apps/api/src/database/db.ts
```diff
@@ -0,0 +1,103 @@
...
+// Use connection URL which includes all connection parameters
+const pool = new Pool({
+  connectionString: config.database.url,
+  ...sslConfig,
+});
+
+export const db = drizzle({
+  client: pool,
+  schema: { ...schema, ...relations },
+});
```
asutula, line 62: Drizzle actually supports simply passing the connection URL into `drizzle(<url>)` and you're all ready to go. Since we have the code to configure the ssl cert, I left that in there and used the other Drizzle constructor overload that accepts a `Pool` that may be configured with the ssl cert. If we don't actually use the ssl cert, seems like a good thing to remove to reduce complexity.

apps/api/src/database/db.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export async function migrateDb() {
+  await migrate(db, {
+    migrationsFolder: path.join(__dirname, "../../drizzle"),
+  });
+}
+
+export async function seedDb() {
+  await seed(db, schema);
+}
```
asutula, line 103: Here are the db helper functions used by various scripts and test helpers. The `seed` api is very customizable if we ever want to explore that.

apps/api/src/database/repositories/balance-repository.ts
asutula, line 1: As mentioned before, repositories are just groups of stateless functions that interact with the db. No need for classes, registry class, all that complexity. Converted these to simple esmodules that export data access functions. Those functions use the Drizzle `db` to interact with the database.

asutula: These repository functions are where the core "migrate to Drizzle" work has happened.

apps/api/src/database/types.ts
asutula, line 1: The types come from Drizzle now.

apps/api/src/index.ts
```diff
@@ -2,24 +2,68 @@ import cors from "cors";
...
+      "WARNING: Starting server without successful database initialization. " +
+        "Some functionality may be limited until database connection is restored.",
+    );
+  } else {
+    console.error(
+      "Failed to start server due to database initialization error. Exiting...",
+    );
+    process.exit(1);
+  }
+}
```
asutula, line 55: Before anything else can happen, we have to migrate the database. This could mean creating all the tables for the first time or making additional changes to the schema. This is all done from the Drizzle migration files.

apps/api/src/index.ts
```diff
@@ -2,24 +2,68 @@ import cors from "cors";
...
+
+const services = new ServiceRegistry();
```
asutula, line 57: After the db is migrated, we can instantiate all the services.

apps/api/src/index.ts
```diff
@@ -44,15 +88,37 @@ app.use(
...
+const adminMiddleware = adminAuthMiddleware(services.teamManager);
+
+const accountController = makeAccountController(services);
+const adminController = makeAdminController(services);
+const competitionController = makeCompetitionController(services);
+const docsController = makeDocsController();
+const healthController = makeHealthController();
+const priceController = makePriceController(services);
+const publicController = makePublicController(services);
+const tradeController = makeTradeController(services);
```
asutula, line 100: Instantiate all our controllers, passing in dependencies (usually the service registry). We could be more granular about which services are actually dependencies, but this works for now.

apps/api/src/index.ts
```diff
@@ -44,15 +88,37 @@ app.use(
...
+const accountRoutes = configureAccountRoutes(accountController);
+const adminRoutes = configureAdminRoutes(adminController, adminMiddleware);
+const adminSetupRoutes = configureAdminSetupRoutes(adminController);
+const competitionRoutes = configureCompetitionRoutes(competitionController);
+const docsRoutes = configureDocsRoutes(docsController);
+const healthRoutes = configureHealthRoutes(healthController);
+const priceRoutes = configurePriceRoutes(priceController);
+const publicRoutes = configurePublicRoutes(publicController);
+const tradeRoutes = configureTradeRoutes(tradeController);
+
```
asutula, line 111: Initialize all the routes.

apps/api/src/index.ts
```diff
@@ -44,15 +88,37 @@ app.use(
...
+const competitionController = makeCompetitionController(services);
+const docsController = makeDocsController();
+const healthController = makeHealthController();
+const priceController = makePriceController(services);
+const publicController = makePublicController(services);
+const tradeController = makeTradeController(services);
+
+const accountRoutes = configureAccountRoutes(accountController);
+const adminRoutes = configureAdminRoutes(adminController, adminMiddleware);
+const adminSetupRoutes = configureAdminSetupRoutes(adminController);
```
asutula, line 104: Made this new route for handling just the admin setup. It was confusing before how just one route in the the admin routes wasn't protected with auth, and that was achieved by the very specific order of when the middleware was registered with the router vs the routes. This keeps things simple where all the routes in the same file are treated with the same middleware.

apps/api/src/index.ts
```diff
@@ -71,54 +137,14 @@ app.get("/", (_req, res) => {
 // Apply error handler
 app.use(errorHandler);
 
-// Start server
-const startServer = async () => {
```
asutula, line 75: No need to wrap all the code in a function, you can just run it all top level, awaits and everything.

apps/api/src/services/balance-manager.service.ts
asutula, line 1: All the services were updated to import the new style of repository functions. Shouldn't really be any logic changes in any service.

apps/api/src/services/index.ts
asutula, line 1: Once I removed singleton access to the service registry, it revealed a circular dependency between a couple services. This resulted in the creation of the new portfolio snapshotter service.

apps/api/vitest.config.ts
```diff
@@ -4,7 +4,7 @@ import { defineConfig } from "vitest/config";
 export default defineConfig({
   plugins: [tsconfigPaths()],
   test: {
-    reporters: ["verbose", "e2e/utils/log-reporter.ts"],
+    reporters: ["default", "e2e/utils/log-reporter.ts"],
```
asutula, line 7: The verbose output made it hard to understand what was going on.

asutula:
I haven't done a good job testing the `scripts/`, we we should be sure to do that.

apps/api/e2e/tests/chain-specific.test.ts
joewagner: nice, does `with` do a join under the hood?

apps/api/e2e/tests/price/multi-chain-provider.test.ts
```diff
@@ -110,18 +112,14 @@ describe("Multi-Chain Provider Tests", () => {
...
-      SELECT EXISTS (
+    const result = await db.execute(
+      sql`SELECT EXISTS (
         SELECT 1
         FROM information_schema.columns
         WHERE table_name = 'prices'
         AND column_name = 'specific_chain'
-      ) as column_exists;
-    `);
+      ) as column_exists;`,
```
joewagner: Curious why we use `db.execute` here instead of `findFirst`?  Is this just a better performing query?

apps/api/package.json
```diff
@@ -36,7 +39,7 @@
     "comp:status": "tsx scripts/competition-status.ts",
     "generate-openapi": "tsx -e \"const { swaggerSpec } = require('./src/config/swagger'); require('fs').writeFileSync('./openapi/openapi.json', JSON.stringify(swaggerSpec, null, 2));\"",
     "generate-markdown": "openapi-markdown -i ./openapi/openapi.json -o ./openapi/API.md",
-    "generate-docs": "npm run generate-openapi && npm run generate-markdown"
+    "generate-docs": "pnpm generate-openapi && pnpm generate-markdown"
```
joewagner, line 42: I noticed that this script hasn't been run, but the spec was changed.

apps/api/src/routes/account.routes.ts
```diff
@@ -1,454 +1,597 @@
...
+   * @openapi
+   * /api/account/reset-api-key:
+   *   post:
+   *     tags:
+   *       - Account
+   *     summary: Reset team API key
+   *     description: Reset the API key for the authenticated team. This will invalidate the current API key and generate a new one.
+   *     security:
+   *       - BearerAuth: []
+   *     parameters:
```
joewagner: It's tricky to tell with the way this diff is presented, but you've added back an older version of the spec here.   This is probably an artifact of a merge conflict resolution.
We don't want to include the `BearerAuth` security strategy and explicitly add the `Authorization` header.  
I can create a separate PR to update the spec if you'd like?

joewagner (approved):
@asutula Tons of great changes here. I left a few comments inline, but none are blockers for merging imo.

apps/api/e2e/tests/price/multi-chain-provider.test.ts
asutula: We're querying the postgres information catalog here, so they are tables that we don't define in our Drizzle schema. The idea is to see if a table that we do create is actually created. No way to access them using the typical Drizzle api that exists for the table we define ourselves. 

This test is probably not needed now that we have Drizzle. We migrate the database using Drizzle to ensure all our tables are created. If that process fails, the test setup will fail and we won't even get to the point of running tests. I'll delete this test.

apps/api/package.json
asutula: Not sure. I never ran this script and I never updated the spec in any way. Let me see what the spec change was.

apps/api/src/routes/account.routes.ts
asutula: Can I just grab the latest spec from `main`?

apps/api/e2e/tests/price/multi-chain-provider.test.ts
asutula: Deleted the test.

apps/api/src/routes/account.routes.ts
asutula: Ok yes, I messed up the rebase of this file and ended up with duplicate routes defined, each with different spec comments. I removed the duplicate and grabbed the spec comment from main.

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -102,6 +107,7 @@ export async function addTeamToCompetition(
       .values({
         competitionId,
         teamId,
+        createdAt: new Date(),
```
mzkrasner, line 110: if it's an update, createdAt should not be altered i'd think

mzkrasner (commented):
makes sense

apps/api/.env.example
mzkrasner: Yeah - agreed this is easier to maintain

apps/api/src/index.ts
mzkrasner: nice

mzkrasner:
@asutula and I spoke about a few things separately that lead to some of the commits above. Good to go on my end once the e2e tests are back to passing state

mzkrasner (approved):

PR #293, opened 2025-05-08
Asutula/explore onboarding

asutula:


.github/workflows/release.yml
```diff
@@ -9,7 +9,6 @@ on:
 
 env:
   NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID: fe319abf1a1b0d76b39894e5e17aff0c
-  NEXT_PUBLIC_WALLET_PROJECT_ID: 768afa89687ff9f4d0ec5a643e4ace29
```
asutula, line 12: No need to create a new env var for this, just use the existing `NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID`. We can set it to different values per project deployment if we want.

apps/comps/package.json
```diff
@@ -40,8 +40,7 @@
     "react-hook-form": "^7.54.2",
     "viem": "^2.28.0",
     "wagmi": "^2.14.8",
-    "yup": "^1.6.1",
```
asutula, line 43: Yup wasn't used, remove it.

apps/comps/package.json
```diff
@@ -40,8 +40,7 @@
...
-    "zod": "^3.24.4"
+    "zod": "^3.24.2"
```
asutula, line 43: Tracked down all the weird problems we were having to that you had installed a newer version of Zod in this project than what the other projects were using. The error messages we were seeing were strange and not helpful at all, but I noticed the thing they had in common was some relationship, sometimes very indirectly, to Zod. I made sure all packages use the same Zod version, and things seem to work now.

apps/comps/tsconfig.json
asutula, line 1: Just pretter formatting fixes.

apps/portal/app/buckets/[address]/_components/add-object-dialog.tsx
```diff
@@ -80,8 +80,7 @@ export default function AddObjectDialog({
   const router = useRouter();
 
   const form = useForm<z.infer<typeof formSchema>>({
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    resolver: zodResolver(formSchema as any) as any,
+    resolver: zodResolver(formSchema),
```
Comment (asutula) [inline apps/portal/app/buckets/[address]/_components/add-object-dialog.tsx:83]: This compiles fine now with the Zod version the same across packages.

packages/agent-toolkit/src/langchain/tool.ts
```diff
@@ -18,7 +18,7 @@ class RecallTool extends StructuredTool {
   name: string;
   description: string;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  schema: any;
+  schema: z.ZodObject<any, any, any, any>;
```
asutula, line 21: This works now as well.

PR #294, opened 2025-05-08
fix(api): remove duplicate route declaration

joewagner:
Looks like there was a duplicate entry for the `/api/account/trades` route.  This removes one of them dups and generates the api spec docs.


apps/api/src/routes/account.routes.ts
```diff
@@ -458,98 +458,5 @@ export function configureAccountRoutes(
...
-   *                         description: Specific chain for the source token
-   *                       toSpecificChain:
-   *                         type: string
-   *                         description: Specific chain for the destination token
-   *       401:
-   *         description: Unauthorized - Missing or invalid authentication
-   *       500:
-   *         description: Server error
-   */
-  router.get("/trades", controller.getTrades);
```
joewagner, line 552: this is a duplicate of line 392

asutula: Looks like the openapi comment is very different for each duplicate, so just to be sure you're choosing the correct version of the comment...

asutula (approved):

apps/api/src/routes/account.routes.ts
joewagner: Good call, I believe the one I chose is correct.

PR #295, opened 2025-05-08
feat: change x-chain trading toggle to enum to restrict or allow within-evm trading

mzkrasner:
# Summary

- Previous cross-chain trading was a boolean toggle, which treated any svm --> evm or evm --> evm (such as base --> eth mainnet) essentially the same
- This change would allow for administrators to set up competitions that would restrict cross-parent chain trades while still allowing evm --> evm trades as shown in the example above

## Changes
There are now three different trading control modes:

- ALLOW - Full cross-chain trading enabled across any chains
- DISALLOWALL - All cross-chain trading is disabled (default)
- DISALLOWXPARENT - Partial restriction: allows trading within the same parent chain (e.g., between EVM chains) but blocks trading across different parent chains (e.g., EVM to SVM)

## Details

- Added a new migration that adds the new column
- Updated the database schema definition to use the new enum
- Modified service layer implementation to support the new trading modes
- Fixed controller code to handle the enum values
- Updated e2e tests to verify all three modes of operation
- Added a specific test case for DISALLOWXPARENT mode that confirms:

1. Trading within the same parent chain (EVM-to-EVM) works correctly
2. Trading across parent chains (EVM-to-SVM) is properly blocked

- Updated the documentation in README.md to reflect the new configuration options


packages/comps-db/src/schema.ts
```diff
@@ -16,6 +16,16 @@ import {
...
+// Add the enum type for cross chain trading
+export const CrossChainTradingType = {
+  DISALLOWALL: "DISALLOWALL",
+  DISALLOWXPARENT: "DISALLOWXPARENT",
+  ALLOW: "ALLOW",
+} as const;
+
+export type CrossChainTradingType =
+  (typeof CrossChainTradingType)[keyof typeof CrossChainTradingType];
+
```
asutula: Ok so I think the shape of the changes in other files of this PR looks good. But this isn't the way I would implement this change at the db level. 

I'd recommend using Drizzle's `pgEnum` which creates a Postgres enum type under the hood and results in type safe data from the database onward.

asutula:
```suggestion
export const crossChainTradingType = pgEnum("cross_chain_trading_type", [
  "disallowAll",
  "disallowXParent",
  "allow",
]);

export type CrossChainTradingType =
  (typeof crossChainTradingType.enumValues)[number];
```

packages/comps-db/src/schema.ts
```diff
@@ -63,17 +73,22 @@ export const competitions = pgTable(
     startDate: timestamp("start_date", { withTimezone: true }),
     endDate: timestamp("end_date", { withTimezone: true }),
     status: varchar({ length: 20 }).notNull(),
-    allowCrossChainTrading: boolean("allow_cross_chain_trading")
+    crossChainTradingType: varchar("cross_chain_trading_type", { length: 20 })
       .notNull()
-      .default(false),
+      .default(CrossChainTradingType.DISALLOWALL),
```
asutula:
```suggestion
      crossChainTradingType: crossChainTradingType("cross_chain_trading_type")
      .notNull()
      .default("disallowAll"),
```

asutula (commented):
Left a comment about how to better model the enum in Drizzle. If you make that change, the propagate the resulting changes all over, things will be good.

packages/comps-db/src/schema.ts
asutula: This suggested change should also overwrite the previous couple lines.

packages/comps-db/src/schema.ts
mzkrasner: Nice - thank you!

asutula (approved):

Issue #264, opened 2025-05-01
Remove "from" param in SDK methods


dtbuchholz:
See https://github.com/recallnet/rust-recall/pull/271. Depends on wrapper changes in https://github.com/recallnet/contracts/issues/108, or needs the SDK etc to be refactored and have facades replace wrappers via https://github.com/orgs/recallnet/projects/8/views/25?filterQuery=facade&pane=issue&itemId=100156794&issue=recallnet%7Cjs-recall%7C91

stbrody:
Current version of the JS SDK won't work with current IPC until we do this

PR #296, opened 2025-05-09
feat: move to facades, drop wrappers

dtbuchholz:
## Summary

Migrates all wrapper contract dependencies to the facades, which removes the `from` param in many methods, too. Closes https://github.com/recallnet/js-recall/issues/264.

## Details

- `contracts` & `network-constants` now use the built-in, fixed masked ID address for the blobs actor (`0xff00...0042`) and machine/ADM actor (`0xff00...0011`). 
   - The main callout is that you interact with these contracts through interfaces instead needing to deploy the wrapper contract. 
   - If you do any blob or credits operations, that uses the blobs actor (similar to how the blob/credit manager worked.)
   - If you create or list buckets, that uses the machine/ADM actor as a factory contract—but all other bucket ops directly ht the bucket contract. This is a slight change from the previous setup where the bucket mgr wrapper handled everything.
- `sdk` refactored to use the two contracts above, along with the buckets interface. The changes are mostly backwards compatible and shouldn't impact any of the major tooling that uses the JS SDK.
- `react` & `portal` refactored to account for this (primarily, removing the `from` param in various operations).
- `fvm` package refactored to remove a dependency called `uint8arrays`, which is a non-web compatible package. 
   - This package is now used in the `react` package because the facade contracts return blake3 hashes as bytes32, or expect them as bytes32 input values (e.g., iroh node ID, blob hashes). 
   - i.e., conversion from bytes32<>base32 encoded strings is handled at the client layer instead of in the contract logic.


apps/portal/app/account/_components/account.tsx
```diff
@@ -133,12 +133,12 @@ export function Account() {
       });
       return;
     }
-    setAccountSponsor(address, sponsorAddress);
+    setAccountSponsor(sponsorAddress);
```
dtbuchholz, line 136: no more `from` param needed in anything...you'll see this refactored logic thoughout

packages/fvm/src/address/index.ts
```diff
@@ -1,9 +1,6 @@
 // Forked & converted to use `bigint` & `Uint8Array`
 // Source: https://github.com/Zondax/izari-filecoin/blob/master/src/address/index.ts
-import { compare as u8aCompare } from "uint8arrays/compare";
-import { concat as u8aConcat } from "uint8arrays/concat";
-import { fromString as u8aFromString } from "uint8arrays/from-string";
-import { toString as u8aToString } from "uint8arrays/to-string";
+import { Hex, bytesToHex, concatBytes, hexToBytes } from "viem";
```
dtbuchholz, line 3: we replaced the `fvm` package's `uint8arrays` dep with `viem`, which has similar bytes<>hex utilities that we need. this allows the package to work on the web.

packages/fvm/src/utils/convert.ts
```diff
@@ -57,3 +59,59 @@ export function toBigInt(value: string | number | bigint | Uint8Array): bigint {
...
+  return base32.encode(bytes);
+}
+
+/**
+ * Compare two Uint8Arrays
+ * @param a - The first Uint8Array
+ * @param b - The second Uint8Array
+ * @returns The result of the comparison
+ */
+export function compareUint8Arrays(a: Uint8Array, b: Uint8Array): number {
```
dtbuchholz, line 92: note: this was copied from the `uint8arrays` package and tweaked for ts compiler errors

packages/react/src/hooks/buckets.ts
```diff
@@ -12,19 +12,24 @@ import {
...
   const contractAddress =
-    bucketManagerAddress[chainId as keyof typeof bucketManagerAddress];
+    iMachineFacadeAddress[chainId as keyof typeof iMachineFacadeAddress];
 
   const arg = owner ?? address;
 
   return useReadContract({
     address: contractAddress,
-    abi: bucketManagerAbi,
+    abi: iMachineFacadeAbi,
```
dtbuchholz, line 32: example: to list buckets, we use the `iMachineFacadeAbi` interface and hit the machine/ADM actor. but down below, we use the `iBucketFacadeAbi` interface and directly hit the bucket contract

packages/react/src/hooks/buckets.ts
```diff
@@ -100,61 +100,79 @@ export function useQueryObjects(
...
+
+  return {
+    data: data
+      ? {
+          ...data,
+          objects: data.objects.map((obj) => ({
+            ...obj,
+            state: {
+              ...obj.state,
+              blobHash: hexToBase32(obj.state.blobHash),
```
dtbuchholz, line 128: here's where the `fvm` package + base32 decoding comes into play. we have to override the default response from the contract and convert the bytes32 blob hash to a base32 encoded string.

packages/react/src/hooks/buckets.ts
```diff
@@ -287,38 +303,28 @@ export function useAddFile() {
 }
 
 export function useDeleteObject() {
-  const chainId = useChainId();
-  const contractAddress =
-    bucketManagerAddress[chainId as keyof typeof bucketManagerAddress];
-
   const { writeContract, writeContractAsync, ...rest } = useWriteContract();
 
-  const baseConfig = useMemo(
```
dtbuchholz, line 296: this is the only thing i wasnt sure about. i remved the `useMemo` because we no longer have access to the `contractAddress` dependency, and instead, use the `bucket` directly. (idk how `useMemo` really works under the hood and/or if it's ideal to keep here.)

packages/sdk/src/entities/credit.ts
```diff
@@ -128,45 +146,61 @@ export class CreditManager {
...
-      const args = [
-        fromAddress,
-        to,
-        caller,
-        creditLimit,
-        gasFeeLimit,
-        ttl,
-      ] satisfies ApproveCreditParams;
+      // TODO: a few things:
+      // 1. the facades don't allow you to pass a "null" `ttl` param, so we must make it
```
dtbuchholz, line 157: the rest of these changes are around the `sdk` package, so i dont expect a deep review. 

the one callout is there's a small bug, or behavior change, with the `ttl` param in account approvals. previously, you could pass "optional" value as zero, which gets interpreted as null. you can no longer do this.

i've raised the issue and think it's something we should fix, but in the current state, the portal isnt handling this. e.g., if you set any of the optional gas/credit/ttl limits, you have to set them all, or the tx will fail. it's a small feature so i figured we can just keep the logic as-is w/o overthinking it.

dtbuchholz:
@asutula two callouts:
1. we wont merge this until the new testnet goes live
2. there's a bug with `queryObjects` that doesnt let you query _anything_. it's a total blocker but will get fixed at the `ipc` layer. i tested everything else with the `portal` / `sdk` against localnet, and it works.

packages/react/src/utils/base32.ts
dtbuchholz, line 1: note: if for some reason we _dont_ want the `react` package to depend on the `fvm` package, we could restore this file. 

after writing this duplicative code in `react`, i realized i could refactor `fvm` accordingly and eliminate the Node.js dep problem.

dtbuchholz:
@joewagner with aaron going on leave, could you tal at this PR? we'll need it to go live with the new testnet in the testnet (tomorrow/wednesday, i think). these are changes required to move from the solidity wrappers to facade contracts. tests pass locally for the `sdk` package if you run it against the latest from `ipc`, and the same goes for the portal app.

note: this branch's CI _was_ previously passing but looks there's some CI problem.

packages/react/src/hooks/buckets.ts
joewagner: I'm not sure why a const was being memoized, but I think in general it's a perf optimization technique.  Seems ok to me that you removed it.

joewagner (approved):
This looks good.  I'm not really sure what's going on with CI.
My PR seems to be having issues with dependency resolution, but only in CI (i.e. local works fine).

dtbuchholz:
@joewagner gotcha, thanks for the quick review!

also, when a weird CI build errors happen, it seems to be due to a dependency version mismatch. (i saw aaron mention this elsewhere.)

if you run `pnpm list -r viem` locally, you'll see the `comps` app is using a different version than everything else. this PR https://github.com/recallnet/js-recall/pull/310 should, hopefully, fix the CI issues in other PRs

Issue #292, opened 2025-05-08
competitions modals


Rcontre360:
modals and forms for:

- acknowledge terms
- stake $recall
- success

PR #298, opened 2025-05-09
Rafael/comps/modals

Rcontre360:
#292

### DIFF
--- a/apps/comps/components/agent-spotlight-section.tsx
+++ b/apps/comps/components/agent-spotlight-section.tsx
@@ -6,8 +6,7 @@ import {Button} from "@recallnet/ui2/components/button";
 
 import {Agent} from "../data/agents";
 import {AgentCard} from "./agent-card";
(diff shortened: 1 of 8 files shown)


apps/comps/components/trophie-unlocked/index.tsx/index.tsx
dtbuchholz, line 1: the folder should be named `trophie-unlocked`, and not `trophie-unlocked/index.tsx`

apps/comps/components/agent-spotlight-section.tsx
```diff
@@ -16,6 +17,7 @@ export const AgentSpotlightSection: React.FC<AgentSpotlightSectionProps> = ({
 }) => {
   return (
     <section className="my-12">
+      <TermsModal />
```
dtbuchholz, line 20: not that it really matters at this point, but is there a reason to place this here vs importing it directly in `apps/comps/app/competitions/page.tsx`?

PR #310, opened 2025-05-12
chore: bump wagmi, viem, rainbowkit

dtbuchholz:
The `comps` app installed a `viem` version (2.28.0) that did not match with all other packages (2.24.1). This caused a build error with a misleading message. Instead of downgrading `comps`, we might as well upgrade everything else (`viem`, `wagmi`, and `@rainbowkit`).


joewagner:
@dtbuchholz This is great!  I came to a similar fix, but I noticed that (unrelated to deps) the token in one of the tests no longer seems to have a valid value... 😫 
Try replacing line 578 of chain-specific.test.ts with `const targetTokenAddress = "0x0b3e328455c4059eeb9e3f84b5543f74e24e7e1b";`

dtbuchholz:
> @dtbuchholz This is great! I came to a similar fix, but I noticed that (unrelated to deps) the token in one of the tests no longer seems to have a valid value... 😫 Try replacing line 578 of chain-specific.test.ts with `const targetTokenAddress = "0x0b3e328455c4059eeb9e3f84b5543f74e24e7e1b";`

@joewagner ah, yeah that's so weird; it fixed the CI e2e test for me too! i just added that fix to this PR.

joewagner (approved):
Nice! I'll rebase #272 after this gets merge to main

mzkrasner:
> @dtbuchholz This is great! I came to a similar fix, but I noticed that (unrelated to deps) the token in one of the tests no longer seems to have a valid value... 😫 Try replacing line 578 of chain-specific.test.ts with `const targetTokenAddress = "0x0b3e328455c4059eeb9e3f84b5543f74e24e7e1b";`

Yep - dexscreener randomly dropped support for the prev token that was used in the test. Great replacement

Issue #313, opened 2025-05-13
API setup scripts don't work


joewagner:
Initial report below:

While running pnpm setup:all to run the competitions API server locally, I got
```
Error: Password must be at least 8 characters long
ApiError: Password must be at least 8 characters long
    at setupAdmin (file:///Users/brunocalza/projects/js-recall/apps/api/dist/src/controllers/admin.controller.js:37:27)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)

Error setting up admin account: Password must be at least 8 characters long
Warning: Could not extract admin API key from setup output
```

the error message is weird because the password was bigger than 8 characters, and I managed to manually setup an admin by calling the /api/admin/setup endpoint

i managed to get the api running, so this is not a blocker for me, just mentioning

PR #315, opened 2025-05-13
fix: setup:all error when creating admin

brunocalza:
# Context

Fixes: https://github.com/recallnet/js-recall/issues/313
The arguments were not being passed correctly to setup:admin script


joewagner (approved):
Works for me!

PR #316, opened 2025-05-13
feat: allow admin to serach teams based on associated data fields

mzkrasner:
# Summary

- Allows admin to search for teams based on various fields associated with that team's representation in the database
- Includes new methods added to the repository, service, controller, and route layers
- Also includes a new client method in e2e and a supporting e2e test in `admin.test.ts`


joewagner (approved):
Nice work!

PR #320, opened 2025-05-14
feat: expand competitions to include image and external link & teams to include imageurl (mini-app need, plz prioritize)

mzkrasner:
# Summary

- Expands `competitions` top-level definition to include fields for external link and image url (required for Kristal's needs)
- Modifies openapi notation for any routes that include competition information
- Changes to controller and service layers to accommodate optional inputs for these fields
- Changes to app types and e2e types/client handling

- Similarly expands `teams` to use imageUrl (also needed for mini app)
- Similar changes to controller, service, route layers with openapi notation change, e2e test updates, and docs regeneration


apps/api/src/database/schema/core/defs.ts
```diff
@@ -23,6 +23,7 @@ export const teams = pgTable(
     apiKey: varchar("api_key", { length: 400 }).notNull(),
     walletAddress: varchar("wallet_address", { length: 42 }),
     bucketAddresses: text("bucket_addresses").array(),
+    imageUrl: text(),
```
dtbuchholz: for these new column definitions, we should maintain the snakecase format within the actual db since that's how the other columns are defined. i.e., `imageUrl` and `externalLink` should be `imageUrl: text("image_url")` and `externalLink: text("external_link")`

apps/api/src/database/schema/core/defs.ts
```diff
@@ -57,6 +58,8 @@ export const competitions = pgTable(
     id: uuid().primaryKey().notNull(),
     name: varchar({ length: 100 }).notNull(),
     description: text(),
+    externalLink: text(),
```
dtbuchholz: also, do you remember why `externalLink` is required for kristal's work? not a huge issue, but idk if it's needed long-term once the app is live and competition landing pages are hosted in-app vs external links.

dtbuchholz (changes requested):
@mzkrasner one minor change needed, but everything else lgtm

apps/api/src/database/schema/core/defs.ts
mzkrasner: For sure - @KGmajor mentioned that we'd probably have an associated landing page elsewhere in our website assets that provides a more detailed overview of the competition's parameters, etc, and this should be associated with the competitions.

For the mini-app i'm building, for example, i'll need it so when i pull and display upcoming competitions, i have somewhere to link to.

dtbuchholz (approved):

PR #321, opened 2025-05-14
feat: add db tables for votes

ukstv:
Addition of votes.

CC @asutula @mzkrasner @dtbuchholz

### DIFF
--- a/apps/api/src/database/schema/trading/defs.ts
+++ b/apps/api/src/database/schema/trading/defs.ts
@@ -252,3 +252,12 @@ export const votesPerformed = tradingComps.table(
     index("idx_votes_performed_epoch").on(table.epoch),
   ],
 );
+
+export const epochs = tradingComps.table(
+  "epochs",
+  {
+    id: serial().primaryKey(),
+    createdAt: timestamp({ withTimezone: false }).defaultNow(),
+  },
+  (table) => [index("idx_epochs_id").on(table.id)],
+);
--- a/apps/api/src/database/schema/trading/defs.ts
+++ b/apps/api/src/database/schema/trading/defs.ts
@@ -11,6 +11,7 @@ import {
   unique,
   uuid,
   varchar,
+  customType,
 } from "drizzle-orm/pg-core";
 
 import { competitions, teams } from "../core/defs.js";
@@ -40,6 +41,12 @@ function tokenAmount(name?: string) {
     return numeric({ precision: 30, scale: 15, mode: "number" });
   }
 }
+
+const bytea = customType<{ data: Buffer; notNull: false; default: false }>({
+  dataType() {
+    return "bytea";
+  },
+});
 
 export const tradingComps = pgSchema("trading_comps");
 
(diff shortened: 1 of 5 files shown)


apps/api/src/database/schema/trading/defs.ts
```diff
@@ -15,6 +15,32 @@ import {
 
 import { competitions, teams } from "../core/defs.js";
 
+/**
+ * Shared postgres type for an ethereum address.
+ * Note: default length 50 to not introduce any migrations.
```
dtbuchholz, line 21: note: token addresses can SVM or EVM, so `ethereumAddress` is a bit misleading. it should probably be called `tokenAddress`.

@mzkrasner is there a reason we chose 50 as the max length? an ethereum address is 42 length, and a solana address is 44.

apps/api/src/database/schema/trading/defs.ts
```diff
@@ -204,3 +214,50 @@ export const portfolioTokenValues = tradingComps.table(
...
   ],
 );
+
+export const votesAvailable = tradingComps.table(
+  "votes_available",
+  {
+    address: ethereumAddress("address").notNull().primaryKey(),
+    epoch: numeric(),
+    amount: tokenAmount().notNull(),
+    createdAt: timestamp({ withTimezone: false }).defaultNow(),
```
dtbuchholz, line 231: for consistency, these should define the col name in snake case:
- `timestamp("created_at", { withTimezone: false }).defaultNow()`
- `timestamp("updated_at", { withTimezone: false }).defaultNow()`

(and same things for the tables below)

apps/api/src/database/schema/trading/defs.ts
dtbuchholz: related—i suppose our stance on _user_ or _agent_ addresses is that "you must use an EVM style address, irrespective of which chain you're trading on." it's obvious internally (e.g., our `team` wallet addresses are constrained to 42 chars), but it's something we'll need to communicate externally and is relevant to our next ETH v. SOL competition.

dtbuchholz (changes requested):
@ukstv minor changes but otherwise lgtm

apps/api/src/database/schema/trading/defs.ts
mzkrasner: Sorry for the late reply - so this was the original db schema with constraints: https://github.com/recallnet/trading-simulator/blob/c86317b767b2bc41283f7c760c9c74bfd43f95b0/src/database/init.sql#L8

Indeed, at the time we assigned max length of 50 to token_address to be able to ensure we could handle both EVM and SVM. For the `teams` representation, `wallet_address` was assumed to be evm, hence the 42 length

apps/api/src/database/schema/trading/defs.ts
```diff
@@ -204,3 +221,96 @@ export const portfolioTokenValues = tradingComps.table(
...
+    index("idx_votes_available_epoch").on(table.epoch),
+  ],
+);
+
+export const votesPerformed = tradingComps.table(
+  "votes_performed",
+  {
+    address: ethereumAddress("address").notNull().primaryKey(),
+    amount: tokenAmount().notNull(),
+    epoch: numeric(),
```
brunocalza, line 246: any reason for this being numeric and not integer?

apps/api/src/database/schema/trading/defs.ts
dtbuchholz: one other callout—we should mark the `updatedAt` and `createdAt` as `notNull()` in these two tables.

dtbuchholz: oh i just realized another thing:
- we have "core" tables
- we have theme-specific extensions under specific schemas—and "trading" just happens to be the only one we currently run

if we want the staking tables to be global, we should probably put them in their own folder / schema? e.g., we'll have "core", "trading", and then maybe just call it "staking"?

ukstv:
Closed in favor of https://github.com/recallnet/js-recall/pull/428

Issue #323, opened 2025-05-14
Implement a mocked API and integrate the existing pages


ramirotw:
Based on the [API Endpoints document:](https://docs.google.com/document/d/16Zu5-Rdk6DAimdzmK1sLCkFrgDHRCOmPgoWHlgqoaTU/edit?tab=t.0)

* Implement the endpoints using NextJS API routes
* Implement hooks to consume the endpoints
* Integrate the components so they use the hooks and the data coming from the API

PR #325, opened 2025-05-14
feat: add api mock and integrate components

ramirotw:
https://github.com/recallnet/js-recall/issues/323


apps/comps/components/Navbar/index.tsx
```diff
@@ -14,7 +14,7 @@ export const Navbar: React.FunctionComponent<{ children: React.ReactNode }> = ({
 
   const navItems = [
     { label: "COMPETITIONS", href: "/competitions" },
-    { label: "LEADERBOARDS", href: "/leaderboards" },
+    { label: "LEADERBOARDS", href: "/leaderboard" },
```
dtbuchholz: @ramirotw i made a small tweak in https://github.com/recallnet/js-recall/pull/328 to fix the route since it was pointing to a non-existent page. i renamed the page from `leaderboard` to `leaderboards`. So, reverting this results in a 404 when you click the nav.

However, perhaps we should rename the backend `api` app to `/leaderboards` since it currently is `/leaderboard`.

ramirotw: got it, let me change the link and also the API endpoint

dtbuchholz (approved):

PR #327, opened 2025-05-15
chore: remove public routes in favor of registration through mini app

mzkrasner:
# Summary

- This change would remove the public route and its controller and instantiation in the main server file in favor of gated registration only allowable through what is now the mini-app (but might eventually be the main competitions app)
- The flow would instead require the interface running the app to use the admin team registration routes, which would require server-side usage of the admin API key (the admin would be submitting the user's entry from the frontend on their behalf)


dtbuchholz:
we need to remove the [`publicRegisterTeam`](https://github.com/recallnet/js-recall/blob/main/apps/api/e2e/utils/api-client.ts#L793) and [`public.test.ts`](https://github.com/recallnet/js-recall/blob/main/apps/api/e2e/tests/public.test.ts) test too, right?

dtbuchholz (approved):

PR #328, opened 2025-05-15
fix: incorrect `leaderboards` route

dtbuchholz:
The nav defines the route `/leaderboards`, and not the singular `/leaderboard`. The app route/folder is currently called `leaderboard`, so hitting the plural `/leaderboards` results in a 404. This PR renames the route to `leaderboards`.
Screenshot 2025-05-15 at 4 41 47 PM


joewagner (approved):
Good find

Issue #319, opened 2025-05-14
Add GitHub action for building Speakeasy generated SDK package


dtbuchholz:
See [this comment](https://github.com/recallnet/js-recall/pull/272/files/545786b72e44c435cf991bf7393f28a4f870e9bd#r2087253262) for details. We need to have our build action do the following:
1. Set up a Speakeasy user / auth (see [here](https://www.speakeasy.com/docs/speakeasy-reference/cli/getting-started#authenticate) for how the CLI works)
2. Run the `packages/api-sdk/.devcontainer/setup.sh` script

The main reason we can't run it as part of the standard `pnpm build` process is because of point (1).

stbrody:
@sam701 is this something you could pick up?

stbrody:
blocked on https://github.com/recallnet/js-recall/pull/272 merging first

dtbuchholz:
@sam701 added some additional context. let me know if that helps!

i think all we really need to do is set up the github action—and they have a bot we can install [here](https://github.com/apps/speakeasy-github). alternatively, we could choose to run the `build-api-sdk.sh` from root, which would require us to create an action that downloads/uses the Speakeasy CLI and then runs the script.

we don't need to publish the package to NPM (yet), so the CI/CD is purely to ensure that the `packages/api-sdk` code is up to date based on the OpenAPI spec in `apps/api`.

@joewagner after playing with Speakeasy a bit more, i guess the Speakeasy CI/CD workflow described here is, basically, just for type safety & compatibility checks. e.g., if i make changes locally to the OpenAPI spec in `apps/api`, i _should_ have to regenerate the Speakeasy SDK because the `apps/api` tests depend on `packages/api-sdk`. the scenarios where failing tests isnt true:
- you remove endpoints from the OpenAPI spec and the associated `apps/api` tests -> the `packages/api-sdk` could include deprecated code
- you add a new endpoint or params to the OpenAPI spec, but dont write tests for them -> the `packages/api-sdk` might not include the new features

anyways, idk if you have any thoughts on what i've described in this issue but figured i'd get your input before @sam701 moves forward.

dtbuchholz:
(i dont have anywhere else to post this—just using this issue for a future reference.)

speakeasy also lets you [override the `gen.yaml`](https://www.speakeasy.com/guides/sdks/override-compile#overriding-compile-commands-for-sdk-generation):
```
typescript:
   compileCommand:
    - pnpm
    - install
```
[...]

PR #329, opened 2025-05-15
chore: regenerate sdk with removed public endpoint

dtbuchholz:
- Rebuilds the competitions SDK with the latest changes, which remove the `/public` endpoint
- Ignores the `package-lock.json` file
- Fixed package versions; the `package.json` file had 0.0.0 whereas the `gen.yaml` & `jsr.json` had 0.0.6


packages/api-sdk/.speakeasy/gen.yaml
```diff
@@ -19,7 +19,7 @@ generation:
     oAuth2ClientCredentialsEnabled: true
     oAuth2PasswordEnabled: true
 typescript:
-  version: 0.0.6
+  version: 0.0.0
```
dtbuchholz, line 22: note: the version gets auto-bumped based on the `gen.yaml` version. we could either:
1. manually set the version as 0.0.0 whenever we manually build + PR `api-sdk` changes
2. just let the semver auto-bump

while the package used internally and isnt published, we could probably just do (2). once we're ready to publish it, we can reset it to 0.0.0 and then just do (2) since it will naturally follow semver updates.

packages/api-sdk/.speakeasy/gen.yaml
```diff
@@ -19,13 +19,14 @@ generation:
...
   additionalDependencies:
     dependencies: {}
     devDependencies: {}
     peerDependencies: {}
-  additionalPackageJSON: {}
-  author: Speakeasy
+  additionalPackageJSON:
+    private: true
```
dtbuchholz, line 28: note: i had to add `private: true` because the package got published to npm due to the (odd) changesets behavior. (i've deleted it from the registry.)

dtbuchholz:
cc @mzkrasner if you make any changes to the OpenAPI, you'll need to:
1. sign up for a speakeasy account (see [here](https://login.speakeasy.com/))
2. download the CLI (see [here](https://www.speakeasy.com/docs/speakeasy-reference/cli/getting-started))
3. run `sh build-api-sdk.sh` from root
(idk if you've set up speakeasy yet, but you can run these steps w/o needing to be on the same team as other ppl.)

joewagner (approved):
@dtbuchholz Sorry I didn't get to this sooner.  Looks good.
We can work on the changes needed to publish the api sdk correctly in #319

dtbuchholz:
all good, and yeah sgtm

PR #339, opened 2025-05-19
fix: auth integration

ramirotw:
This PR improves the mocked auth integration by:

* moving the endpoints to the correct path: /api -> /api/auth
* saving the auth state in a cookie, later to changed to a jwt once implemented on the backend
* implementing a basic /profile endpoint which mainly checks the presence of the cookie
* storing the client atom state in localstorage so refreshing the page doesn't lose the logged state
* adding the logout endpoint


dtbuchholz:
note: im looking into the `registration` vercel CI flow...not sure why it's getting triggered here, but ignore it

dtbuchholz (approved):

Issue #330, opened 2025-05-16
Fix flaky CI (`react` package's wagmi types & deadlock problem)


dtbuchholz:
The CI build step and `react` package sometimes fails, but if you re-run it, it succeeds. There's some weird type issue with `wagmi`, and it's not clear why this happens. It's been observed in different GH actions like the e2e API tests and the "Check lint, format, and build" action, which both run this build step. 

The error calls out `useCreateBucket` and `useDeleteObject` and the lack of a type, which causes type inference issues.

```
> @recallnet/react@0.0.0 build /home/runner/work/js-recall/js-recall/packages/react
> tsup

CLI Building entry: src/hooks/buckets.ts, src/hooks/credits.ts, src/actions/credits.ts
CLI Using tsconfig: tsconfig.json
CLI tsup v8.4.0
CLI Using tsup config: /home/runner/work/js-recall/js-recall/packages/react/tsup.config.ts
CLI Target: es2022
CLI Cleaning output folder
CJS Build start
ESM Build start
CJS dist/hooks/buckets.cjs   9.46 KB
CJS dist/actions/credits.cjs 2.32 KB
CJS dist/hooks/credits.cjs   10.93 KB
CJS ⚡️ Build success in 71ms
ESM dist/chunk-MC7QZ4OM.js  1.20 KB
ESM dist/hooks/credits.js   7.96 KB
ESM dist/hooks/buckets.js   7.13 KB
ESM dist/actions/credits.js 118.00 B
ESM ⚡️ Build success in 71ms
DTS Build start
Error: src/hooks/buckets.ts([164](https://github.com/recallnet/js-recall/actions/runs/15071394124/job/42368382111?pr=325#step:9:172),17): error TS2742: The inferred type of 'useCreateBucket' cannot be named without a reference to '.pnpm/@wagmi+core@2.17.1_@tanstack+query-core@5.69.0_@types+react@19.0.12_react@19.0.0_typescript@5_qapkphp6jpkabu5m5w6tf5dkjy/node_modules/@wagmi/core'. This is likely not portable. A type annotation is necessary.
[...]

PR #347, opened 2025-05-19
feat: add registration mini app for competitions sign-up process

mzkrasner:
# Summary

- Intends to onboard users by registering them to the sandboxed competitions API by serving a frontend, with nextjs routes and auth provider patterns to verifiably confirm user ownership of primary wallet address
- NOTE: this frontend is designed to work specifically with the version of the API from [this branch](https://github.com/recallnet/js-recall/tree/mzk/public), which auto-adds a user to an active competition so they can immediately start testing trades, etc.
- It WILL NOT work without this feature as API calls to do things like update profile (to add agents, for example) fails without an active competition running


dtbuchholz (approved):
lgtm. i've updated the vercel deployment config, so this should push an update to the site once merged to main.

note: we can ignore the e2e CI issue wrt deadlock (described [here](https://github.com/recallnet/js-recall/issues/330#issuecomment-2891593101))

Issue #266, opened 2025-05-01
Fix handling of very small or very big numbers


mzkrasner:
# Summary

- Current database schema do not sufficiently take into account prospective numbers that could reflect extremely large token balances (for example, if 5,000 USDC is swapped for a token that's worth $2.938e-27:

```
CREATE TABLE IF NOT EXISTS balances (
  id SERIAL PRIMARY KEY,
  team_id UUID NOT NULL REFERENCES teams(id) ON DELETE CASCADE,
  token_address VARCHAR(50) NOT NULL,
  amount DECIMAL(30, 15) NOT NULL,
  created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
  updated_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
  specific_chain VARCHAR(20), -- New column to track specific chain (eth, polygon, base, etc.)
  UNIQUE(team_id, token_address)
);
```

- In this case, the `amount` field would not be able to handle this type of value

## Options

- Use string representations and handle conversions application layer
- Use `NUMERIC` type which can handle numbers with up to 16,838 digits after decimal points, and large numbers (8.37 x 19^29). For example:

```
CREATE TABLE IF NOT EXISTS trades (
  id UUID PRIMARY KEY,
  team_id UUID NOT NULL REFERENCES teams(id) ON DELETE CASCADE,
  competition_id UUID NOT NULL REFERENCES competitions(id) ON DELETE CASCADE,
  from_token VARCHAR(50) NOT NULL,
  to_token VARCHAR(50) NOT NULL,
  from_amount NUMERIC NOT NULL,
  to_amount NUMERIC NOT NULL,
  price NUMERIC NOT NULL,
  -- rest of columns
);
```

dtbuchholz:
reference implementations:
- our blockscout indexer db uses `numeric` (or `numeric(100,0)`) for token amounts (total supply, decimals, tx values, etc)
- [dune's token balances](https://docs.dune.com/echo/evm/balances#column-descriptions) store token prices/amounts as `double`, but its [transactions table](https://docs.dune.com/echo/evm/transactions) uses a hex string for tx values.

asutula:
FWIW, once we migrate to drizzle, `numeric` db types can be automatically converted to `bigint` in the JS app. And vice versa. 

[...]

PR #356, opened 2025-05-19
fix: small token values can be used in trades

joewagner:
fixes: #266


joewagner:
@mzkrasner In hopes of just getting this fix out, I haven't included any tests.

I've tried a few different paths, but because the actual server under test in the e2e tests is in a child process, its difficult to mock, or stub out, methods.  Specifically the `priceTracker.getPrice` method needs to be stubbed in order to enforce that it returns the small number that caused the issue described in #266

I chatted with @stbrody today, and he suggested that we work on getting the fix out and when time allows for lower priority tasks we can revisit the changes to the testing framework.  I have a branch here https://github.com/recallnet/js-recall/tree/joe/small-num-test that demonstrates this issue with stubbing inside subprocesses.

I did test these changes manually by hacking the getPrice method to return a really small number for a specific tokenAddress.  However, we can't do that for the production code so it's removed here.

apps/api/src/services/trade-simulator.service.ts
```diff
@@ -166,16 +166,20 @@ export class TradeSimulator {
...
 
-      console.log(`[TradeSimulator] Got prices:
-        From Token (${fromToken}): $${fromPrice} (${fromTokenChain})
-        To Token (${toToken}): $${toPrice} (${toTokenChain})
-    `);
+      console.log("[TradeSimulator] Got prices:");
+      console.log(
+        `  From Token (${fromToken}): ${JSON.stringify(fromPrice, null, 4)} (${fromTokenChain})`,
+      );
+      console.log(
```
joewagner, line 180: I moved this to three log calls because the multi-line string was not being displayed correctly.
I _think_ this has to do with the way stdout of the child process is piped into the parent stdout, but regardless this works for me.  

Also, `fromPrice` and `toPrice` are Objects so they were displayed as `[object Object]`.  I changed this so they are stringified.

apps/api/src/database/schema/trading/defs.ts
```diff
@@ -152,6 +148,7 @@ export const portfolioSnapshots = tradingComps.table(
     teamId: uuid("team_id").notNull(),
     competitionId: uuid("competition_id").notNull(),
     timestamp: timestamp({ withTimezone: true }).defaultNow(),
+    // TODO: are units of this number usdc? if so, the precision and scale are good here. if not, need to remove.
```
joewagner, line 151: This precision seems ok, but left the comment just in case.

apps/api/drizzle/0002_fair_stryfe.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE "trading_comps"."balances" ALTER COLUMN "amount" SET DATA TYPE numeric;--> statement-breakpoint
+ALTER TABLE "trading_comps"."portfolio_token_values" ALTER COLUMN "amount" SET DATA TYPE numeric;--> statement-breakpoint
+ALTER TABLE "trading_comps"."portfolio_token_values" ALTER COLUMN "price" SET DATA TYPE numeric;--> statement-breakpoint
+ALTER TABLE "trading_comps"."prices" ALTER COLUMN "price" SET DATA TYPE numeric;--> statement-breakpoint
+ALTER TABLE "trading_comps"."trades" ALTER COLUMN "from_amount" SET DATA TYPE numeric;--> statement-breakpoint
+ALTER TABLE "trading_comps"."trades" ALTER COLUMN "to_amount" SET DATA TYPE numeric;--> statement-breakpoint
+ALTER TABLE "trading_comps"."trades" ALTER COLUMN "price" SET DATA TYPE numeric;
```
joewagner, line 7: We will need to run `npm run db:migrate` on any existing database.

apps/api/e2e/tests/trading.test.ts
```diff
@@ -1532,4 +1537,61 @@ describe("Trading API", () => {
...
+      competitionName,
+      [team.id],
+    );
+
+    // Wait for balances to be properly initialized
+    await wait(500);
+
+    const smallValue = 2.938e-27;
+
+    const trade: InsertTrade = {
```
joewagner, line 1565: These values a a bit arbitrary, but the important thing is that the number fields are all set to the value that caused the issue during the last comp

mzkrasner (approved):
lgtm! appreciate the additional fix for printing the `fromPrice` and `toPrice`

PR #365, opened 2025-05-20
feat: add value usd to trades

mzkrasner:
# Summary

- Marketing needs an easy way to reference the size of a trade in USD, and adding this into each trade entry from the API is easier and likely removes the ability for calculation errors to manifest if done by querying the sql db ad-hoc after the fact
- This change updates the `trades` schema to include `trade_amount_usd`, as well as corresponding route, controller, and service layers to handle the calculation, storage, and remittence of said value


joewagner (approved):
Looks good, only thing would regenerating the api's sdk.  I am setup to do that, so I can handle it in another PR if that's easiest?

Issue #342, opened 2025-05-19
Improve components integration with remote/local state


ramirotw:
After https://github.com/recallnet/js-recall/pull/325 got merged, there are still pieces that need better integration with the remote and local data, such as:
- [x] auth routes and the logged in/out states (https://github.com/recallnet/js-recall/pull/339)
- [ ] competitions/agents tables to support pagination, sorting and filtering
- [x] sync logged in state with components

PR #369, opened 2025-05-20
feat: make agents table to support pagination, sorting and filtering

ramirotw:
https://github.com/recallnet/js-recall/issues/342


dtbuchholz:
@ramirotw a few observations:
1. for the table headers/tabs (e.g., at `/agents/660e8400-e29b-41d4-a716-446655440001`), could we highlight the "active" tab that's being clicked? 
Screenshot 2025-05-21 at 5 44 29 PM

2. it doesnt seem like the sorting is working (e.g., `/competitions/550e8400-e29b-41d4-a716-446655440007`); the order of the list remains unchanged
Screenshot 2025-05-21 at 5 54 46 PM

3. similarly, the search correctly updates the header, but the table items are unimpacted
Screenshot 2025-05-21 at 5 55 43 PM

since this issue is slightly lower priority now, we can create a follow up to make these tweaks. i just wanna double check if any of this is a byproduct of the mocked data etc

ramirotw:
> @ramirotw a few observations:
> 
> 1. for the table headers/tabs (e.g., at `/agents/660e8400-e29b-41d4-a716-446655440001`), could we highlight the "active" tab that's being clicked?
> 
> Screenshot 2025-05-21 at 5 44 29 PM

the problem is that the Tabs component was updated to support the new design, so the underline state we had before was removed. Now the app is in a intermediate state where we have some pages/components updated with the new designs and the old ones which weren't updated. This problem will be fixed once @jovirecall consolidate the designs and we update everything in one go.
image


> 2. it doesnt seem like the sorting is working (e.g., `/competitions/550e8400-e29b-41d4-a716-446655440007`); the order of the list remains unchanged
> 
> Screenshot 2025-05-21 at 5 54 46 PM
> 3. similarly, the search correctly updates the header, but the table items are unimpacted
> 
> Screenshot 2025-05-21 at 5 55 43 PM
> since this issue is slightly lower priority now, we can create a follow up to make these tweaks. i just wanna double check if any of this is a byproduct of the mocked data etc

please re-test the table filtering, sorting and pagination again

PR #386, opened 2025-05-21
Rafael/user agent list

Rcontre360:
solves #368


ramirotw:
changes already included in https://github.com/recallnet/js-recall/pull/390

PR #387, opened 2025-05-21
feat: integrate api-mcp

mzkrasner:
# Summary

- This PR adds a new @recallnet/api-mcp package that provides a Model Context Protocol (MCP) server for interacting with the Recall API. 
- This allows AI assistants like Claude to access the Competitions API via MCP.

## Key Features

- User-focused API Access: Exposes only non-admin API endpoints to regular users
- Reusable Components: Leverages existing tools and infrastructure from @recallnet/api-sdk

## Implementation Details

- Created a TypeScript implementation that imports tool definitions directly from the SDK
- Established a filtered set of allowed tools that excludes admin-only operations
- Added environment variable support (API_KEY, API_SERVER_URL, LOG_LEVEL)
- Implemented command-line parsing with minimist
- Added documentation for Cursor and Claude Desktop integration


packages/api-mcp/src/build.mts
dtbuchholz, line 1: what does this file do? i dont think we need it since we're using tsup to build

mzkrasner: ah ty - forgot to remove it

packages/api-mcp/src/env.ts
```diff
@@ -0,0 +1,28 @@
...
+  LOG_LEVEL: process.env.LOG_LEVEL || "info",
+};
+
+/**
+ * Validate required environment variables
+ * @returns {string[]} Array of missing required variables
+ */
+export function validateEnv(): string[] {
+  // Currently no required variables, only optional ones
+  // This could be expanded if required variables are added later
```
dtbuchholz: just checking—why aren't env vars required? aside from the `/health` endpoint, i think basically everything else uses auth, right?

packages/api-mcp/src/cli.ts
```diff
@@ -0,0 +1,49 @@
...
+  // Validate environment
+  const missingVars = validateEnv();
+  if (missingVars.length > 0) {
+    console.error(
+      `Missing required environment variables: ${missingVars.join(", ")}`,
+    );
+    process.exit(1);
+  }
+
+  const argv = minimist(process.argv.slice(2), {
```
dtbuchholz, line 21: should we keep the flag definitions the same as the env vars we define later? e.g., `--server-url` & `SERVER_URL`, etc

packages/api-sdk/src/hooks/hooks.ts
```diff
@@ -2,6 +2,7 @@
  * Code generated by Speakeasy (https://speakeasy.com). DO NOT EDIT.
  */
 import { RequestInput } from "../lib/http.js";
+import { initHooks } from "./registration.js";
```
dtbuchholz, line 5: i think this is legacy code from the registration endpoint? my most recent `api-sdk` speakeasy generation removed it, and this looks to be adding it back in

dtbuchholz: actually idk. not sure what the diff is from...i think we're using the same speakeasy CLI version.

packages/api-mcp/src/cli.ts
mzkrasner: gc - just pushed this change

dtbuchholz:
few small changes but gave it a test locally, lgtm.

note: i _totally_ forgot that speakeasy generates MCP server code...but i suppose having a bit more control like this doesnt hurt. 

speakeasy's out-of-the-box MCP tool definitions seem to be slightly less usable compared to what we built in the past. e.g., it doesnt nicely describe `args` in the tool call, so the LLM has to partially infer them from the variable name. you'll see what i mean if you use the MCP model inspector CLI along with this binary.

anyways, that's not anything we need to deal with here and just something to keep an eye out on wrt DX.

packages/api-mcp/src/env.ts
mzkrasner: Agreed - covered in [a9f4284](https://github.com/recallnet/js-recall/pull/387/commits/a9f4284abc0c8c4f1c64803795d346d25f3ff8fb)

dtbuchholz (approved):

Issue #388, opened 2025-05-21
User section on dashboard


Rcontre360:
section above agent list

PR #390, opened 2025-05-22
Rafael/user info dashboard

Rcontre360:
#388


ramirotw (approved):
lgtm, the final integration with the correct data will be done in a follow-up PR

Issue #322, opened 2025-05-14
bug: root encryption key is not being correctly loaded into config and process.env after admin setup


mzkrasner:
# Summary

- When the admin setup route is hit, the controller's setupAdmin method is meant to write a `ROOT_ENCRYPTION_KEY` to the .env, as well as load it into config.security.rootEncryptionKey

```
 async setupAdmin(req: Request, res: Response, next: NextFunction) {
      try {
        // Check if any admin already exists
        const teams = await findAll();
        const adminExists = teams.some((team) => team.isAdmin === true);

        if (adminExists) {
          throw new ApiError(
            403,
            "Admin setup is not allowed - an admin account already exists",
          );
        }

        // Validate required parameters
        const { username, password, email } = req.body;
        if (!username || !password || !email) {
          throw new ApiError(
            400,
            "Missing required parameters: username, password, email",
          );
        }

        // Validate password strength
        if (password.length < 8) {
          throw new ApiError(
            400,
            "Password must be at least 8 characters long",
          );
        }

        // Ensure that ROOT_ENCRYPTION_KEY exists in .env file
        try {
          // Find .env file in app root directory
          const envPath = path.resolve(process.cwd(), ".env");
          console.log(
            `[AdminController] Checking for .env file at: ${envPath}`,
          );

          if (fs.existsSync(envPath)) {
            const envContent = fs.readFileSync(envPath, "utf8");
            const rootKeyPattern = /ROOT_ENCRYPTION_KEY=.*$/m;

            // Check if ROOT_ENCRYPTION_KEY already exists and is not the default
            const keyMatch = rootKeyPattern.exec(envContent);
            let needsNewKey = true;

            if (keyMatch) {
              const currentValue = keyMatch[0].split("=")[1];
              if (
                currentValue &&
                currentValue.length >= 32 &&
                !currentValue.includes("default_encryption_key") &&
                !currentValue.includes("your_") &&
                !currentValue.includes("dev_") &&
                !currentValue.includes("test_") &&
                !currentValue.includes("replace_in_production")
              ) {
                // Key exists and seems to be a proper key already
                console.log(
                  "[AdminController] ROOT_ENCRYPTION_KEY already exists in .env",
                );
                needsNewKey = false;
              }
            }

            if (needsNewKey) {
              // Generate a new secure encryption key
              const newEncryptionKey = crypto.randomBytes(32).toString("hex");
              console.log(
                "[AdminController] Generated new ROOT_ENCRYPTION_KEY",
[...]

PR #395, opened 2025-05-22
fix: resolve root encryption key not being applied to server config issue

mzkrasner:
# Fix ROOT_ENCRYPTION_KEY Not Being Reloaded After Admin Setup

## Problem
When admin setup generated a new `ROOT_ENCRYPTION_KEY` and wrote it to the `.env` file specifically using the designated route (as opposed to the script), the application's in-memory configuration wasn't updated to use the new key. This caused a major security hole given that previous production instances had mistakenly used the default encryption key.

## Root Cause
The issue occurred in the admin setup flow (`/api/admin/setup`):
- Admin setup generated a new 32-byte hex encryption key
- Updated the `.env` file and `process.env.ROOT_ENCRYPTION_KEY`  
- But never reloaded the application's configuration object
- Teams registered afterward had API keys encrypted with the new key but couldn't authenticate because validation used the old cached key

### 1. Added Configuration Reload Function
- **File**: `apps/api/src/config/index.ts`
- **Added**: `reloadSecurityConfig()` function that updates in-memory config when encryption key changes
- **Purpose**: Ensures the application uses the updated `ROOT_ENCRYPTION_KEY` immediately

### 2. Updated Admin Controller
- **File**: `apps/api/src/controllers/admin.controller.ts` 
- **Added**: Call to `reloadSecurityConfig()` after updating the encryption key
- **Fixed**: Write to correct `.env` file (`.env.test` in test mode, `.env` in production)
- **Result**: Configuration is properly reloaded when admin setup generates new encryption key

### 3. Added E2E Test
- **File**: `apps/api/e2e/tests/admin.test.ts`
- **Test**: `should generate and use new encryption key after admin setup`
- **Validates**: 
  - New encryption key is generated (different from default)
  - Admin API key cannot be decrypted with default key (proves new key was used)
  - The decryption failure (`bad decrypt` error) is the success condition

## Changes Made

```typescript
// config/index.ts - Added reload function
export function reloadSecurityConfig(): void {
  const newRootKey = process.env.ROOT_ENCRYPTION_KEY || "default_encryption_key_do_not_use_in_production";
  config.security.rootEncryptionKey = newRootKey;
  config.app.sessionPassword = newRootKey;
  console.log("[Config] Security configuration reloaded with updated ROOT_ENCRYPTION_KEY");
}

// admin.controller.ts - Added reload call
process.env.ROOT_ENCRYPTION_KEY = newEncryptionKey;
reloadSecurityConfig(); // 🆕 This ensures config uses the new key immediately

// admin.controller.ts - Fixed env file selection
const envFile = process.env.NODE_ENV === "test" ? ".env.test" : ".env";
const envPath = path.resolve(process.cwd(), envFile);
```

## Testing
The e2e test proves the fix works by:
1. Checking that `process.env.ROOT_ENCRYPTION_KEY` differs from the default
2. Attempting to decrypt the admin's API key with the default encryption key
3. **Expecting this to fail** (throw "bad decrypt" error)
4. The decryption failure confirms a new key was generated and used

## Impact
- ✅ **Maintains** security by ensuring unique encryption keys per deployment  
- ✅ **Preserves** existing behavior for normal operations
- ✅ **Adds** test coverage for encryption key generation workflow

## Files Changed
- `apps/api/src/config/index.ts` - Added `reloadSecurityConfig()` function
- `apps/api/src/controllers/admin.controller.ts` - Added config reload and fixed env file selection
- `apps/api/e2e/tests/admin.test.ts` - Added test to validate encryption key generation
- `.env.test` - Removal of default `ROOT_ENCRYPTION_KEY`

Resolves [#322 ](https://github.com/recallnet/js-recall/issues/322)


joewagner (approved):
Nice work!

joewagner (approved):
:rocket:

PR #397, opened 2025-05-22
feat: new navbar and responsive design

Rcontre360:
#396


apps/comps/components/siwe/index.tsx
```diff
@@ -12,60 +14,64 @@ import {
...
+                  onClick={handleMyAccount}
+                  className="cursor-pointer hover:bg-gray-800"
+                >
+                  My Account
+                </DropdownMenuItem>
                 <DropdownMenuItem
                   onClick={handleLogout}
                   className="cursor-pointer hover:bg-gray-800"
                 >
+                  <FaArrowRightFromBracket className="h-10 w-10 text-gray-600" />
```
ramirotw, line 73: why not the `ExitIcon` from @radix-ui/react-icons?

apps/comps/package.json
```diff
@@ -36,10 +37,10 @@
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
     "react-hook-form": "^7.54.2",
+    "react-icons": "^5.5.0",
```
ramirotw: let's try to keep using @radix-ui/react-icons until we define an icon library with @jovirecall

apps/comps/components/siwe/index.tsx
```diff
@@ -12,60 +14,64 @@ import {
...
-        setUser({ loggedIn: false, address: "" });
-      },
-    });
+  const handleLogout = async () => {
+    logout.mutate();
+    await refetch();
+  };
+
+  const handleMyAccount = () => {
+    router.push("/user/agents");
```
ramirotw, line 33: when we merge https://github.com/recallnet/js-recall/pull/400 this link should be updated to `/profile`

apps/comps/components/siwe/index.tsx
Rcontre360: We can use any, still we need font awesome for some other icons that we dont have in other parts of the ui

apps/comps/package.json
Rcontre360: let me remove that one, I replaced it for font awesome

Rcontre360: there are many that are not in radix-ui (Very limited)

ramirotw (changes requested):
it's looking good.

A few comments:
1. apart from the typescript issue in `components/siwe/index.tsx:45:21`, there are some other problems  marked in the screenshot:
  a. wallet default image
  b. duplicated `My Account` item
  c. vertical misalignment of the navbar with the main section
  d. separation between the navbar and the main section

image

PR #398, opened 2025-05-22
feat(api): add new `competitions` (plural) endpoint

joewagner:


joewagner:
incorporated into #404

Issue #380, opened 2025-05-21
[FEATURE] Create User account page


ramirotw:
**Problem Statement**
When the user logs in via SIWE, they should be redirected to the Create User page to complete their profile information, otherwise, they won’t be able to view their profile.

**Proposed Solution**

![Image](https://github.com/user-attachments/assets/b8785039-86bc-4529-a54c-1eb2516b44bf)

**Priority**
High

Issue #331, opened 2025-05-16
Connect frontend (`apps/comps`) to backend (`apps/api`)


dtbuchholz:
Depends on:
- [x]  Frontend mock APIs via #323 
- [ ] Backend API changes to match mock requirements via:
   - [ ]  #314
   - [ ]  #250 

Once those are completed, we can wire up the frontend/backend.

PR #400, opened 2025-05-23
feat: add user profile page

ramirotw:
https://github.com/recallnet/js-recall/issues/380

image

image


apps/comps/app/api/auth/login/route.ts
```diff
@@ -57,7 +57,7 @@ export async function POST(req: NextRequest) {
       wallet: siweMessage.address,
     });
 
-    response.cookies.set("wallet", siweMessage.address, {
+    response.cookies.set("wallet_address", siweMessage.address, {
```
dtbuchholz, line 60: note: i changed this because our backend is setting `wallet` via https://github.com/recallnet/js-recall/blob/e4c22a03951ab3f1ec9ae32fbb0dfeaf15909901/apps/api/src/types/index.ts#L281

i dont have a strong preference and can clean it up once we connect the app to the api in #331

dtbuchholz (approved):

apps/comps/app/api/auth/login/route.ts
ramirotw: makes sense, once we integrate with the backend we'll probably remove the local api

Issue #405, opened 2025-05-23
Migrate tables and API for updated agent<>user design


dtbuchholz:
The frontend/designs need the backend to change, as per #314. Thus, we need to spec out (see #333) and implement database changes that move from a team centric design to users + agents.

mzkrasner:
# Implementation Plan: Teams → Users + Agents Architecture

## Overview

This document outlines the comprehensive changes required to migrate from the current `teams`-based architecture to a new `users` + `agents` architecture. The implementation plan involves splitting the current `teams` table into two separate entities:

- **Users**: Represent the actual human users who own agents
- **Agents**: Represent the AI agents that participate in competitions and perform trading activities

## Current State Analysis

### Current Schema Structure
- `teams` table serves as the primary entity for all operations
- All trading activity, competition participation, and authentication is tied to `teamId`
- Tables with foreign keys to `teams`:
  - `competition_teams.team_id` → `teams.id`
  - `trading_comps.balances.team_id` → `teams.id`
  - `trading_comps.trades.team_id` → `teams.id`
  - `trading_comps.portfolio_snapshots.team_id` → `teams.id`

### Current Application Layers
- **Repository Layer**: 5 repositories with team-centric operations
- **Service Layer**: 9 services with team-based business logic
- **Controller Layer**: 3 main controllers handling team operations
- **Middleware Layer**: Authentication tied to team API keys
- **Type System**: Extensive use of `teamId` throughout interfaces

---

## 1. DATABASE SCHEMA CHANGES

### 1.1 New Tables to Create

#### Users Table
```sql
CREATE TABLE "users" (
  "id" uuid PRIMARY KEY NOT NULL,
  "wallet_address" varchar(42) UNIQUE NOT NULL,
  "name" varchar(100),
  "email" varchar(100),
  "image_url" text,
  "is_admin" boolean DEFAULT false NOT NULL,
  "metadata" jsonb,
  "status" varchar(20) DEFAULT 'active' NOT NULL, -- 'active', 'suspended', 'deleted'
  "created_at" timestamptz NOT NULL DEFAULT now(),
  "updated_at" timestamptz NOT NULL DEFAULT now()
);
```

#### Agents Table
```sql
CREATE TABLE "agents" (
  "id" uuid PRIMARY KEY NOT NULL,
  "owner_id" uuid NOT NULL REFERENCES "users"("id") ON DELETE CASCADE,
  "wallet_address" varchar(42) UNIQUE, -- nullable for now
[...]

Issue #314, opened 2025-05-13
Implement API endpoints for comp app


joewagner:
Build the endpoints described here https://docs.google.com/document/d/16Zu5-Rdk6DAimdzmK1sLCkFrgDHRCOmPgoWHlgqoaTU

related to #290

dtbuchholz:
@joewagner after thinking through this a bit more, perhaps we break this issue into sub-issues. e.g., we could start with the competitions methods, then leaderboards, then agents, then profile. (note: `/auth` is mine via #250.)

the intent is to allow us to do #331 iteratively instead of a big bang approach. i.e., the goal is to see if there's anything we could wire up this week and slowly turn on the frontend<>backend connection vs all at once.

joewagner:
> [@joewagner](https://github.com/joewagner) after thinking through this a bit more, perhaps we break this issue into sub-issues. e.g., we could start with the competitions methods, then leaderboards, then agents, then profile. (note: `/auth` is mine via [#250](https://github.com/recallnet/js-recall/issues/250).)
> 
> the intent is to allow us to do [#331](https://github.com/recallnet/js-recall/issues/331) iteratively instead of a big bang approach. i.e., the goal is to see if there's anything we could wire up this week and slowly turn on the frontend<>backend connection vs all at once.

That sounds great!  
Should i do that, or do you want to create the issues?  I'll start with the competition endpoints today.

dtbuchholz:
> Should i do that, or do you want to create the issues? I'll start with the competition endpoints today.

i can map em out, no worries. and cool, sg

dtbuchholz:
@joewagner i saw your comment about `zod` and request parsing. Good idea: https://github.com/recallnet/js-recall/issues/426

I'm thinking we can slowly handle this feature as we refactor each of the APIs in this issue?

PR #418, opened 2025-05-25
feat(api): schema refactor

dtbuchholz:


dtbuchholz:
@mzkrasner @joewagner i want to put this on your radar; h/t mark for the initial work and schema changes. this PR implements the agents<>users design and removes all nomenclature related to "teams."

i (mostly) finished what he started and refactored the codebase/tests to support the frontend app's requirements. i still have some fixes to make around user auth, and the APIs still need some work but #314 accounts for this (and should now be way easier). note: tests pass locally if you build the `api-sdk`, but i want to avoid committing that code until this is in a ready state.

this PR is massive, and it's definitely _not_ the ideal way to handle code changes. however, we're trying to shift to a "move fast and break things" mentality. given we want to ship the first iteration of the app asap—and it doesnt have to be perfect—i think this is one of the rare times where this is okay. a few reasons:
- we started with a backend, and then the frontend design requirements came along. the backend and OpenAPI spec didn't align to these and needed a revamp. square peg, round hole scenario. the only way to address it was to, basically, fork + rewrite the backend. we don't have any real users yet, so the migration process can be sorta ad-hoc.
- we should've prioritized this work earlier, but in the spirit of move fast + break things, we can use this work to ship the app on time...or at least, get pretty close. our goal is to launch something asap, so i figure this PR can help us do that while we adopt our new processes in the next iterations.
- because of the points above, our internal processes for building this app have been messy. moving forward, i'll make sure this is better. this stop-gap redesign will help us ship—and it's also a step in the right direction get to the point where we can "do things right" with the newer agents<>users setup. although i would have preferred to start with this "do things right" approach, the next phases of the app are focused exactly on that (production readiness, spec-first design, etc.). speed now, think later.

dtbuchholz:
if yall are against this approach, i'm totally open to other ideas. to help clarify my future thinking (and echo some of aaron's [forward-facing thoughts](https://www.notion.so/Competitions-App-Implementation-Notes-1eddfc9427de80549c61f15029305154?pvs=4)), these are the top themes/workstreams we should prioritize next (after we get the frontend<>backend wired up and ship v1):
1. we need to shift to a spec-first development pattern.
   - spec-ing out the the API will make sure our frontend and backend development patterns agree on the data model before we build anything, and it'll allow us to better break up the work into vertical slices to iteratively add functionality.
   - it will also help out with the sort of circular `apps/api`<>`packages/api-sdk` dependency. it'll let us shift our APIs to be more strongly typed and reduce type inference.
   - we want to get to the point where this is the process: internal designs -> update spec -> scope frontend + backend work -> ship features.
2. we need to enable programmatic user API calls, and probably figure out a minimal right role-based access control pattern
   - this PR splits the UX into three categories/tables: `admins`, `users`, and `agents`. admins are for internal-only users (superadmin API key), users are session/cookie/UI-only auth (for now), and agents are API-only bots.
   - this stop-gap should make a future migration easier. we'll probably merge `admins` into `users`, and then set up the proper RBAC while also adding user API admin-like support (e.g., cURL to create agent; update user / owned agent profiles).
3. with (1) and (2), we can implement the rest of the [proposed the backend](https://www.notion.so/Comps-db-API-redesign-1fbdfc9427de809c8697ee92e74ac752?pvs=4) and make the APIs more robust. 
   - after (1), we can start to refactor our express responses to be strongly typed. the frontend app will do the same (albeit, it can use the `api-sdk` w/o needing this to happen first).
   - there's also some internal "larger vision" thinking that will be shared soon—and timing-wise, i think that aligns well with the "do things right" steps above. it could drive some schema changes, tbd.

brunocalza:
Awesome work @dtbuchholz ! 

I wonder how this impacts voting. We were designing voting so that a voter could vote on a [team wallet's address](https://github.com/recallnet/js-recall/pull/321/files#diff-ee5e5148f99f501bd8debd2c3ac320f821a4a5f5885b3dbde349ff103d95cc53R10). I guess the `votes_performed` table should reference the `agents` table now, right? Because `agents` are the one competing. Or it could be the case that we may want to cast votes on a user?

cc @ukstv

dtbuchholz:
> Awesome work @dtbuchholz !
> 
> I wonder how this impacts voting. We were designing voting so that a voter could vote on a [team wallet's address](https://github.com/recallnet/js-recall/pull/321/files#diff-ee5e5148f99f501bd8debd2c3ac320f821a4a5f5885b3dbde349ff103d95cc53R10). I guess the `votes_performed` table should reference the `agents` table now, right? Because `agents` are the one competing. Or it could be the case that we may want to cast votes on a user?
> 
> cc @ukstv

@brunocalza correct, reference the `agents` table and their wallet address. voting is a signal on agents/competitors, and not the user/admin of the agent. 

this PR is the first step in splitting a team (where every db entry is effectively an agent) into agents + users (who have different permissions). the gap it introduces for _agent_ wallets: how do agents prove wallet ownership? i put down some additional thoughts in https://github.com/recallnet/js-recall/issues/419, but we should be able to easily address it with an API call that lets agents verify wallet ownership as a prerequisite to compete.

brunocalza:
A follow up to that: what about rewards? I assume the user would be the one claiming the rewards for all of their agents. Makes sense?

dtbuchholz:
> A follow up to that: what about rewards? I assume the user would be the one claiming the rewards for all of their agents. Makes sense?

@brunocalza agreed. the comps app is very user-centric and should rely on that user->agent(s) relationship, allowing users to accumulate/claim rewards for all of the agents they deploy. so i guess the distinction to make is:
- agent wallets exist to uniquely identify AIs with an onchain representation, which then lets users signal an agent's competency by voting on that agent's wallet address.
- user wallets exist to stake / vote / claim, which means that rewards are tied to the user wallet and not the agent(s) it owns.

longer term, i think this does deserve some meta/vision discussion. does there always have to be a human-in-loop, or can agents participate w/o oversight? can agents own tokens, and then stake/vote on other agents? should agents be "allowed" to claim their own rewards? idk what this looks like in the grand scheme of things and what the human vs agent feedback loop should be, outside of the comps app.

dtbuchholz:
note: i added some fast follow up issues in https://github.com/recallnet/js-recall/issues/405; unless there's anything else super glaring, we can use that issue and fire away at cleanups.

apps/api/src/controllers/agent.controller.ts
```diff
@@ -53,57 +75,90 @@ export function makeAccountController(services: ServiceRegistry) {
...
      * @param next Express next function
      */
     async updateProfile(req: Request, res: Response, next: NextFunction) {
       try {
-        const teamId = req.teamId as string;
-        const { contactPerson, metadata, imageUrl } = req.body;
+        const agentId = req.agentId as string;
+        const { name, description, imageUrl } = req.body;
+
+        // Validate that only allowed fields are being updated
```
joewagner, line 89: Not something to change in this PR, but just a call out for future work.  `zod` is designed to do this kind of validation.  Basically any time we are doing `req.field as <type>` we can use zod's parse method.

PR #428, opened 2025-05-27
feat: voting defs and rels

ukstv:
@brunocalza Moved the rewards table here as well.

This should be in a mergeable state.

Also cc @avichalp

### DIFF
--- a/apps/api/src/database/schema/voting/defs.ts
+++ b/apps/api/src/database/schema/voting/defs.ts
@@ -8,6 +8,7 @@ import {
   serial,
   text,
   timestamp,
+  uniqueIndex,
   uuid,
 } from "drizzle-orm/pg-core";
 
@@ -21,29 +22,34 @@ const bytea = customType<{ data: Uint8Array; notNull: false; default: false }>({
 });
 
 export const epochs = pgTable("epochs", {
-  id: serial().primaryKey(),
+  id: uuid().primaryKey().notNull(),
+  number: serial().notNull().unique(),
   startedAt: timestamp("started_at", { withTimezone: true }).notNull(),
   endedAt: timestamp("ended_at", { withTimezone: true }),
 });
 
-export const stakes = pgTable("stakes", {
-  id: uuid().primaryKey().notNull(),
-  userId: uuid("user_id")
-    .notNull()
-    .references(() => users.id, { onDelete: "cascade" }),
-  amount: tokenAmount("amount").notNull(),
-  address: blockchainAddress("address").notNull(),
-  expiresAt: timestamp("expires_at", { withTimezone: true }).notNull(),
-  withdrawalAt: timestamp("withdrawal_at", { withTimezone: true }),
-  withdrawnAt: timestamp("withdrawn_at", { withTimezone: true }),
-  epochCreated: uuid("epoch_created").references(() => epochs.id),
-  createdAt: timestamp("created_at", { withTimezone: true })
-    .defaultNow()
-    .notNull(),
-  updatedAt: timestamp("updated_at", { withTimezone: true })
-    .defaultNow()
-    .notNull(),
-});
+export const stakes = pgTable(
+  "stakes",
+  {
+    id: uuid().primaryKey().notNull(),
+    userId: uuid("user_id")
+      .notNull()
+      .references(() => users.id, { onDelete: "cascade" }),
+    amount: tokenAmount("amount").notNull(),
+    address: blockchainAddress("address").notNull(),
+    expiresAt: timestamp("expires_at", { withTimezone: true }).notNull(),
+    withdrawalAt: timestamp("withdrawal_at", { withTimezone: true }),
+    withdrawnAt: timestamp("withdrawn_at", { withTimezone: true }),
+    epochCreated: uuid("epoch_created").references(() => epochs.id),
+    createdAt: timestamp("created_at", { withTimezone: true })
+      .defaultNow()
+      .notNull(),
+    updatedAt: timestamp("updated_at", { withTimezone: true })
+      .defaultNow()
+      .notNull(),
+  },
+  (table) => [index("idx_stakes_address").on(table.address)],
+);
 
 export const voteAssignments = pgTable(
   "vote_assignments",
@@ -54,7 +60,7 @@ export const voteAssignments = pgTable(
     userId: uuid("user_id")
       .notNull()
       .references(() => users.id, { onDelete: "cascade" }),
-    epoch: uuid("epoch")
+    epoch: uuid()
       .notNull()
       .references(() => epochs.id, { onDelete: "cascade" }),
     createdAt: timestamp("created_at", { withTimezone: true })
@@ -63,9 +69,12 @@ export const voteAssignments = pgTable(
     amount: tokenAmount("amount").notNull(),
   },
   (table) => [
-    index("idx_vote_assignments_user_id").on(table.userId),
+    primaryKey({
+      columns: [table.stakeId, table.userId, table.epoch],
+      name: "vote_assignments_pkey",
+    }),
+    index("idx_vote_assignments_user_epoch").on(table.userId, table.epoch),
     index("idx_vote_assignments_epoch").on(table.epoch),
-    index("idx_vote_assignments_stake_id").on(table.stakeId),
   ],
 );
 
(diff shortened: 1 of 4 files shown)


apps/api/src/database/schema/voting/defs.ts
```diff
@@ -27,96 +28,156 @@ export const epochs = pgTable("epochs", {
...
-  updatedAt: timestamp("updated_at", { withTimezone: true }).defaultNow().notNull(),
-}, (table) => [
-  index("idx_votes_performed_user_id").on(table.userId),
-  index("idx_votes_performed_agent_id").on(table.agentId),
-  index("idx_votes_performed_epoch").on(table.epoch),
-]);
+export const votesPerformed = pgTable(
+  "votes_performed",
+  {
+    userId: uuid("user_id")
```
brunocalza, line 109: Do we need a `user_id` on this table?

ukstv: Vote performed is effectively a triplet: (who, how many votes, destination). "userId" is the `who`.

ukstv: "agentId" is the destination

apps/api/src/database/schema/voting/defs.ts
```diff
@@ -27,96 +28,156 @@ export const epochs = pgTable("epochs", {
...
-  primaryKey({ columns: [table.userId, table.epoch], name: "votes_available_pkey" }),
-  index("idx_votes_available_user_id").on(table.userId),
-]);
+export const votesAvailable = pgTable(
+  "votes_available",
+  {
+    userId: uuid("user_id")
+      .notNull()
+      .references(() => users.id, { onDelete: "cascade" }),
+    epoch: uuid("epoch")
```
brunocalza: I thought votes available would be a global thing and not really tied to a particular epoch. As far as I understand, you can vote on epoch 2 using the votes from epoch 1.

ukstv: That was my understanding as well, until few days ago.

apps/api/src/database/schema/voting/defs.ts
brunocalza: Oh, I see makes sense, so the users table have both competitors and voters?

ukstv: the competitors are in "agents" table now

brunocalza: sorry, I mean by **competitors** the human users that own agents. I'm assuming there are **users** that own zero agents and are just voting. So the **users** table would have both **human competitors** and **human voters**, and I guess a **human competitor** can also be a **voter** is he stakes some amount. 

Sorry for this rambling, the table is good. I'm just trying to make sense of the different user personas

ukstv: As far as I know, the team id is the representation of an agent. @dtbuchholz  correct me if I am wrong here.

ukstv: And user ID is the closest to a "voter".

apps/api/src/database/schema/voting/defs.ts
```diff
@@ -0,0 +1,183 @@
...
+import { blockchainAddress, tokenAmount } from "@/database/schema/util.js";
+
+const bytea = customType<{ data: Uint8Array; notNull: false; default: false }>({
+  dataType() {
+    return "bytea";
+  },
+});
+
+export const epochs = pgTable("epochs", {
+  id: serial().primaryKey(),
```
brunocalza: just noticed this is not an uuid but in other tables epochs are uuids

apps/api/src/database/schema/voting/defs.ts
dtbuchholz: > As far as I know, the team id is the representation of an agent. @dtbuchholz correct me if I am wrong here.

yeah, sort of. we used to have `teams` where every entry was a `teamId` (agent+user=team). instead, we now have an `agents` table with `agentId`, and a `users` with `userId`. agents participate in competitions, and users stake/vote on agents (and/or create them).

so yes, a user ID is the closest thing to a "voter".

> sorry, I mean by competitors the human users that own agents. I'm assuming there are users that own zero agents and are just voting. So the users table would have both human competitors and human voters, and I guess a human competitor can also be a voter is he stakes some amount.

yeah, that sounds right. staking unlocks the ability to let you enter you agents in a competition, and it also gives you the ability to vote. the `users` table doesn't really care which persona you are since that's a calculated "attribute" that gets handled by these voting tables + staking contract indexer.

apps/api/src/database/schema/voting/defs.ts
dtbuchholz: yeah i fixed that here if yall want it https://github.com/recallnet/js-recall/pull/435

dtbuchholz: well, i fixed it so that the other tables are `integer` types so that they properly map to the `serial` type. idk if that's preferred, or if we want the opposite and should make this `serial` type a `uuid`?

brunocalza: maybe keep uuid just to keep every table in the db using uuid, so we have a nice pattern that and we don´t have to think too much about it?

apps/api/src/database/schema/util.ts
```diff
@@ -0,0 +1,15 @@
+import { numeric, varchar } from "drizzle-orm/pg-core";
+
+/**
+ * PG column for amount of tokens in humanese
+ */
+export function tokenAmount(name: string) {
+  return numeric(name, { precision: 30, scale: 18, mode: "number" });
```
brunocalza, line 7: sorry to keep commenting on things, but it's hard to see the nuances of the schema all at once. I feel like the  `mode: "number"` is not enough to represent token amount correctly, and we should use bigint. but I don´t know enough about these conversions from postgres to typescript

apps/api/src/database/schema/voting/defs.ts
```diff
@@ -0,0 +1,183 @@
...
+      .notNull()
+      .references(() => users.id, { onDelete: "cascade" }),
+    epoch: uuid("epoch")
+      .notNull()
+      .references(() => epochs.id, { onDelete: "cascade" }),
+    amount: tokenAmount("amount").notNull(),
+    createdAt: timestamp("created_at", { withTimezone: true })
+      .defaultNow()
+      .notNull(),
+    updatedAt: timestamp("updated_at", { withTimezone: true })
```
avichalp, line 94: Should we have columns such as `block_number`, `tx_hash`, `log_index` to record which "Stake" event made these votes available. These columns will be duplicated here and in the `Events` table from indexing DB.  

I am also wondering at this point if we need an Events table at all? Instead of the storing all events in a db first, should the indexer directly populate the votes tables when the events are fetched from RPC. What do you think?

apps/api/src/database/schema/util.ts
dtbuchholz: @mzkrasner / @joewagner i know you guys dealt with some number precision issues. is there a reason we chose the pattern in the snippet (or see [here](https://github.com/recallnet/js-recall/blob/main/apps/api/src/database/schema/trading/defs.ts#L157) in the `main` branch) vs bigints?

apps/api/src/database/schema/voting/defs.ts
dtbuchholz: > Should we have columns such as block_number, tx_hash, log_index to record which "Stake" event made these votes available. These columns will be duplicated here and in the Events table from indexing DB.

could the `events` table handle this w/o needing to add the columns to this `votes_available` table? e.g., maybe `events` tracks `epoch` for each event. (note: i dont remember the full design here, so i could be off-base.)

> I am also wondering at this point if we need an Events table at all? Instead of the storing all events in a db first, should the indexer directly populate the votes tables when the events are fetched from RPC. What do you think?

i think having an `events` table is useful so that we can have the proper audit trail of events -> votes calculations.

apps/api/src/database/schema/voting/defs.ts
dtbuchholz: agreed, makes sense...i'll update my branch. (i wasn't sure if `serial` was for blockchain related reason.)

apps/api/src/database/schema/voting/defs.ts
avichalp: > i think having an events table is useful so that we can have the proper audit trail of events -> votes calculations.

yes, thats the current idea. 

We thought about having two processes running. One for indexing events and storing in the `events` table. The other one for querying the events to at regular intervals and adding votes in the `votes_available` table (in js-recall db). The second process needs to know the next (block, tx, event) that it should query. Hence we need to add those columns. 

An alternative could be store all events in the `events` and the same time update vote counts in the `vote_available` table in the same process.

PR #435, opened 2025-05-27
fix(api): integer epoch type | gen migrations

dtbuchholz:


apps/api/src/database/schema/voting/defs.ts
```diff
@@ -21,7 +21,7 @@ const bytea = customType<{ data: Uint8Array; notNull: false; default: false }>({
 });
 
 export const epochs = pgTable("epochs", {
-  id: serial().primaryKey(),
+  id: uuid().primaryKey().notNull(),
```
ukstv, line 25: I am a bit conflicted about that. On one hand there is consistency and disambiguation with regards of parallel epochs. On the other hand, the business folks would have to consume the epoch identifiers/or names. What if we add something like whimsical Docker container names here? So an epoch is uniquely identified by this uuid id, and it also has a name like "backstabbing-roentgen".

Issue #373, opened 2025-05-21
Build the rewards merkle tree


brunocalza:
From the rewards calculation output, build the merkle tree

Check out the logic and database schema at https://www.notion.so/Minimum-Token-Utility-Rewards-Spec-1f2dfc9427de8027951ec3645bf02fc5?d=1f8dfc9427de80c59b80001cc7cfd01e#1f2dfc9427de8019b7c6fa1f3e363ded

PR #436, opened 2025-05-27
feat: adds rewards service

brunocalza:
# Summary

Closes https://github.com/recallnet/js-recall/issues/373

This PR implements the logic of building a Merkle Tree out of rewards calculations, for rewarding competitors and voters .

See the [Rewards Spec](https://www.notion.so/Minimum-Token-Utility-Rewards-Spec-1f2dfc9427de8027951ec3645bf02fc5)

# Implementation

- It adds a new service called `RewardsService` that is responsible for calculating rewards, allocating it and proof retrieval. In this PR, **we don't implement rewards calculation**.
- It adds tests for this new service
- It adds a script to be called when we need to allocate the rewards. Usage:
   ```
   pnpm run rewards:allocate d9c141de-796b-4576-a031-dcf065187962
   ```


apps/api/package.json
```diff
@@ -64,9 +64,12 @@
     "crypto": "^1.0.1",
     "dotenv": "^16.0.3",
     "drizzle-orm": "^0.43.1",
+    "ethers": "^6.13.7",
```
dtbuchholz: jfyi—we use [viem](https://viem.sh/) (version ^2.28.1) instead of ethers in all of our packages and apps. there might be some examples [here](https://viem.sh/docs/ethers-migration.html#abi-utilities) on the ethers -> viem methods.

brunocalza: thanks for pointing that out ;)

apps/api/package.json
```diff
@@ -19,6 +19,8 @@
     "test:e2e:runner": "tsx e2e/run-tests.ts",
     "test:multi-chain": "vitest run src/services/providers/__tests__/multi-chain.provider.test.ts",
     "test:multi-chain:e2e": "vitest run e2e/tests/price/multi-chain-provider.test.ts",
+    "rewards:allocate": "tsx scripts/rewards-allocate.ts",
+    "rewards:insert-test-data": "tsx scripts/insert-epoch-rewards.ts",
```
brunocalza, line 23: This script is added to help manual testing

apps/api/src/services/rewards.service.ts
```diff
@@ -0,0 +1,196 @@
...
+              });
+            }
+          }
+        }
+      }
+
+      const rootHash = merkleTree.getRoot();
+
+      await db.transaction(async (tx) => {
+        //TODO: call addDistribution blockchain method
```
brunocalza, line 95: Here we will call the smart contract method [addAllocation](https://docs.google.com/document/d/1tA3DknAQqiCIFkeRjPuZxPBXLYZqqSNJ9q774p3CQbg/edit?tab=t.0#heading=h.l9kfj6etlu8x). To be done in another PR

apps/api/src/database/schema/util.ts
```diff
@@ -1,10 +1,10 @@
 import { numeric, varchar } from "drizzle-orm/pg-core";
 
 /**
- * PG column for amount of tokens in humanese
+ * PG column for amount of tokens in wei (BigInt representation)
  */
 export function tokenAmount(name: string) {
-  return numeric(name, { precision: 30, scale: 18, mode: "number" });
+  return numeric(name, { precision: 78, scale: 0, mode: "bigint" });
```
brunocalza, line 7: Almost sure we need this change. We don't want to do rewards calculations on `number` which is very imprecise

apps/api/src/services/rewards.service.ts
```diff
@@ -0,0 +1,196 @@
...
+import { InsertReward } from "@/database/schema/voting/types.js";
+
+/**
+ * Service for handling reward-related operations
+ */
+export class RewardsService {
+  /**
+   * Calculate rewards for a given input
+   */
+  public async calculateRewards(): Promise<void> {
```
brunocalza, line 21: This method calculates the rewards and store them in the `rewards` table. We did not implement the rewards calculation in this PR. Will do in the next one.

apps/api/src/database/repositories/rewards-repository.ts
```diff
@@ -0,0 +1,164 @@
...
+  }
+}
+
+/**
+ * Get Merkle proof for a specific leaf node in the rewards tree
+ * @param epoch The epoch ID (UUID)
+ * @param leafHash The hash of the leaf node to generate proof for
+ * @returns Array of proof hashes as Uint8Array
+ */
+export async function getMerkleProof(
```
brunocalza: This query is a bit hard to grasp. If we think it's too complicated to maintain, we can move this logic to JS

apps/api/src/database/schema/util.ts
brunocalza: cc @mzkrasner @joewagner @dtbuchholz @ukstv

dtbuchholz: this method is only used by the staking-related tables, so whatever makes sense to you should work. later, we can consider reusing this method in the other tables and replace the `numeric(name, { precision: 30, scale: 18, mode: "number" });` definitions.

apps/api/e2e/tests/rewards.test.ts
```diff
@@ -0,0 +1,314 @@
...
+  epochs,
+  rewards,
+  rewardsRoots,
+  rewardsTree,
+} from "@/database/schema/voting/defs.js";
+import { InsertReward } from "@/database/schema/voting/types.js";
+import { RewardsService, createLeafNode } from "@/services/rewards.service.js";
+
+// Custom cleanup function to avoid dependency on test-helpers
+async function cleanupTestState(): Promise<void> {
```
dtbuchholz: just wondering—why do you need to avoid `test-helpers.ts`?

the `cleanupTestState` from `test-helpers.ts` calls a special `resetDb` method, which catches a flaky "deadlock" postgres error. i just wanna be cognizant of that and avoid potentially reintroducing the dealock error in CI.

apps/api/src/database/repositories/rewards-repository.ts
dtbuchholz: if it's straightforward enough (e.g., translate with LLM), having this in JS would be nice so we can avoid casting the result and use drizzle types. it's not the end of the world tho.

apps/api/e2e/tests/rewards.test.ts
brunocalza: No special reason TBH. I didn't know it had this special logic. Thanks for calling that out. I'll make the change

apps/api/src/database/repositories/rewards-repository.ts
brunocalza: sounds good!

ukstv (approved):

brunocalza:
@dtbuchholz I've made the changes
- moved the proof retrieval to JS
- use `cleanupTestState` from test-helpers

It's a bit hard to review the last commit because I formatted the code, but lmk if it's good to merge

Issue #424, opened 2025-05-27
Update table schemas for v1 launch


dtbuchholz:
A few things we need to address:

- [x] ~~Our drizzle schema uses `withTimezone: true`. This should, instead, be false. I _think_ means we'll have to change insertions from `new Date()` to `new Date().toISOString()`.~~
- [ ] There are a few missing fields needed for #314 in these tables:
   - [ ] `agents`: `skills`, `trophies`, and `stats`. Perhaps we can just get away with storing this as agent metadata?
   - [x] `competitions`: `type` (this will all be trading related...we should create an enum to encapsulate this...maybe `finance` is the overarching umbrella? or `trading`.
- [x] The enums, like `status`, need to be migrated to a `pgEnum`. Also, review this field for consistency since a new field was added to better align with an agent being deactivated as part of the post-competition logic. The fields should be: active, inactive, suspended, deleted.
- [x] Wallet addresses need to be inserted as lowercased values, and making queries against them should do the same. This is ESSENTIAL for uniqueness checks. 
- [x] The `agents` table should have a `UNIQUE` constraint on the agent's `walletAddress`. And should we constrain the agent's table to enforce non-overlapping `walletAddress` values with the `users` table? Probably: yes. Note that we don't actually use the agent's wallet field, yet, because we need #419.
- [x] The `agents` table needs the email address as a dedicated field, not metadata.

Note: we can treat the tables as brand new and delete migrations.

dtbuchholz:
note: the internet (see [here](https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_timestamp_.28without_time_zone.29), [here](https://community.spiceworks.com/t/zone-of-misunderstanding/928839), etc.)and LLMs overwhelming agree to use `withTimezone: true` (aka `timestampz`) since it stores an absolute instant / UTC without ambiguity and avoids the silent-shift bugs that plague `timestamp`.

Issue #426, opened 2025-05-27
Use `zod` for Express request parsing


dtbuchholz:
**Problem Statement**

Use `zod` for Express request parsing.

**Proposed Solution**
Below is how we mostly handle parsing Express request params/body. `zod` is designed to do this kind of validation. Basically, any time we are doing `req.field as <type>`, we can use zod's parse method. (via @joewagner)

The code below is the existing state, and we can use `zod` to improve this handling.
```js
 async updateProfile(req: Request, res: Response, next: NextFunction) {
   try {
    const teamId = req.teamId as string;
    const { contactPerson, metadata, imageUrl } = req.body;
    const agentId = req.agentId as string;
    const { name, description, imageUrl } = req.body;

    // Validate that only allowed fields are being updated
    // 
   } 
}
```

**Impact**
Both frontend and backend comps work. It'll make our APIs more robust.

**Priority**
Medium. Not mission critical. However, in #314, we're already refactoring the APIs, so maybe this can be slowly built into those changse.

joewagner:
> [@joewagner](https://github.com/joewagner) re: the comment above—there are some places where we do this:
> 
> const { success, data, error } = UpdateUserProfileSchema.safeParse({
>   userId: req.userId,
>   body: req.body,
> });
> so building off of what you mentioned, does this demonstrate what we want to do? i.e., just pass the full `req` instead of destructuring it in the method:
> 
> const { success, data, error } = UpdateUserProfileSchema.safeParse(req);
> or, would we want dedicated methods to handle it, like your original example? or maybe it's both:
> 
> function checkUpdateUserProfileRequest(request) {
>   const { success, data, error } = UpdateUserProfileSchema.safeParse(request);
>   // Handle 400 error
>   // Then, destructure things for cleaner usage
>   const { ... } = data
>   return { ... }
> }
> 
> const { ... } = checkUpdateUserProfileRequest(req);

Good question.  
The idea is that if we have half a dozen routes that require and check for userId or agentId, it seems better to have a function(s)/method(s) that ensure the required thing is present.  This could be setup as middleware, or return the required thing depending on the pattern we like best.  It also works for stuff that should be in the querystring and so on

An example of what I mean:
```
/**
     * Get all agents owned by the authenticated user
     * @param req Express request with userId from session
     * @param res Express response
     * @param next Express next function
     */
    async getAgents(req: Request, res: Response, next: NextFunction) {
      try {
        const userId = ensureUserId(req);
[...]

Issue #438, opened 2025-05-28
Import / migrate old competition data


rohhan:
**Problem Statement**
We are launching a new application, but there is existing competition data we should show in there.

**Proposed Solution**
Import data from the micro-site databases into the new competition app DB.  Do our best to match the schemas.

- Get a CSV of existing competition data from @KGmajor / @mzkrasner 
- Write a script to import that data into our database

**Impact**
Users to the new site will have increased confidence by being able to see the historical data.

**Priority**
Medium

stbrody:
Not a blocker for launch, but good to get soon

mzkrasner:
@dtbuchholz @stbrody a few q I think we'd need to define first:

1. i'm assuming the import process should look at the final shape of the database we deliver for mvp - have we defined what the final schema will be? It looks like there are still some tickets in flight that have not yet been merged that change db schema, which would impact how we manage this import and backfill process

2. Does this ticket include teardown and redeployment of the "production" server? Reason here is because we'd need somewhere to import into, and right now the production server yet again uses an outdated version of the database design. In reading @rohhan 's description, I'm unsure if this should also encompass the backfilling process, or just the prepping process

dtbuchholz:
1. hm good callout. there are tables that don't exist, yet. they shouldn't _directly_ conflict with the data you'll be writing, but i'd imagine all of should be done by this week. the outstanding tables are:
   - #557 (just merged)
   - #528 (soon to be merged; not sure if there's anything else for staking #305)
   - #561
2. agreed w/ spencer. it should incorporate preparing + writing the data to the db. 

note: i think this'll just be a one-time script. once we have a production db loaded with this data, we can just leverage Drizzle migrations to handle db upgrades or schema changes.

mzkrasner:
@dtbuchholz @stbrody as per the conversation in slack, I'm proposing that DoD is an actual pg_dump of the normalized data from the past two competitions that uses the database design at the following commit: [c964956d2f0996edd4505b0532587604362cd765](https://github.com/recallnet/js-recall/commit/c964956d2f0996edd4505b0532587604362cd765)

dtbuchholz:
@mzkrasner locking to that commit and the `pg_dump` approach both make sense.

as for the "script" to eventually upload the data to our prod db—we should be able to just run something like this, right?: 
1. `pg_dump --data-only backfilled_comps > normalized.sql` (aka what you're doing here)
[...]

Issue #437, opened 2025-05-27
Add email verification logic


dtbuchholz:
**Problem Statement**
We let users and agents set an email address mapped to their profiles. This is required in the short term for capturing marketing related data (e.g., distribution lists), but it's not a secure practice. 

Note: potentially related to https://github.com/recallnet/js-recall/issues/336
 
**Proposed Solution**
Add an email verification step. If the user inputs their email address in their profile or agent page, we need to send a variation email. Once the email is verified, we update the db.

Note: what about other socials? What's the best practice there, e.g., verify your X or TG account?

**Impact**
Mostly backend. Also marketing, to ensure we have legit emails to reach out to. 

**Priority**
Low. Something we need to do, but isn't essential for anything right now.

stbrody:
Emails are important for user attribution.  We need a way to activate the users who interact with the app, especially before TGE.

We do collect emails, but we don't verify them currently.

dtbuchholz:
> Do we have an email provider that I can use for this (e.g. SendGrid, Mailgun)? who can help setting up this?

we _do_ have a Loops account, which [marketing uses](https://recalllabs.slack.com/archives/C08T8C9S119/p1750285541362279) to send emails out. you might be able to do something programmatic and send a [transactional email](https://loops.so/docs/transactional/guide). 

if that doesnt work, then whatever's cheapest / easiest. any of those that you mentioned, [amazon ses](https://aws.amazon.com/ses/), etc

> Where should I mark the user or agent as verified? New column (is_verified) on users table?

maybe a new column `is_email_verified` in the `users` and `agents` table?

note: our agent-related API response includes a computed `isVerified` field. it gets populated if the agent has a `walletAddress` field, and this only happens if the agent executes a signature verification API call. i dont think that matters here because the email-related verification isnt important to the app or APIs, and only matters to marketing. (we can deal with it later if this changes in the future.)

dtbuchholz:
not to complicate things, but Privy also offers [email verification](https://docs.privy.io/authentication/user-authentication/login-methods/email). if the amount of work to add Privy was the same as the other options, i think that is also another viable option. it adds more work to the scope of this issue, though, so i'd imagine it's something we can continue to keep as a future-state idea.

PR #441, opened 2025-05-28
feat: database and openapi cleanups

dtbuchholz:
## Summary

Database and OpenAPI cleanups to meet frontend requirements. Closes https://github.com/recallnet/js-recall/issues/424, https://github.com/recallnet/js-recall/issues/424

## Details

- Starts to move some enums to a zod-friendly setup (related: #426)
- Use `pgEnum` in table schemas (related to the new types in the point above).
- Small table tweaks like adding a `status` field on the `competitions` table, adding `email` to the agents table, or enabling some new fields PUT / POST in agent or user endpoints (for UI reqs)
- OpenAPI additions to reflect the point above and new fields that are accessible.


dtbuchholz:
the only outstanding issue from #424: how should we represent agents skills, stats, and trophies? 

namely, i'm not sure if these are defined anywhere in particular.
- we could store these as arbitrary agent metadata field, and then build the schema afterward? that's how we do in the mini app
- we could add them as dedicated columns to make it easier to look up/mutate. maybe skills are an enum, and trophies are jsob
- both of these could very well be their own set of tables that store FKs to the agent's `agentId` table.

apps/api/src/types/index.ts
```diff
@@ -362,18 +385,85 @@ export interface LoginResponse {
   wallet?: string;
 }
 
-// Zod definitions allow user input validation at runtime and define types
+/**
+ * Actor (user, agent, admin) status values for zod or database enum.
+ */
+export const ACTOR_STATUS_VALUES = [
+  "active",
+  "inactive",
```
dtbuchholz, line 393: first, i started to move away from some TS enum types to this pattern. since these types are used by zod, we want to avoid enums per https://zod.dev/api?id=enums. you'll see this for the actor (admin, user, or agent) status, competition status, and cross-chain trading type.

this pattern is a little duplicative because `ACTOR_STATUS_VALUES` and `ACTOR_STATUS` store the same values. the former is useful for zod type inference or PG enums, and the latter can be used _like_ a TS enum. i suppose a TS enum could still be used for `ACTOR_STATUS`...but Claude/Cursor convinced me to use this pattern.

apps/api/e2e/tests/competition.test.ts
```diff
@@ -312,7 +312,7 @@ describe("Competition API", () => {
     expect(adminStatusResponse.competition?.name).toBe(competitionName);
     expect(adminStatusResponse.competition?.status).toBe("active");
     expect(adminStatusResponse.competition?.description).toBeDefined();
-    expect(adminStatusResponse.competition?.externalLink).toBeDefined();
+    expect(adminStatusResponse.competition?.externalUrl).toBeDefined();
```
dtbuchholz, line 320: note: we change the competition `externalLink` to `externalUrl` for consistency purposes (e.g., we use `imageUrl` elsewhere, so `...Url` is ideal)

apps/api/e2e/tests/competition.test.ts
```diff
@@ -481,17 +481,17 @@ describe("Competition API", () => {
     const createResponse1 = (await adminClient.createCompetition(
       comp1Name,
       "Test competition 1",
-      CrossChainTradingType.allow,
+      CROSS_CHAIN_TRADING_TYPE.ALLOW,
```
dtbuchholz, line 489: here's how the pattern changes w/o the TS enum type, and with `as const` type instead. (fwiw, i personally dont have a strong preference b/w the two)

apps/api/src/controllers/agent.controller.ts
```diff
@@ -84,10 +84,16 @@ export function makeAgentController(services: ServiceRegistry) {
     async updateProfile(req: Request, res: Response, next: NextFunction) {
       try {
         const agentId = req.agentId as string;
-        const { name, description, imageUrl } = req.body;
+        const { name, description, imageUrl, email, metadata } = req.body;
```
dtbuchholz: the agent + update profile endpoint can now support email and generic metadata

apps/api/src/controllers/user.controller.ts
```diff
@@ -56,10 +56,10 @@ export function makeUserController(services: ServiceRegistry) {
     async updateProfile(req: Request, res: Response, next: NextFunction) {
       try {
         const userId = req.userId as string;
-        const { name, imageUrl } = req.body;
+        const { name, imageUrl, email } = req.body;
```
dtbuchholz, line 59: the user + update profile endpoint can now support email. (metadata isnt support for users, yet, because we dont have any functionality that really uses it yet.)

apps/api/src/controllers/user.controller.ts
```diff
@@ -281,14 +294,20 @@ export function makeUserController(services: ServiceRegistry) {
       try {
         const userId = req.userId as string;
         const { agentId } = req.params;
-        const { name, description, imageUrl } = req.body;
+        const { name, description, imageUrl, email, metadata } = req.body;
```
dtbuchholz, line 322: users can update their agent + update profile endpoint with email and generic metadata (e.g., from SIWE session)

apps/api/src/database/repositories/agent-repository.ts
```diff
@@ -19,14 +19,14 @@ import { PartialExcept } from "./types.js";
...
     const now = new Date();
-    const [result] = await db
-      .insert(agents)
-      .values({
-        ...agent,
-        createdAt: agent.createdAt || now,
-        updatedAt: agent.updatedAt || now,
-      })
-      .returning();
+    const normalizedWalletAddress = agent.walletAddress?.toLowerCase();
```
dtbuchholz, line 50: we need to ensure that all inserts or queries that touch `walletAddress` are properly lowercased. else, we'll run into problems with UNIQUE constraints on inserts, or a "no match" on queries when the wallet _does_ exists (e.g., trying to check: checksummed address === all-lowercase address).

the alternative is storing the data as `bytea`, but i think this is fine for now

apps/api/src/database/repositories/agent-repository.ts
```diff
@@ -78,6 +78,26 @@ export async function findByOwnerId(ownerId: string): Promise<SelectAgent[]> {
   }
 }
 
+/**
+ * Find agents by wallet address
+ * @param walletAddress Wallet address to search for
+ */
+export async function findByWalletAddress(
```
dtbuchholz: (not used, but i figure it'll be useful for staking/voting related stuff since those tables care about wallets > uuids)

apps/api/src/database/schema/core/defs.ts
```diff
@@ -11,6 +12,33 @@ import {
...
+import {
+  ACTOR_STATUS_VALUES,
+  COMPETITION_STATUS_VALUES,
+  COMPETITION_TYPE_VALUES,
+} from "@/types/index.js";
+
+/**
+ * Statuses for users, agents, and admins.
+ */
+export const actorStatus = pgEnum("actor_status", ACTOR_STATUS_VALUES);
```
dtbuchholz, line 24: here's the PG enums that use those new consts, which also get used by zod

apps/api/src/controllers/agent.controller.ts
mzkrasner: Makes sense. Posted in Discord, but it looks like the unique email constraint for users was removed when we moved over to drizzle. is this fine that multiple users can have the same email address? Do we also want to prevent users and agents from having the same email?

dtbuchholz: for posterity from our dms—unique has been added [here](https://github.com/recallnet/js-recall/pull/441/files/6f22df33aac2d5a83400c9877c9b4c008a2b1fc4#diff-63a1cbabf1a6ace5e3b5b93d6038347d81c34a9539f69417bbe844edcaf2ba5eR51) / [here](https://github.com/recallnet/js-recall/pull/441/files/6f22df33aac2d5a83400c9877c9b4c008a2b1fc4#diff-63a1cbabf1a6ace5e3b5b93d6038347d81c34a9539f69417bbe844edcaf2ba5eR83)...albeit, they're not unique across users<>agents. partially related to #437, or we can just enforce it sooner than later.

somewhat related, wallet addrs are not unique across users<>agents, but #419 can handle this part. this is scoped work that we can probably land next week.

mzkrasner:
> @mzkrasner lmk if this all makes sense
> 
> > My only question is whether (or at what point) moving forward should we care about the integrity of migration safety.
> 
> i've been planning on this launch to _not_ have a db migration from legacy teams-centric design, and we can see how easy backfilling will be via #438 (a manual effort; i'm not sure how clean this will be).
> 
> but moving forward, yes, we'll definitely leverage db migrations.
> 
> > The direct varchar-to-enum conversions using USING clauses could fail if we have any unexpected status values in production.
> 
> true. i _think_ this PR covers all of the existing varchars that need to be enums (like statuses, or x-chain trading types). thus, if we migrate in the future, the new logic should prevent exactly what you noted. albeit, i guess you might talking about the case where we remove an enum value. anyways, lmk if you see anything missing tho, and i can sneak that into this PR. (this obviously presumes the point above is true wrt ignoring migrations before launch.)
> 
> > For wallet address normalization, should we be concerned about the existing data (if migrating existing) and mixed case.
> 
> since i think we can just "backfill" (i.e., not a true migration) the legacy db data, we can massage it beforehand and convert to lowercase before writing to the db. i'm open to ideas, though...the point is that we just need the db queries to agree on how they insert/read data so that wallet uniqueness checks work. e.g., we could use viem + checksum conversions instead of lowercasing, or use bytea and convert from/to strings.
> 
> > Also - should we be doing email format validation in:
> 
> sure, i can toss that in there, along with the other fields in the user/agents apis. stop gap until #426 happens, which will defer body/param parsing & validation to zod and make it more robust via `z.email(...)` typechecks.

Makes sense and thank you for reminding me of the plan from now through launch! I'll go ahead and approve for now and assume you'll add in the minor form validation you mentioned above

mzkrasner (approved):
Notes already covered in sep comments

apps/api/e2e/tests/agent.test.ts
```diff
@@ -791,4 +791,80 @@ describe("Agent API", () => {
     expect(foundAgent).toBeDefined();
     expect(foundAgent?.imageUrl).toBe(newImageUrl);
   });
+
+  test("agents cannot update email or metadata", async () => {
```
dtbuchholz, line 795: note: i ended up reverting the changes i made that allow agents to update their email or metadata. i kept the user->agent profile updates.

in general, agents should not be able to update metadata or emails; they're unpredictable. by default, _users_ should own that permission. RBAC related stuff. (this was probs obvious but it took me a second to realize this, hah.)

anyways, we now allow the user to update agent metadata, but only through the UI/session, which means we're still in control of what data is written.

PR #447, opened 2025-05-28
feat(api): getting agents with sort filter page

joewagner:
This implements the GET `/agents` endpoint per the Spec https://docs.google.com/document/d/16Zu5-Rdk6DAimdzmK1sLCkFrgDHRCOmPgoWHlgqoaTU/edit?pli=1&tab=t.0

This also includes the ability to create a wallet address for an agent in the tests, since we need to test that the endpoint can be filtered by wallet.


apps/api/src/database/repositories/agent-repository.ts
```diff
@@ -42,9 +61,23 @@ export async function create(agent: InsertAgent): Promise<SelectAgent> {
...
  * Find all agents
  */
-export async function findAll(): Promise<SelectAgent[]> {
+export async function findAll(
+  pagingParams?: PagingParams,
+): Promise<SelectAgent[]> {
   try {
-    return await db.select().from(agents);
+    if (!pagingParams) {
+      return db.select().from(agents);
```
joewagner, line 78: leaving this here to keep backward compatibility, but going forward we should avoid database queries that might load an entire table into memory.

apps/api/e2e/tests/agent.test.ts
```diff
@@ -264,6 +265,90 @@ describe("Agent API", () => {
...
+    expect(adminLoginSuccess).toBe(true);
+
+    // Register multiple users and agents
+    const agentData = [
+      {
+        userName: `User A ${Date.now()}`,
+        userEmail: `usera${Date.now()}@example.com`,
+        agentName: `Agent A ${Date.now()}`,
+        walletAddress: generateRandomEthAddress(),
+        agentWalletAddress: generateRandomEthAddress(),
```
joewagner, line 281: I had to add to the `registerUserAndAgentAndGetClient` method, routes, and so on, so that we can test getting by agent wallet address.

apps/api/openapi/openapi.json
```diff
@@ -1023,6 +1023,11 @@
                         "twitter": "hey_kso"
                       }
                     }
+                  },
+                  "agentWalletAddress": {
```
joewagner, line 1027: As noted above, this is so we can test filtering by agent wallet address when getting agents.

apps/api/src/database/repositories/agent-repository.ts
```diff
@@ -98,6 +131,67 @@ export async function findByApiKey(
...
+ */
+export async function findByName({
+  name,
+  pagingParams,
+}: {
+  name: string;
+  pagingParams: PagingParams;
+}): Promise<SelectAgent[]> {
+  try {
+    console.log();
```
dtbuchholz: note: empty/unused `console.log()`

apps/api/src/database/repositories/agent-repository.ts
```diff
@@ -98,6 +131,67 @@ export async function findByApiKey(
...
+
+    return await query;
+  } catch (error) {
+    console.error("[AgentRepository] Error in findByWallet:", error);
+    throw error;
+  }
+}
+
+/**
+ *
```
dtbuchholz: (and might as well add the tsdoc here)

joewagner: Good call.  The tsdoc got lost in the rebase.

joewagner (commented):
I just rebased with some manual conflicts, so I need to double check everything is still good

apps/api/src/services/agent-manager.service.ts
```diff
@@ -753,4 +758,38 @@ export class AgentManager {
...
+    pagingParams: PagingParams;
+  }) {
+    if (filter?.length === 42) {
+      return findByWallet({ walletAddress: filter, pagingParams });
+    }
+    if (typeof filter === "string" && filter.length > 0) {
+      return findByName({ name: filter, pagingParams });
+    }
+
+    return findAll(pagingParams);
```
dtbuchholz, line 779: do you know if `findAll` automatically filters out the agent `apiKey` field? i'm guessing it _doesn't_...so the API calls that use this have to filter out that value like [here](https://github.com/recallnet/js-recall/blob/33d3e5478e7202d61a4e466b08ce9770d3c24d53/apps/api/src/controllers/user.controller.ts#L202).

if this is true, it's obv not ideal for internal DX...we should address in a separate issue. most likely, we'll have a [separate table for API keys](https://www.notion.so/Comps-db-API-redesign-1fbdfc9427de809c8697ee92e74ac752?pvs=4#1fbdfc9427de8069b3d1c4760f4cd53c)

joewagner: Very good catch.  And yes to the idea of a separate table for api keys.

dtbuchholz: cool, created this as a follow up https://github.com/recallnet/js-recall/issues/450

apps/api/src/database/repositories/agent-repository.ts
```diff
@@ -1,17 +1,48 @@
...
+ * allowable order by database columns
+ */
+const agentOrderByFields: Record<string, AnyColumn> = {
+  id: agents.id,
+  ownerId: agents.ownerId,
+  walletAddress: agents.walletAddress,
+  name: agents.name,
+  description: agents.description,
+  imageUrl: agents.imageUrl,
+  apiKey: agents.apiKey,
```
dtbuchholz: ah just saw this field—related to the point below, `apiKey` should be removed. idk if `deactivationReason` and `deactivationDate` should be exposed to, either, but less of an issue.

apps/api/src/services/agent-manager.service.ts
```diff
@@ -753,4 +758,54 @@ export class AgentManager {
...
+      return countByWallet(filter);
+    }
+    if (filter?.length) {
+      return countByName(filter);
+    }
+
+    return count();
+  }
+
+  sanitizeAgent(agent: SelectAgent) {
```
joewagner, line 796: @dtbuchholz I added this method to handle all of the places we are sanitizing agents.  There was a few and they might as well all return the same shaped Object.

apps/api/src/controllers/user.controller.ts
```diff
@@ -167,18 +167,7 @@ export function makeUserController(services: ServiceRegistry) {
...
         res.status(201).json({
           success: true,
-          agent: {
-            id: agent.id,
-            ownerId: agent.ownerId,
-            name: agent.name,
-            description: agent.description,
-            imageUrl: agent.imageUrl,
-            metadata: agent.metadata,
-            apiKey: agent.apiKey, // Include API key for user to use
```
dtbuchholz: ah, this is the one place where we _do_ want to expose the `apiKey`; else, when a user creates an agent, they wont be able to get the agent's API key in the response.

since `sanitizeAgent` sanitizes some of the other fields, maybe it can take an optional to `includeApiKey: boolean = false` or something? or just add a separate method.

dtbuchholz (approved):
left one final comment about the `createAgent` method in the user controller. but once that's addressed, this looks g2g...will approve now so you a subsequent review doesnt hold you up

apps/api/src/database/repositories/agent-repository.ts
joewagner: This is also a good callout, I removed all three of these.

apps/api/src/services/agent-manager.service.ts
dtbuchholz: yeah this is perfect...it'll make it a lot easier to handle the API key issue until we create the proper tables

apps/api/src/controllers/user.controller.ts
joewagner: Oh yeah, duh.

dtbuchholz: partially my fault...the user tests aren't comprehensive enough, so this got missed in logic checks.

joewagner:
@mzkrasner I see the user test failing, just now trying to figure out what's going on. 
I don't see the agent test failing though..

Issue #358, opened 2025-05-19
Implement mutating `/competitions` backend endpoints


dtbuchholz:
- [x] POST /competitions/{competitionId}/agents/{agentId}
- [x] DELETE /competitions/{competitionId}/agents/{agentId}

API details:

<details><summary>POST /competitions/{competitionId}/agents/{agentId}</summary>
<p>
Join a competition

Body: empty body

Response:

- 200 OK
- 401 Unauthorized
- 403 Forbidden: possible causes could be an ongoing/ended competition, an agentId owned by a different user, insufficient stake, etc.
- 404 Not Found: no competition/agent found

</p>
</details> 

<details><summary>DELETE /competitions/{competitionId}/agents/{agentId}</summary>
<p>
Leave a competition

Response:

- 200 OK
- 401 Unauthorized
- 403 Forbidden: possible causes could be an ongoing/ended competition, an agentId owned by a different user, etc.
- 404 Not Found: no competition/agent found

</p>
</details>

mzkrasner:
In review: https://github.com/recallnet/js-recall/pull/455

Issue #451, opened 2025-05-28
Post-launch cleanups (PLACEHOLDER)


dtbuchholz:
This is a meta issue to simply track things we find that should be address after the MVP wraps up the week of May 30. Regardless of how big or small, we can just drop comment in this issues.

Then, @dtbuchholz can sift through for duplicates in the existing backlog, or create new ones, and then get everything prioritized for final fixes.

dtbuchholz:
- sorting logic is duplicative in certain APIs or db queries (noted via https://github.com/recallnet/js-recall/pull/446)

PR #455, opened 2025-05-29
feat: integrate post competition routes

mzkrasner:
### 🎯 **Competition Join/Leave API Endpoints Implementation**

This PR implements two new competition management endpoints as specified in [GitHub Issue #358](https://github.com/recallnet/js-recall/issues/358), enabling agents to join and leave competitions programmatically.

#### **API Endpoints**
- **POST** `/api/competitions/{competitionId}/agents/{agentId}` - Join a competition
- **DELETE** `/api/competitions/{competitionId}/agents/{agentId}` - Leave a competition

#### **Authentication Support**
- **SIWE User Authentication**: Users can manage their agents' competition participation
- **Note that joining a pending competition using the agent API key method will currently fail as the current intended use for these routes is user authenticated only**

#### **Business Logic Implementation**
- **Join Validation**: Only allowed during `pending` competitions, prevents duplicate registrations
- **Leave Behavior**: 
  - `pending` competitions: Remove from roster
  - `active` competitions: Deactivate agent 
  - `ended` competitions: Return 403 forbidden
- **Ownership Validation**: Ensures agents belong to requesting user
- **Status Validation**: Prevents deleted/suspended agents from joining

#### **Database Layer**
- Added `removeAgentFromCompetition()` function to competition repository
- Implemented atomic database operations for competition roster management

#### **Service Layer** 
- Added `joinCompetition()` and `leaveCompetition()` methods to competition manager
- Comprehensive validation logic for all business rules
- Proper error handling with specific error messages

#### **Controller Layer**
- Dual authentication path handling (user session vs agent API key)
- Parameter validation using Zod schemas
- Consistent error mapping and response formatting

#### **API Layer**
- Complete OpenAPI documentation with detailed response schemas
- RESTful route design following existing patterns
- Comprehensive error response documentation

#### **Comprehensive Test Suite**
- **45/45 tests passing** in competition test suite
- Full coverage of both authentication methods
- Error case validation for all 403/404/401 scenarios
- Integration tests for complete user workflows

#### **Test Coverage**
- Successful join/leave operations
- Competition status validation (pending/active/ended)
- Agent ownership verification
- Agent status validation (active/inactive/suspended/deleted)
- Authentication edge cases
- Concurrent operation handling

#### **Admin Controller**
- Enhanced `startCompetition` to support empty `agentIds` array
- Automatically includes pre-registered agents from database
- Maintains backward compatibility

#### **Authentication Middleware**
- Fixed agent authentication error propagation
- Preserves specific inactive agent errors instead of generic "Invalid API key"

#### **User Controller**
- Added `deactivationReason` and `deactivationDate` fields to agent responses
- Updated OpenAPI documentation for enhanced agent status information

#### **SDK Integration**
- Resolved type synchronization issues with `agentWalletAddress` field
- Added new response types for join/leave operations
- Enhanced API client with competition management methods


apps/api/src/controllers/competition.controller.ts
```diff
@@ -654,6 +655,197 @@ export function makeCompetitionController(services: ServiceRegistry) {
...
       }
     },
+
+    /**
+     * Join an agent to a competition
+     * @param req AuthenticatedRequest with competitionId and agentId params
+     * @param res Express response object
+     * @param next Express next function
+     */
+    async joinCompetition(
```
dtbuchholz, line 665: nre: made a note https://github.com/recallnet/js-recall/issues/451#issuecomment-2917733714 to consider an agent-only route in the future (not needed here) that removes the `.../agents/:agentId` path & param.

apps/api/src/controllers/user.controller.ts
```diff
@@ -244,6 +244,8 @@ export function makeUserController(services: ServiceRegistry) {
           email: agent.email,
           metadata: agent.metadata,
           status: agent.status,
+          deactivationReason: agent.deactivationReason,
```
dtbuchholz, line 247: what's the reason we need these? i feel like they're admin-related, and should't be exposed.

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -181,6 +181,49 @@ export async function addAgentToCompetition(
...
+ * @param agentId Agent ID to remove
+ * @returns Boolean indicating if a row was deleted
+ */
+export async function removeAgentFromCompetition(
+  competitionId: string,
+  agentId: string,
+): Promise<boolean> {
+  try {
+    const result = await db
+      .delete(competitionAgents)
```
dtbuchholz, line 196: this looks good, and doesnt need any changes in this PR.

but in the future, i wonder if a _hard_ delete in [`competition_agents`](https://github.com/recallnet/js-recall/blob/8deb5fc482accc10776a38fef6fce8addaa82678/apps/api/src/database/schema/core/defs.ts#L183) is the right approach—vs adding a new column that tracks the agent's status within the competition. e.g., we probably need the posterity tracking to ensure agents dont just leave halfway through if theyre performing poorly.

there are implications elsewhere because our [`agents`](https://github.com/recallnet/js-recall/blob/8deb5fc482accc10776a38fef6fce8addaa82678/apps/api/src/database/schema/core/defs.ts#L88) table uses the `status` field to indicate if an agent is active in a competition. e.g., once a competition ends, the agent is [effectively deactivated](https://github.com/recallnet/js-recall/blob/8deb5fc482accc10776a38fef6fce8addaa82678/apps/api/src/database/repositories/agent-repository.ts#L380) globally. this logic should probably exist in the `competition_agents`, whereas the `agents` state is more of a global view, irrespective of an active competition.

apps/api/src/controllers/user.controller.ts
mzkrasner: Well since it's only used by the authenticated user (`userController.getAgents`), figured they'd want to see this. Also - since users can `deactivate` their agents partway through an active competition and therefore automatically get assigned this message:

```
await this.agentManager.deactivateAgent(
        agentId,
        `Left competition ${competition.name} (${competitionId})`,
      );
      
```

figured they might want to be able to look back and remember that they might've made the decision to deactive their agents

dtbuchholz (approved):
aside from the `deactivationReason`/ date, this lgtm.

if there's a good reason to include those fields, then we can merge. else, just remove em.

apps/api/src/controllers/user.controller.ts
dtbuchholz: ah kk. yeah i suppose it has a bit of a dual-purpose, so it's fine

apps/api/src/database/repositories/competition-repository.ts
mzkrasner: Sounds good and agreed. Resolving for now but will be picked up in a future pr

apps/api/src/controllers/competition.controller.ts
mzkrasner: Good idea

PR #460, opened 2025-05-30
Rafael/api front/auth user

Rcontre360:


apps/comps/types/profile.ts
```diff
@@ -1,14 +1,20 @@
...
+    walletAddress: string;
+    id: string;
+
+    isVerified?: boolean;
+    agents?: AgentResponse[];
+    name?: string;
+    email?: string;
+    imageUrl?: string;
+    website?: string;
+  };
```
ramirotw, line 17: what about these fields? some of them can be moved to the `metadata` field which is basically a json, but not all. Let's discuss it on Slack

ramirotw (dismissed):

apps/comps/types/profile.ts
dtbuchholz: the openapi spec should describe what we need [here](https://github.com/recallnet/js-recall/blob/d1eafb45d5a1c0d1c204d48b1f0604d0a7935bf5/apps/api/openapi/openapi.json#L4749):
- `isVerified` -> remove, because user "verification" is a prerequisite via SIWE...but i dont think it's relevant to users imo. if needed for component compatability purposes, we can keep it for now (and it'll always just be undefined)
- `agents` -> this isn't returned yet, but i'm adding this right now
- `website` -> not in the response, but it can/should be part of the metadata


note: agent-related info is at the endpoint [`/user/agents/:id`](https://github.com/recallnet/js-recall/blob/d1eafb45d5a1c0d1c204d48b1f0604d0a7935bf5/apps/api/openapi/openapi.json#L5157) and can infer `isVerified` based on `walletAddress` existing

dtbuchholz (approved):
note: I'm fine if we merge this now so I can help get the types better aligned

apps/comps/types/profile.ts
ramirotw: @dtbuchholz fyi, `PUT /api/user/profile` needs to allow passing a `metadata` field to include the website. I excluded it from the payload now so we don't have a 400

dtbuchholz: Yeah good callout, on my to-do list 👍

PR #462, opened 2025-05-30
Joe/agents api

joewagner:


apps/api/e2e/tests/agent.test.ts
```diff
@@ -867,4 +867,109 @@ describe("Agent API", () => {
...
+    } = await register();
+    const {
+      res: { apiKey: apiKey2 },
+    } = await register();
+
+    expect(agent).toBeDefined();
+    expect(agent.id).toBeDefined();
+
+    // Make a GET request to fetch the agent details using the agent ID
+    const response = await axios.get(`${getBaseUrl()}/api/agent/${agent.id}`, {
```
joewagner, line 912: Specifically calling out here that the new endpoint is available for an agent to see details about other agents, i.e. not just themselves.

apps/api/e2e/tests/agent.test.ts
```diff
@@ -867,4 +867,109 @@ describe("Agent API", () => {
...
+    expect(agentData.agent.id).toBe(agent.id);
+    expect(agentData.agent.name).toBe(agentName);
+    expect(agentData.agent.description).toBe(agentDescription);
+    expect(agentData.agent.stats).toBeDefined();
+    expect(agentData.agent.trophies).toBeDefined();
+    expect(Array.isArray(agentData.agent.trophies)).toBe(true);
+    expect(agentData.agent.hasUnclaimedRewards).toBe(false);
+    expect(agentData.agent.stats).toBeDefined();
+    // make sure public info is not exposed
+    expect(agentData.agent.apiKey).not.toBeDefined();
```
joewagner, line 932: As this is a semi-public route, we don't want to sned private info.

apps/api/e2e/utils/api-types.ts
```diff
@@ -121,6 +121,9 @@ export interface Agent {
   email?: string;
   apiKey?: string; // Only included in certain admin responses
   metadata?: AgentMetadata;
+  stats?: any; // eslint-disable-line @typescript-eslint/no-explicit-any
```
joewagner, line 124: I left this as `any` for now, but when we actually generate stats this should be typed

apps/api/src/controllers/agent.controller.ts
```diff
@@ -138,7 +145,7 @@ export function makeAgentController(services: ServiceRegistry) {
         });
         const totalCount = await services.agentManager.countAgents(filter);
 
-        // Return the competitions
+        // Return the agents
```
joewagner, line 148: cleanup from an older PR

apps/api/src/database/repositories/agent-repository.ts
```diff
@@ -93,6 +97,26 @@ export async function findAll(
   }
 }
 
+/**
+ * Find all competitions that given agent is, or has, participated in
+ * @param agentId the ID of the agent used for lookup
+ */
+export async function findAgentCompetitions(agentId: string) {
```
joewagner, line 104: note that there isn't any pagination here.  It will take a while for enough competitions to stack up that we would need pagination, so I left it out for simplicity.

apps/api/src/services/agent-manager.service.ts
```diff
@@ -819,4 +819,18 @@ export class AgentManager {
...
+  attachAgentMetrics(
+    sanitizedAgent: ReturnType<AgentManager["sanitizeAgent"]>,
+  ) {
+    const metadata = sanitizedAgent.metadata as any; // eslint-disable-line @typescript-eslint/no-explicit-any
+    const stats = metadata?.stats || {};
+
+    return {
+      ...sanitizedAgent,
+      stats,
+      trophies: metadata?.trophies || [],
```
joewagner, line 832: These values don't exist in the metadata currently, but I'm hoping the defaults will work for wiring up the front end.

apps/api/e2e/tests/agent.test.ts
dtbuchholz: nice, yeah that sounds like the right pattern

apps/api/src/services/agent-manager.service.ts
dtbuchholz: agreed, i'll keep an eye out and, if needed, just revert to keeping it within the metadata. but it seems like it should work. we could also try typing the `AgentMetadata` with these fields.

dtbuchholz (approved):
lgtm

PR #465, opened 2025-05-30
feat(api): auth login & fixup user/agent method

dtbuchholz:
in testing the frontend app, there were a few of problems i noticed:
- the user auth flows dont have a way to create a user upon sign up...so unless a user already exist in the db, they cant use the app. now, if the user doesn't exist in the db, we create an entry with their wallet address (after SIWE).
- other user/agent methods (like create an agent, or update user profile) handle request body/params by manually checking types & trimming. this allowed for certain NOT NULL columns to have empty strings inserted, or incorrect formats (e.g., url or email). instead, we start to use `zod` to handle this in a bit cleaner fashion.
- the `type` from competitions was missing. this is now added in the openapi spec.


dtbuchholz:
note: i basically removed `api-sdk` checks in https://github.com/recallnet/js-recall/commit/a358eeffde2644eebcdfd3e658d2e3d76ad560f8. instead, i think we can just be manually cognizant of when it does/doesnt need to be updated—and we'll probs move it outside of the monorepo. e.g., in this PR, i made a slight OpenAPI spec change, but i'm just going to regen it in some chore PR.

apps/api/src/types/index.ts
```diff
@@ -501,10 +503,126 @@ export type AgentFilter = z.infer<typeof AgentFilterSchema>;
...
 });
 
 export type CompetitionAgentParams = z.infer<
   typeof CompetitionAgentParamsSchema
 >;
+
+/**
+ * Update user profile parameters schema
+ */
+export const UpdateUserProfileBodySchema = z
```
dtbuchholz, line 517: added some new zod parsing types to help with some of the user/agent methods. they all use `strict` mode so that extra fields in POST requests will throw to maintain the existing behavior, and they make trim/length checks for arbitrary string values.

apps/api/src/routes/competitions.routes.ts
```diff
@@ -295,6 +303,10 @@ export function configureCompetitionsRoutes(
    *                       type: string
    *                       enum: [pending, active, ended]
    *                       description: Competition status
+   *                     type:
```
dtbuchholz, line 306: missing competitions datapoint in spec

apps/api/src/controllers/user.controller.ts
```diff
@@ -362,56 +303,19 @@ export function makeUserController(services: ServiceRegistry) {
...
+        const updateData = {
           id: agentId,
+          name: name ?? agent.name,
+          description,
+          imageUrl,
+          email,
+          metadata,
         };
 
-        if (name !== undefined) {
```
dtbuchholz, line 376: example of where we can just defer to zod instead of parsing all of these by hand

apps/api/src/services/auth.service.ts
```diff
@@ -50,9 +53,20 @@ export class AuthService {
       session.nonce = undefined;
 
       // Attempt to find the user that matches the verified wallet address
-      const user = await findByWalletAddress(siweData.address);
-      const userId = user?.id;
       const wallet = siweData.address;
+      let user = await findByWalletAddress(wallet);
```
dtbuchholz, line 57: the new auth logic where if you log in with a new wallet, we'll create the user if they dont already exist

joewagner (approved):
Awesome!

Issue #420, opened 2025-05-26
Implement new Competition Page design


ramirotw:
**Problem Statement**
Implements https://github.com/recallnet/js-recall/issues/408

**Impact**
Competitions App

**Priority**
High

PR #468, opened 2025-06-02
feat(comps): bump and integrate competition page

ramirotw:
#420 

image
localhost_3001_competitions_dd0b607a-a38d-47d8-98df-0c6d98cd211a


packages/ui2/tsconfig.json
```diff
@@ -6,6 +6,6 @@
       "@recallnet/ui2/*": ["./src/*"]
     }
   },
-  "include": ["."],
+  "include": [".", "../../apps/comps/components/create-agent.tsx"],
```
dtbuchholz: why is this needed?

ramirotw: it's not, good catch!

dtbuchholz (approved):

Issue #470, opened 2025-06-02
Add paging param to `/competitions` response


dtbuchholz:
per the [spec](https://docs.google.com/document/d/16Zu5-Rdk6DAimdzmK1sLCkFrgDHRCOmPgoWHlgqoaTU/edit?pli=1&tab=t.0#heading=h.jmu9rjwkswih), we're missing the following `metadata` field and paging-related info:
- metadata: 
   - total: number
   - limit: number
   - offset: number

PR #471, opened 2025-06-02
feat: add paging param to /competitions response

mzkrasner:
# Summary

Added pagination metadata to the `GET /competitions` endpoint to provide clients with essential information for implementing pagination UI and navigation. Closes https://github.com/recallnet/js-recall/issues/470.

## Problem

The `GET /competitions` endpoint accepted `limit` and `offset` query parameters but did not return pagination metadata in the response. This made it impossible for clients to:
- Know the total number of available competitions
- Determine if there are more results available
- Build proper pagination navigation

## Solution

Implemented pagination metadata across all layers of the application:

### New Response Structure
```json
{
  "success": true,
  "competitions": [...],
  "pagination": {
    "total": 25,        // Total number of competitions matching filter
    "limit": 10,        // Maximum results returned in this request  
    "offset": 0,        // Number of results skipped
    "hasMore": true     // Whether more results are available
  }
}
```

### Implementation Details

**Repository Layer** (`competition-repository.ts`):
- Updated `findByStatus` to execute parallel queries for data and count
- Returns `{competitions, total}` structure instead of just array
- Optimized with `Promise.all()` for better performance
- Maintains consistent filtering between data and count queries

**Service Layer** (`competition-manager.service.ts`):
- Updated JSDoc to reflect new return structure
- `getCompetitions` and `getUpcomingCompetitions` methods now return pagination data

**Controller Layer** (`competition.controller.ts`):
- Destructures `{competitions, total}` from service response
- Calculates `hasMore` using `(offset + limit) < total`
- Returns complete pagination object in API response

**OpenAPI Documentation** (`competitions.routes.ts`):
- Added new schema to response definition
- Documented all four pagination fields with descriptions and examples

**E2E Tests** (`competition.test.ts`):
- Updated existing test to validate pagination in `getUpcomingCompetitions()`
- Added comprehensive pagination test with multiple scenarios:
  - First page with `hasMore = true`
  - Second page with `hasMore = false` 
  - Beyond last page with empty results

**API-SDK Regeneration**
- Regenerated `api-sdk` given the changes in response format


apps/api/src/controllers/competition.controller.ts
```diff
@@ -461,15 +461,25 @@ export function makeCompetitionController(services: ServiceRegistry) {
...
 
-        // Return the competitions
+        // Calculate hasMore based on total and current page
+        const hasMore = pagingParams.offset + pagingParams.limit < total;
+
+        // Return the competitions with metadata
         res.status(200).json({
           success: true,
           competitions: competitions,
+          metadata: {
```
dtbuchholz: my fault for initially saying this should be `metadata`, but we want to call it `pagination`

joewagner: I think the spec doc calls it `metadata`. Maybe change to spec to pagination too?

dtbuchholz: > I think the spec doc calls it metadata. Maybe change to spec to pagination too?

agreed. some chats about this from the morning [here](https://recalllabs.slack.com/archives/C08NS37EMDM/p1748877446235329); `pagination` feels more accurate, anyways

apps/api/e2e/tests/competition.test.ts
```diff
@@ -514,6 +514,14 @@ describe("Competition API", () => {
     expect(Array.isArray(upcomingResponse.competitions)).toBe(true);
     expect(upcomingResponse.competitions.length).toBe(3);
 
+    // Verify metadata fields
+    expect(upcomingResponse.metadata).toBeDefined();
+    expect(upcomingResponse.metadata.total).toBe(3);
+    expect(upcomingResponse.metadata.limit).toBe(10); // default limit
+    expect(upcomingResponse.metadata.offset).toBe(0); // default offset
+    expect(typeof upcomingResponse.metadata.hasMore).toBe("boolean");
+    expect(upcomingResponse.metadata.hasMore).toBe(false); // 3 competitions < 10 limit
```
joewagner: The spec doesn't include the hasMore field. No problem if it's included, but seems good to keep that spec doc inline with what whatever we go with.

mzkrasner: I was referencing this issue: https://github.com/recallnet/js-recall/issues/470

Good call - we should ensure the spec doc and issues are always in sync

PR #482, opened 2025-06-03
feat(api): add `/leaderboard` endpoint

### LINKED ISSUE #359: Implement `/leaderboards` backend endpoint
GET /leaderboards

Query params:

- type
- limit
- offset

Response:

- metadata:
  - total: number
  - limit: number
  - offset: number
- stats:
  - activeAgents: number
  - totalTrades: number
  - totalVolume: number
- agents[]: 
  - id: string
  - name: string
  - imageUrl: string
  - metadata: AgentCompetitionMetadata // { “walletAddress”: “0x”, “roi”: 0.55, “trades”: 23 }
  - score: number

dtbuchholz:
## Summary

Add the `/leaderboard` (singular) endpoint; this slightly deviates from the spec because we only have a single leaderboard  page vs the old designs with multiple leaderboards. Closes #359.

## Details

Reuses some of the competition manager's logic around calculating agent scores (via `getCompetitionAgents`). But, it accumulates the scores across all competitions, and ranks them based on combined PnL percentage. This logic is slightly different from the competition-specific leaderboard ranking, which simply uses the agent's total portfolio value.

This is a temporary global rank until we have a real elo-like system in place.


apps/api/e2e/utils/api-types.ts
```diff
@@ -658,3 +658,32 @@ export interface CompetitionLeaveResponse extends ApiResponse {
   success: true;
   message: string;
 }
+
+// Agent rank (global rankings)
+export interface AgentRank {
```
dtbuchholz, line 658: some type to encapsulate the `AgentRank` that will eventually come from a new `leaderboards` or `rankings` table + elo score (tbd on future schema).

apps/api/e2e/utils/api-types.ts
```diff
@@ -658,3 +658,32 @@ export interface CompetitionLeaveResponse extends ApiResponse {
...
+  metadata?: AgentMetadata;
+  rank: number;
+  score: number;
+  numCompetitions: number;
+}
+
+// Global leaderboard response
+export interface GlobalLeaderboardResponse extends ApiResponse {
+  success: true;
+  stats: {
```
dtbuchholz, line 671: and then we use the global `AgentRank` in our leaderboards response, along with global stats about all of the competitions (stats are needed for ui-related information)

apps/api/openapi/openapi.json
```diff
@@ -4281,6 +4302,150 @@
...
+      "get": {
+        "tags": [
+          "Leaderboard"
+        ],
+        "summary": "Get global leaderboard",
+        "description": "Get global leaderboard data across all relevant competitions",
+        "parameters": [
+          {
+            "in": "query",
+            "name": "type",
```
dtbuchholz, line 4320: technically, you can get the global leaderboard w/ a query param: `/leaderboard?type=trading`. i figured we can keep this in case we need thematic leaderboards in the future. since we only have one type of comp, `/leaderboard` just defaults to `?type=trading`

apps/api/src/controllers/leaderboard.controller.ts
```diff
@@ -0,0 +1,162 @@
...
+
+        if (stats.competitionIds.length === 0) {
+          throw new ApiError(404, "No competitions found");
+        }
+
+        // Collect agents and their metrics across all competitions
+        const agentMetricsMap = new Map<string, AgentRank>();
+
+        // Process each competition (need to get all agents in order to properly calculate metrics)
+        await Promise.all(
```
dtbuchholz, line 54: not the most efficient logic here, but since it's temporary, i figured it'd be okay to keep w/o optimizing. this is where we copy the per-competition leaderboard/agent metrics logic, but at the end, we use the combined PnL % as the rank.

apps/api/src/index.ts
```diff
@@ -79,21 +82,9 @@ app.use(express.json());
...
-const agentApiKeyRoutes = [
-  "/api/agent",
-  "/api/trade", // Trade operations use agent API keys
-];
-
-const userSessionRoutes = [
-  "/api/user", // User operations use SIWE sessions
-];
-
-const legacyRoutes = [
```
dtbuchholz, line 91: note: i removed the `legacyRoutes` and duplicative auth middleware. this was left over from the refactor. i just wanna double check the auth settings are correct here.

dtbuchholz:
@mzkrasner @joewagner lmk if you have questions here or see any logic issues. it's not the cleanest setup but a temp solution pre-rankings w/ elo scores; we'll replace this with the real leaderboard ranking logic in the next 2-3 weeks.

joewagner (approved):
LGTM!
As for the update to auth middleware, I think this is right, but I haven't really seen that kind of thing specked out anywhere.

Issue #486, opened 2025-06-03
Allow user to update metadata


dtbuchholz:
Users APIs do not let them update their profile info. The frontend need to add arbitrary user-related fields, like a website url, but is blocked by the inability to do so.

The API was initially designed this way to that users could update something like app-controlled user preferences. However, user auth is dicated by SIWE to our application endpoints, so this isn't an issue.

PR #487, opened 2025-06-03
feat(api): allow user to update metadata

dtbuchholz:
allows users to update their `metadata` field, which is needed for frontend app purposes. we also expose this property in user-related response since it's relevant for displaying profile information. closes https://github.com/recallnet/js-recall/issues/486.

### DIFF
--- a/apps/api/e2e/utils/api-types.ts
+++ b/apps/api/e2e/utils/api-types.ts
@@ -86,12 +86,7 @@ export interface User {
 
 // User metadata structure
 export interface UserMetadata {
-  preferences?: {
-    notifications?: boolean;
-    theme?: "light" | "dark";
-  };
-  settings?: Record<string, unknown>;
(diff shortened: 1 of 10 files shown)


apps/api/e2e/utils/api-types.ts
```diff
@@ -86,12 +86,7 @@ export interface User {
 
 // User metadata structure
 export interface UserMetadata {
-  preferences?: {
-    notifications?: boolean;
-    theme?: "light" | "dark";
-  };
-  settings?: Record<string, unknown>;
-  custom?: Record<string, unknown>;
+  [key: string]: unknown;
```
dtbuchholz, line 89: with #426, we can start to clean up the expected frontend vs backend metadata fields. the fields in this interface were arbitrary values created during the teams overhaul, and they're not explicitly needed.

PR #498, opened 2025-06-04
feat(comps): improve auth flow

ramirotw:
This PR improves the overall authentication flow: login, logout, redirects, user state management and also some other misc stuff regarding the user profile.

misc changes:
* moved ui2 components from the shadcn folder to the root. This change triggered prettier over these files (before they were ignored as the shadcn folder was in the .prettierignore file)
* improvements in some components such as buttons and dropdown
* user profile tweaks and fixes (e.g. handle metadata field correctly)
* navbar tweaks


apps/comps/components/providers.tsx
```diff
@@ -57,15 +54,13 @@ function WalletProvider(props: { children: ReactNode }) {
           wallet: siweMessage.address,
```
dtbuchholz, line 54: to resolve the linting error, the `const res = await login(...)` should just be `await login(...)` since `res` is no longer used

dtbuchholz (approved):
other than that linting issue, lgtm!

apps/comps/components/providers.tsx
ramirotw: yep, I forgot this one

PR #500, opened 2025-06-04
feat(api): use `/agents` plural endpoint & allow open `/competitions` reads

dtbuchholz:
- a few of the new agent endpoints used `/agent` instead of `/agents`
- the `/competitions` API needs open access so non-auth users can see data


dtbuchholz:
cc @ramirotw re: the `/competitions` endpoint and access issues

PR #501, opened 2025-06-04
fix(comps): small fixes and cleanup

joewagner:
## Overview

I ran through starting up a dev environment for both `api` and `comps` starting from scratch this morning, and I found a few issues.

## Details 

These aren't really related but I noticed a few things:
 - The README doesn't explain how to get a WalletConnect Cloud project_id
 - There is an invalid sort value for agents.

### DIFF
--- a/apps/comps/README.md
+++ b/apps/comps/README.md
@@ -85,10 +85,10 @@ In another terminal, change into the `comps` directory (aka this directory):
 cd apps/comps
 ```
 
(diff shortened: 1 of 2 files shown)


joewagner:
@dtbuchholz This is kind of a drive-by commit. So if there's other PRs or tickets that capture this, we don't need to merge this.

dtbuchholz:
@joewagner nice, yeah this works! 

one thing—i noticed a related `/agent` issue as well. i think it's actually: we _want_ the plural endpoint, and not the singular one. see the changes i started to make here: https://github.com/recallnet/js-recall/pull/500/files#diff-8fd2632f0054ce7579f0448b55c4c3c9c8eb0bb664deb5b8211c36e91e80005c. we should have `/agents/:id` and `/agents/:id/competitions`, but those two are scoped to the singular `/agent` (which is more geared to single-agent activities and bearer auth).

PR #507, opened 2025-06-05
feat(comps): leaderboard & minor fixups

dtbuchholz:
- the default bg color was different than `bg-black`, so you could see a slight border-like rendering on pages. e.g., the area surrounding the landing page's video.
- the `/leaderboard` API was changed to singular, and now that it's live, the leaderboards component can properly use the data. 
   - the top positions (1st, 2nd, 3rd) were displayed correctly, but numbers thereafter started at "3"...this has been fixed to properly increment values.
   - some of the datapoints are also hidden, including the "change in rank" arrow and the "score". 
   - note: the score exists in the API response, but we should ignore it for now since it's a somewhat arbitrary things until real ranks arrive.
- fixed the BigNumber component. it was causing issues for the leaderboard's `totalVolume` field when decimals were present
- updated some API client types and methods:
   - `CompetitionStatus.Ended` has replaced `...Completed`
   - the routes for getting an agent's profile used the user-scoped `/user/agents/:id` endpoint, so this has been replaced with the generic `/agents/:id` endpoint to prevent auth issues where necessary.

### DIFF
--- a/apps/comps/components/leaderboard-table.tsx
+++ b/apps/comps/components/leaderboard-table.tsx
@@ -29,7 +29,7 @@ const emptyAgent: (i: number) => LeaderboardAgent = (i: number) => ({
   status: "",
   rank: i,
   score: 0,
-  totalVotes: 0,
+  voteCount: 0,
   numCompetitions: 0,
 });
 
@@ -175,7 +175,7 @@ export function LeaderboardTable({
               <TableCell className="flex items-center justify-end pr-10 text-gray-500 sm:pr-0">
                 {loaded ? (
                   // TODO: this is not part of the API response, yet, but it will be guaranteed to be present via the `useLeaderboards` hook
-                  <>{agent.totalVotes || 0}</>
+                  <>{agent.voteCount || 0}</>
                 ) : (
                   <Skeleton className="h-2 w-10 rounded-full" />
                 )}
--- a/apps/comps/components/leaderboard-table.tsx
+++ b/apps/comps/components/leaderboard-table.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { ArrowUp, AwardIcon, ExternalLink, Trophy } from "lucide-react";
+import { AwardIcon, ExternalLink, Trophy } from "lucide-react";
 import Image from "next/image";
 import Link from "next/link";
 
--- a/apps/comps/components/leaderboard-table.tsx
+++ b/apps/comps/components/leaderboard-table.tsx
@@ -2,6 +2,7 @@
 
 import { ArrowUp, AwardIcon, ExternalLink, Trophy } from "lucide-react";
 import Image from "next/image";
+import Link from "next/link";
 
 import { Button } from "@recallnet/ui2/components/button";
 import { Skeleton } from "@recallnet/ui2/components/skeleton";
@@ -16,9 +17,9 @@ import {
 } from "@recallnet/ui2/components/table";
 import { cn } from "@recallnet/ui2/lib/utils";
 
-import { Agent } from "@/types/agent";
+import { LeaderboardAgent } from "@/types/agent";
 
-const emptyAgent: (i: number) => Agent = (i: number) => ({
+const emptyAgent: (i: number) => LeaderboardAgent = (i: number) => ({
   id: i.toString(),
   name: `some-name-${i}`,
   walletAddress: "",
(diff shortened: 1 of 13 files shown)


apps/comps/components/leaderboard-table.tsx
```diff
@@ -88,23 +94,35 @@ export function LeaderboardTable({
...
                   </div>
                 ) : (
                   <div className="mx-6 flex items-center justify-center rounded bg-gray-800 px-2 py-3">
-                    {i}
+                    {i + 1}
                   </div>
                 )}
-                <span className="text-gray-500">{"23,533"}</span>
+                {/* TODO: hide this until official "scores" are implemented; the API does return the data, but it's not elo-based */}
+                {/* <span className="text-gray-500">{agent.score}</span> */}
```
dtbuchholz, line 125: the API currently returns an ugly score with many decimals. for the short term, let's just hide it since it will lead to user confusion on it's meaning—until we have the official ranking / scores in the next week or two

apps/comps/components/leaderboard-table.tsx
```diff
@@ -74,10 +79,11 @@ export function LeaderboardTable({
             >
               <TableCell className="xs:flex-row flex flex-col items-center gap-2 py-6">
                 <div className="flex gap-2">
-                  <ArrowUp size={20} className="text-green-500" />
-                  <span>3</span>
+                  {/* TODO: enable once we have official rankings & net change in rank */}
```
dtbuchholz, line 82: same theme here as the comment below...since we cant show change in rank, we might as well hide it, for now

apps/comps/components/leaderboard-table.tsx
```diff
@@ -88,23 +94,35 @@ export function LeaderboardTable({
                     <Trophy size={17} />
                     <span>1st</span>
                   </div>
-                ) : i < 3 ? (
+                ) : i === 1 ? (
```
dtbuchholz, line 97: logic fixes for the 1st/2nd/3rd place and beyond

note: i didnt set up / test the pagination, so i'm not sure if this will hold if you e.g., sort in reverse order. can re-test / adjust once that's working

apps/comps/components/leaderboard-table.tsx
```diff
@@ -146,22 +166,25 @@ export function LeaderboardTable({
...
                 ) : (
                   <Skeleton className="h-2 w-10 rounded-full" />
                 )}
               </TableCell>
 
               <TableCell className="flex items-center justify-end pr-10 text-gray-500 sm:pr-0">
                 {loaded ? (
-                  "22,550"
+                  // TODO: this is not part of the API response, yet, but it will be guaranteed to be present via the `useLeaderboards` hook
+                  <>{agent.totalVotes || 0}</>
```
dtbuchholz: we'll add this https://github.com/recallnet/js-recall/issues/508

apps/comps/components/ongoing-competition.tsx
```diff
@@ -33,7 +33,7 @@ export const OngoingCompetition: React.FC<OngoingCompetitionProps> = ({
   return (
     <div className="bg-card p-8">
       <div className="mb-30 flex items-start justify-between">
-        <StringList strings={["ONGOING", ...competition.type]} />
+        <StringList strings={["ONGOING", competition.type]} />
```
dtbuchholz, line 36: this was a bug that caused the `competition.type` to get split and look like `ONGOING / T / R / A / D / I / N / G`

apps/comps/lib/api-client.ts
```diff
@@ -200,14 +211,12 @@ export class ApiClient {
...
-   * Get agent by ID
+   * Get agent by ID (unauthenticated)
    * @param id - Agent ID
    * @returns Agent details
    */
   async getAgent(id: string): Promise<{ success: boolean; agent: Agent }> {
-    return this.request<{ success: boolean; agent: Agent }>(
-      `/user/agents/${id}`,
-    );
+    return this.request<{ success: boolean; agent: Agent }>(`/agents/${id}`);
```
dtbuchholz, line 219: uses the public `/agents/:id` route, which prevents auth issues vs the older `/user/agents/:id`

apps/comps/types/agent.ts
```diff
@@ -53,12 +64,13 @@ export interface LeaderboardStats {
   activeAgents: number;
   totalTrades: number;
   totalVolume: number;
+  totalCompetitions: number;
```
dtbuchholz, line 67: unused in the UI, but if we want it, we can include it in the "summary" section of the leaderboard (top section)

ramirotw (approved):
nice! 

regarding pagination and sorting, I can work on that later

PR #510, opened 2025-06-05
feat: api prefix routing

mzkrasner:
# Summary

- Assuming we are going to continue using prefixed routing in the load balancer off of `https://api.competitions.recall.network/` (such as `sandbox`), we should standardize our approach that works both when running the server outside and inside containerization
- These changes primarily edit the main server file to absorb the routing, and uses the dependency injection `config` object, but also impacts the health check endpoint within the docker configuration, as well as ensuring our client is properly constructed in `apps/api/e2e`

### DIFF
--- a/.github/workflows/api-ci.yml
+++ b/.github/workflows/api-ci.yml
@@ -257,6 +257,9 @@ jobs:
           DISABLE_PARTICIPANT_LEADERBOARD_ACCESS=false
           EOF
 
+          # Optional: API prefix
+          API_PREFIX=testing
+
       - name: Run E2E tests
         working-directory: apps/api
         run: |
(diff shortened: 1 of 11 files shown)


packages/contracts/contracts
dtbuchholz, line 1: looks like the old `contracts` submodule was accidentally included here

mzkrasner: whoops - will remedy here in a moment

dtbuchholz:
> Assuming we are going to continue using prefixed routing in the load balancer off of https://api.competitions.recall.network/ (such as sandbox), we should standardize our approach that works both when running the server outside and inside containerization

note: short term, yes. long term, this is the incorrect assumption. with the IaC approach, we'll have:
- a single URL for both production and testing—without an api path prefix. like Stripe, keys will look something like `sk_test_..` or `sk_prod_...`, so the key dictates the underlying db queries
- our staging env will have its own subdomain

Issue #497, opened 2025-06-04
Add new APIs for user profile + competitions hub / page


dtbuchholz:
**Problem Statement**
The comps app needs to get all competitions that a user's agent is participating in.

**Proposed Solution**
Add the following API:
- Description: get all of the competitions a user's agent(s) have participated in
   - i.e., the `agents` are those who are registered for a pending competition, participating in an `active` one, or participated in an `ended` one
- Endpoint: GET /api/user/competitions
- Query params: `sort`, `limit`, `offset`, `competitionId`
- Response: 
   - `success`: boolean
   - `competitions`: array of objects (competition info + underlying agents)
      - `id`: uuid for the competition
      - `name`: competition name
      - `status`: competition status (pending, active, ended)
      - `agents`: array of objects
        - `id`: uuid for agent
         - `name`: agent name
         - `description`: agent description
         - `walletAddress`: agent wallet address
         - `rank`: rank in the competition (note: since we dont have ranking system, i dont think it needs an actual value)
   - `pagination`: `total`, `limit`, `offset`, `hasMore`

**Impact**
frontend + backend

**Priority**
high/highest. needed for new user/comps designs

stbrody:
Remaining work:
- looking through snapshots for different agents in different competitions to calculate the ranking - work is done but working on tests now

PR #518, opened 2025-06-05
feat(api): add endpoint to get user's comps

joewagner:
fixes: https://github.com/recallnet/js-recall/issues/497

## Summary

This creates the `/api/user/competitions` endpoint that returns all competitions a user's agents are participating in. 

## Details

This adds to agent-repository with a `findUserAgentCompetitions` function that queries competitions for multiple agent IDs.
The agent manager service  has a new function `getCompetitionsForUserAgents`.  This gets all agents owned by a user, so that the `findUserAgentCompetitions` can be called with the user's agent ids.
The user controller method `getCompetitions` handles parsing request params/pagination/etc...

The testing is a little limited since there is still a limitation that only one comp can be active at a time.  There was a test helper added to quickly setup multiple users, agents, and competitions.

### DIFF
--- a/apps/api/e2e/tests/user.test.ts
+++ b/apps/api/e2e/tests/user.test.ts
@@ -1,14 +1,19 @@
 import axios from "axios";
 import { beforeEach, describe, expect, test } from "vitest";
 
+import { config } from "@/config/index.js";
 import {
   AdminSearchUsersAndAgentsResponse,
   AdminUsersListResponse,
   Agent,
   AgentProfileResponse,
+  CROSS_CHAIN_TRADING_TYPE,
   ErrorResponse,
   GetUserAgentsResponse,
+  StartCompetitionResponse,
+  TradeResponse,
   UserAgentApiKeyResponse,
+  UserCompetitionsResponse,
   UserProfileResponse,
 } from "@/e2e/utils/api-types.js";
 import { getBaseUrl } from "@/e2e/utils/server.js";
@@ -19,7 +24,9 @@ import {
   cleanupTestState,
   createSiweAuthenticatedClient,
   createTestClient,
+  generateTestCompetitions,
   registerUserAndAgentAndGetClient,
+  wait,
 } from "@/e2e/utils/test-helpers.js";
 
 describe("User API", () => {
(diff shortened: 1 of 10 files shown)


apps/api/e2e/tests/user.test.ts
```diff
@@ -477,4 +479,31 @@ describe("User API", () => {
       "Invalid request format",
     );
   });
+
+  test("SIWE user can get competitions for their agents", async () => {
```
joewagner, line 1129: When we enable multiple active competitions, we should add more tests for this endpoint.

apps/api/e2e/utils/api-types.ts
```diff
@@ -385,6 +385,18 @@ export interface UpcomingCompetitionsResponse extends ApiResponse {
   };
 }
 
+// User competitions response
+export interface UserCompetitionsResponse extends ApiResponse {
```
joewagner, line 404: test response type

apps/api/e2e/utils/test-helpers.ts
```diff
@@ -370,3 +371,67 @@ export async function createSiweAuthenticatedClient({
     loginData: loginResponse,
   };
 }
+
+export async function generateTestCompetitions(adminApiKey: string) {
```
joewagner, line 411: This is helper that creates 3 users, 3 agents (one for each), and 20 competitions.  agent 1 is participating in 15 comps, agent 2 is in 13, and agent 3 is in 11.  All comps are pending unless the test starts one.

dtbuchholz: not essential, but maybe a comment about this behavior or configurable params would help make sure that it's clear to the dev about what to expect here.

apps/api/openapi/openapi.json
```diff
@@ -6009,6 +6009,178 @@
...
+                  "properties": {
+                    "success": {
+                      "type": "boolean",
+                      "example": true
+                    },
+                    "userId": {
+                      "type": "string",
+                      "format": "uuid"
+                    },
+                    "competitions": {
```
dtbuchholz, line 6434: i think this is missing the `agents` object in #497? (sorry, i shouldve spec'd out the actual object instead of using bullet points)

i.e., the `competitions` array should include competition objects which each include an `agents` array, and _this_ array should include which of the user's agents are in that specific competition
image

apps/api/src/database/repositories/agent-repository.ts
```diff
@@ -170,6 +171,86 @@ export async function findAgentCompetitions(
...
+      .where(and(...whereConditions))
+      .$dynamic();
+
+    if (params.sort) {
+      query = getSort(query, params.sort, agentCompetitionsOrderByFields);
+    }
+
+    query.limit(params.limit).offset(params.offset);
+
+    const results = await query;
```
dtbuchholz, line 223: re: the point above, i think this `results` object should have the agents-related info needed to incorporate it in the final response down below

apps/api/openapi/openapi.json
dtbuchholz:
```json
{
  "competitions": [
    {
      "id": "...",
      "name": "Example Competition",
      "status": "...",
      // can include other `Competition` fields, although UI doesnt require it
      "agents": [
        {
          "id": "...",
          "name": "Example Agent",
          "description": "Short description of what the agent does.",
          "walletAddress": "0xabc123...def",
          "rank": 1234
        }
      ]
    }
  ]
}
```

apps/api/e2e/utils/test-helpers.ts
joewagner: Really good call, I'll add that

apps/api/openapi/openapi.json
joewagner: Ah, I see.  So the `competitions.agents` array is filtered to only contain agents owned by the user making the request?

dtbuchholz: > Ah, I see. So the competitions.agents array is filtered to only contain agents owned by the user making the request?

correct! the UI wants to know: _what competitions (active, pending, or ended) have my agents been involved in, and for each of those competitions, which agent(s) are involved_

apps/api/e2e/tests/user.test.ts
```diff
@@ -1118,4 +1125,142 @@ describe("User API", () => {
...
+    const competitionsWithParamsResponse = (await client1.getUserCompetitions({
+      limit: 5,
+      offset: 0,
+    })) as UserCompetitionsResponse;
+
+    expect(competitionsWithParamsResponse.success).toBe(true);
+    expect(competitionsWithParamsResponse.pagination.limit).toBe(5);
+  });
+
+  test("user can get competitions for their agents with correct rank", async () => {
```
joewagner, line 1168: Added a test for rank

apps/api/openapi/openapi.json
joewagner: I added agents, which includes rank.  You'll see the new logic for calculating rank below.

apps/api/src/services/agent-manager.service.ts
```diff
@@ -808,6 +814,100 @@ export class AgentManager {
...
+          comp.agents.push(agent);
+        }
+
+        agentCompetitions.set(data.competitions.id, comp);
+      });
+
+      // TODO: since rankings are not really done, need to make a bunch of db calls and get rankings
+      //  for every competition in this result set and make sure to append the agent's rank to the
+      //  response data.
+      await Promise.all(
```
joewagner, line 877: This is the place rank is calculated and attached to the response data.  It all seems a bit hacky since we are making potentially 10+ database requests (one for each competition), but I don't know how else to do this right now.

dtbuchholz: potentially related to #546

dtbuchholz (approved):
lgtm! and now that the agent rank stuff is prioritized, we should (soon) be able to revisit / optimize the logic around some of your callouts

apps/api/src/types/index.ts
```diff
@@ -190,13 +190,14 @@ export interface AdminMetadata {
...
  * Agent's metadata
  */
-export interface AgentMetadata {
-  stats?: Record<string, unknown>;
-  skills?: string[];
-  trophies?: string[];
-  hasUnclaimedRewards?: boolean;
-  [key: string]: unknown;
-}
+export const AgentMetadataSchema = z.looseObject({
```
dtbuchholz, line 193: (just making reference to #403 and #524 since those are about zod and/or fixing up our types—aka what you have here can be used there)

PR #522, opened 2025-06-06
fix(api): include agent wallet in res

dtbuchholz:
- fixes a missing `walletAddress` in the `getAgent` / `getAgents` response
- also, adds some zod parsing

### DIFF
--- a/apps/api/src/controllers/user.controller.ts
+++ b/apps/api/src/controllers/user.controller.ts
@@ -27,10 +27,7 @@ export function makeUserController(services: ServiceRegistry) {
       try {
         const { success, data, error } = UuidSchema.safeParse(req.userId);
         if (!success) {
-          throw new ApiError(
-            400,
-            `Invalid request format: ${error.message}, actual value: ${req.userId}`,
-          );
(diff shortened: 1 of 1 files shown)


apps/api/src/controllers/user.controller.ts
```diff
@@ -185,26 +159,21 @@ export function makeUserController(services: ServiceRegistry) {
      */
     async getAgents(req: Request, res: Response, next: NextFunction) {
       try {
-        const userId = req.userId as string;
+        const { success, data, error } = UuidSchema.safeParse(req.userId);
```
joewagner, line 162: No need to do this in this PR, but we seem to be repeating this code in a few places.  Probably worth creating an `ensureUserId(req: Request)` helper that returns the userId.

joewagner (approved):
Nice!

apps/api/src/controllers/user.controller.ts
dtbuchholz: good call. i added a note in this issue https://github.com/recallnet/js-recall/issues/426

Issue #488, opened 2025-06-03
join competition logic + modals


Rcontre360:
join competition logic, check 

https://protofire.slack.com/archives/C08NS37EMDM/p1748562358843939
and
https://www.figma.com/design/uuMfHT5O7zW4dfTgj6ryGG/Wednesday-s-Designs?node-id=10798-32680

Issue #512, opened 2025-06-05
Competition page


dtbuchholz:
- [ ] Add the stats now that the API returns them
- [ ] Integrate the Join Competition Flow Rafael was working on

PR #523, opened 2025-06-06
feat(comps): join competition flow

ramirotw:
Depends on https://github.com/recallnet/js-recall/pull/518
Fixes https://github.com/recallnet/js-recall/issues/488 https://github.com/recallnet/js-recall/issues/512

* Integrates the Join Competition Flow in the Competition page
* Integrates the competition stats into the Competition page


apps/comps/app/create-agent/page.tsx
```diff
@@ -74,3 +80,13 @@ export default function CreateAgentPage() {
...
   );
 }
+
+export default function CreateAgentPage() {
+  return (
+    <Suspense
+      fallback={<div className="py-12 text-center text-white">Loading...</div>}
+    >
+      <CreateAgentView />
+    </Suspense>
```
ramirotw, line 90: this was to fix the following error

>  ⨯ useSearchParams() should be wrapped in a suspense boundary at page "/create-agent". Read more: https://nextjs.org/docs/messages/missing-suspense-with-csr-bailout

apps/comps/app/layout.tsx
```diff
@@ -20,11 +21,12 @@ export default function RootLayout({
   children: React.ReactNode;
 }>) {
   return (
-    <html lang="en" suppressHydrationWarning className="overflow-x-hidden">
+    <html lang="en" suppressHydrationWarning>
```
ramirotw, line 24: @dtbuchholz I reverted the change that was meant to fix an issue on the landing page because it was causing a side effect when popovers were rendered. Since the landing page wasn’t fully functional and isn’t part of Jovi’s new designs, I’ve set it to redirect to /competitions, which is supposed to be the landing page for now

apps/comps/components/agents-table/rank-badge.tsx
```diff
@@ -9,11 +9,11 @@ interface RankBadgeProps {
...
   const getBackgroundColor = () => {
     switch (position) {
-      case 0:
-        return "bg-yellow-800";
       case 1:
-        return "bg-gray-700";
+        return "bg-yellow-800";
       case 2:
+        return "bg-gray-700";
+      case 3:
```
ramirotw, line 16: position is not zero indexed

Rcontre360 (approved):

apps/comps/app/layout.tsx
dtbuchholz: make sense. i had a feeling it might've been a temporary solution. 

re: new landing page—yeah, that make sense

Issue #509, opened 2025-06-05
Fix pagination logic for app<>API consistency


dtbuchholz:
The APIs below may or may not support pagination. They either need to be implemented altogether, or updated to match the per-column sorting requirements in the frontend. Namely, the `sort` fields need to be properly matched. Note: keep in mind #426 for the parsing the query params. The pattern [here](https://github.com/recallnet/js-recall/blob/afc9bf0e5bf5566bdbaee38c075a3a5aba0dea09/apps/api/src/controllers/user.controller.ts#L62) seems to work well because it parses body the req, params, and body all at once.

### `/agents`

- `/`: `name`, `-name`

etc...

dtbuchholz:
@joewagner it looks like this API is another to look at: `/api/agents/{agentId}/competitions` (mentioned via https://github.com/recallnet/js-recall/pull/531#pullrequestreview-2910041433)

joewagner:
> [@joewagner](https://github.com/joewagner) it looks like this API is another to look at: `/api/agents/{agentId}/competitions` (mentioned via [#531 (review)](https://github.com/recallnet/js-recall/pull/531#pullrequestreview-2910041433))

@dtbuchholz It looks like the spec was out of date, but filtering and pagination was already setup.  Here is a [PR](https://github.com/recallnet/js-recall/pull/563) to clean up the spec and make some changes for https://github.com/recallnet/js-recall/issues/426

joewagner:
As best I can tell this is done.  We can reopen if we find any endpoints that need to be fixed.

PR #529, opened 2025-06-06
fix(api): include `hasMore` in /agents response

joewagner:
part of #509 

Very small PR here.  I'll keep adding to this as I find more places the api needs fixes to pagination and sorting, but feel free to review.


dtbuchholz (approved):
🚀

PR #533, opened 2025-06-09
feat: testnet v0.1.9

dtbuchholz:


dtbuchholz:
closing since the testnet is broken with these values

PR #534, opened 2025-06-09
chore(api): remove default API_PREFIX value

dtbuchholz:
The addition of `API_PREFIX` in the `.env` file with a preset value led to unexpected behavior within the scripts and the expected API URL to use. E.g., users will copy the `.env.example` to `.env`, and then accidentally use an API path prefix. This simply sets an empty value for the env var so that people can use the API w/o this issue.


mzkrasner (approved):
Makes sense

PR #552, opened 2025-06-09
Update feature request template

stbrody:
I've noticed some people using the "impact" section to list the specific changes that will be made.  That belongs more in the "Proposed Solution" section.  The impact section is more useful as input to deciding issue priority, so should be focused on how making this change will improve user experience, or on what the cost of *not* doing this will be.  I'm hoping this reword to the template will make that clearer

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -2,8 +2,9 @@
 name: Feature request
 about: Suggest an idea for this project
 title: "[FEATURE]"
-labels: ""
-assignees: ""
+labels: ''
+assignees: ''
+
 ---
 
 **Problem Statement**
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -13,7 +14,7 @@ What is the problem you are trying to solve? Ex. I'm always frustrated when [...
 A clear and concise description of what you want to happen.
 
 **Impact**
-Who is impacted by this change? How will it affect them?
+What is the effect on the end user of doing this or not doing this?  Why is this something that needs to be done?
```
stbrody:
```suggestion
What is the effect on the end user of doing this?  What is the cost or risk of not doing this?  Why is this something that needs to be done?
```

Issue #546, opened 2025-06-09
Refactor to store competitions final leaderboard in the database


brunocalza:
Currently, the competitions results are computed on the fly and is a very complicated calculation based on portfolio snapshots. We should compute the competitions final leaderboard at the end of every competition and store them. That will make things easier for UI grabbing that info and for agent rank calculations

Suggestion on how to implement this:

- Refactor [`endCompetition`](https://github.com/recallnet/js-recall/blob/47fa4594ea86172a5367cd907d0d94a1e6cb6fea/apps/api/src/services/competition-manager.service.ts#L209) to add a [leaderboard calculation](https://github.com/recallnet/js-recall/blob/47fa4594ea86172a5367cd907d0d94a1e6cb6fea/apps/api/src/services/competition-manager.service.ts#L308) and storage of it

stbrody:
This might be unnecessary - you can just query the most recent competition snapshot instead

PR #557, opened 2025-06-10
feat: adds agent rank tables

### LINKED ISSUE #545: Agent Rank database migration
Implements the following database migration according to [Implementation Plan for Agent Rank (OpenSkill)](https://www.notion.so/Implementation-Plan-for-Agent-Rank-OpenSkill-20adfc9427de803db1e7c47d52d52639)

## `agent_rank`

Stores the current agent rank for an agent

| Column | Type | Constraints |
| --- | --- | --- |
| id | UUID | PRIMARY KEY NOT NULL |
| agent_id | UUID | NOT NULL REFERENCES agents.id |
| mu | NUMERIC(6,2) | NOT NULL |
| sigma | NUMERIC(6,2) | NOT NULL |
| created_at | TIMESTAMPTZ | NOT NULL (with timezone) |
| updated_at | TIMESTAMPTZ | NOT NULL (with timezone) |

Alternatively, we can simply add the `mu` and `sigma` columns to the `agents` table instead of creating another table

## `agent_rank_history`

Stores the history of every agent rank score after each competition

| Column | Type | Constraints |
| --- | --- | --- |
| id | UUID | PRIMARY KEY NOT NULL |
| agent_id | UUID | NOT NULL REFERENCES agents.id |
| competition_id | UUID | NOT NULL REFERENCES competitions.id |
| mu | NUMERIC(6,2) | NOT NULL |
| sigma | NUMERIC(6,2) | NOT NULL |
| created_at | TIMESTAMPTZ | NOT NULL (with timezone) |

## `competitions_leaderboard`

Stores competitions final leaderboard results when it ends

| Column | Type | Constraints |
| --- | --- | --- |
| id | UUID | PRIMARY KEY NOT NULL |
| agent_id | UUID | NOT NULL REFERENCES agents.id |
| competition_id | UUID | NOT NULL REFERENCES competitions.id |
| rank | INTEGER | NOT NULL |
| score | NUMERIC | NOT NULL |
| created_at | TIMESTAMPTZ | NOT NULL (with timezone) |

brunocalza:
Closes #545

### DIFF
--- a/apps/api/drizzle/0007_big_gargoyle.sql
+++ b/apps/api/drizzle/0007_big_gargoyle.sql
@@ -1,0 +1,25 @@
+CREATE TABLE "agent_rank" (
+	"id" uuid PRIMARY KEY NOT NULL,
+	"agent_id" uuid NOT NULL,
+	"mu" numeric(6, 2) NOT NULL,
+	"sigma" numeric(6, 2) NOT NULL,
+	"created_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"updated_at" timestamp with time zone DEFAULT now() NOT NULL
+);
+--> statement-breakpoint
+CREATE TABLE "agent_rank_history" (
+	"id" uuid PRIMARY KEY NOT NULL,
+	"agent_id" uuid NOT NULL,
+	"competition_id" uuid NOT NULL,
+	"mu" numeric(6, 2) NOT NULL,
+	"sigma" numeric(6, 2) NOT NULL,
+	"created_at" timestamp with time zone DEFAULT now() NOT NULL
+);
+--> statement-breakpoint
+ALTER TABLE "agent_rank" ADD CONSTRAINT "agent_rank_agent_id_fkey" FOREIGN KEY ("agent_id") REFERENCES "public"."agents"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "agent_rank_history" ADD CONSTRAINT "agent_rank_history_agent_id_fkey" FOREIGN KEY ("agent_id") REFERENCES "public"."agents"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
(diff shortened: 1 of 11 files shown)


dtbuchholz (approved):
should we create `competitions_leaderboard` just so we can reduce future migrations? or we can just handle it in https://github.com/recallnet/js-recall/issues/546

brunocalza:
> should we create `competitions_leaderboard` just so we can reduce future migrations? or we can just handle it in #546

I think so, if we are certain about it. Today in the call looked like there were some doubts about the need for this table. Looks like we are aggregating data from portifolio snapshots and the last portifolio snapshot is not really the final competition results, so this table seems useful.

brunocalza:
> > I think so, if we are certain about it. Today in the call looked like there were some doubts about the need for this table. Looks like we are aggregating data from portifolio snapshots and the last portifolio snapshot is not really the final competition results, so this table seems useful.
> 
> @brunocalza ah, right. i agree that we _do_ we want `competitions_leaderboard` to track per-competition rankings vs computing it on the fly. the question is:
> 
> 1. should we have it _only_ get triggered at the end of the competition? (as described in [Refactor to store competitions final leaderboard in the database #546](https://github.com/recallnet/js-recall/issues/546))
> 2. do we want to include any data for in-flight competition rankings to also alleviate on-the-fly queries? or, only write to this table at the end of a competition?
>    
>    * e.g., we'd need some service that creates snapshots of the portfolio snapshots at some fixed interval (e.g., every 30 mins).
> 
> irrespective of (2), i think it's a required feature. up to you if you wanna include it now or in future the issue above

I see. I guess we can start with 1 for launch, which is pretty simple, then reevaluate 2 after launch? 2 sounds a lot more complex to be done right now, although I agree it will eventually be needed if things scale. Unless we already know the query is very slow

dtbuchholz:
I'm aligned with that approach. Makes sense to me.

brunocalza:
added `competitions_leaderboard` as well

dtbuchholz:
one thing to keep in mind: many of the existing APIs include both `score` and `rank` (as you did the in leaderboard table). with the new impl, it's possible we only need one of these fields?

Issue #521, opened 2025-06-06
Update `/competitions/:id` & agents endpoints with votes


dtbuchholz:
- `/competitions/:id`
   - include `totalVotes` in `stats` field (it's currently top-level instead of alongside the other `stats`). or, we can consider moving the `stats` up a level, if we thing the deserve to be there.
- `/agents/:id` and `/user/agents/:id`
   - `totalVotes` across all competitions
   - `bestPlacement`: array of objects with fields `position` and `total` (e.g., so you can say 1st out of 1000 agent). this might depend on #506; it can be an empty array, for now.
   - `rank` or `score` etc: need to align on naming convention, but these would use the new ranks in #221. the values can be placeholders

Issue #524, opened 2025-06-06
Strongly type Express responses and unify types


dtbuchholz:
**Problem Statement**
- The types across APIs are inconsistent. E.g., calling `/agents/:agentId` vs `/competitions/:compId/agents` vs `/agent/profile` all result in slightly different response values. This could be a "good" thing, but it's hard to sift through the response types and what should or shouldn't be included.
- This is further exacerbated by the fact Drizzle types are defined as e.g. `walletAddress: string | null` instead of ` ... | undefined`. Idiomatic TypeScript uses `undefined` and not `null`, so we wind up [creating our own types](https://github.com/recallnet/js-recall/blob/main/apps/api/src/types/index.ts) to better fit to ideal API responses. 
- Lastly, our internal tests and `comps` app are both maintaining their own, custom axios API clients ([here](https://github.com/recallnet/js-recall/blob/main/apps/comps/lib/api-client.ts) / [here](https://github.com/recallnet/js-recall/blob/main/apps/api/e2e/utils/api-client.ts)). These should integrate with an single source of truth like the Speakeasy autgen SDK, but our internal dev patterns have made this challenging. More on this below.

**Proposed Solution**
1. First, we probably need the spec-first development pattern (#392). This will ensure that we're creating custom types and will help resolve the problems with the first above.
2. Second, we need validate that Speakeasy's autgen SDK's types (`packages/api-sdk`) will work for our setup. One of the main benefits of this SDK is to address the Express server response values. In practice, it's not clear how easy it will be to use Speakeasy in our application code because the autogen naming conventions are long/messy, and without the spec-first dev pattern, there's a circular dependency issue (since the spec lives in the Express API app). We sorta need (a) to happen to really know what we can do here.
3. Third, we need to unify the Drizzle responses to our application expectations. This means dealing with the `null` and `undefined` discrepancies to make it easy to easier to convert from a db query to the actual type and keep everything consistent.

For point (3), we could do something like this to convert `null` to undefined at the "repository" db query layer:
```ts
function normalizeNullToUndefined<T extends Record<string, any>>(obj: T): {
  [K in keyof T]: T[K] extends null ? undefined : T[K];
} {
  return Object.fromEntries(
    Object.entries(obj).map(([k, v]) => [k, v === null ? undefined : v]),
  ) as any;
}

// Usage
const dbResult = await db.query.users.findFirst();
const safeResult = normalizeNullToUndefined(dbResult);
```

Or, `zod` transformations could be another option:
[...]

PR #565, opened 2025-06-11
feat(api): include API agent votes & stats

### LINKED ISSUE #508: Add `totalVotes` to per-agent `/leaderboard` response
**Problem Statement**
The UI needs to tally up the # of votes per agent and display that on the leaderboard. i.e., the cumulative sum per `agentId` in the new `votes` table.

**Proposed Solution**
Include `totalVotes` in the `/leaderboard` response for each agent. It's just a query + sum by agent ID in the `votes` table.

**Impact**
frontend/backend

**Priority**
highest. it's a small feature but should be part of the new voting UIs

dtbuchholz:
closes #508 / #521

### DIFF
--- a/apps/api/e2e/tests/agent.test.ts
+++ b/apps/api/e2e/tests/agent.test.ts
@@ -932,6 +932,13 @@ describe("Agent API", () => {
     expect(Array.isArray(agentData.agent.trophies)).toBe(true);
     expect(agentData.agent.hasUnclaimedRewards).toBe(false);
     expect(agentData.agent.stats).toBeDefined();
+    expect(agentData.agent.stats?.completedCompetitions).toBe(0);
+    expect(agentData.agent.stats?.totalTrades).toBe(0);
+    expect(agentData.agent.stats?.totalVotes).toBe(0);
+    // TODO: once `agent_rank` + `competitions_leaderboard` are implemented, these should be defined
+    expect(agentData.agent.stats?.bestPlacement).toBeUndefined();
+    expect(agentData.agent.stats?.rank).toBeUndefined();
+    expect(agentData.agent.stats?.score).toBeUndefined();
     // Validate owner information is included
     expect(agentData.owner).toBeDefined();
     expect(agentData.owner.id).toBeDefined();
--- a/apps/api/src/controllers/agent.controller.ts
+++ b/apps/api/src/controllers/agent.controller.ts
@@ -205,7 +205,7 @@ export function makeAgentController(services: ServiceRegistry) {
 
         const sanitizedAgent = services.agentManager.sanitizeAgent(agent);
         const computedAgent =
-          services.agentManager.attachAgentMetrics(sanitizedAgent);
+          await services.agentManager.attachAgentMetrics(sanitizedAgent);
 
         // Return the agent with owner information
         res.status(200).json({
--- a/apps/api/src/controllers/user.controller.ts
+++ b/apps/api/src/controllers/user.controller.ts
@@ -172,12 +172,16 @@ export function makeUserController(services: ServiceRegistry) {
         );
 
         // Remove sensitive fields, but add back the email and deactivation since the user should see them
-        const sanitizedAgents = agents.map((agent) => ({
-          ...services.agentManager.sanitizeAgent(agent),
-          email: agent.email,
-          deactivationReason: agent.deactivationReason,
-          deactivationDate: agent.deactivationDate,
-        }));
+        const sanitizedAgents = await Promise.all(
+          agents.map(async (agent) => ({
+            ...(await services.agentManager.attachAgentMetrics(
+              services.agentManager.sanitizeAgent(agent),
+            )),
+            email: agent.email,
+            deactivationReason: agent.deactivationReason,
+            deactivationDate: agent.deactivationDate,
+          })),
+        );
 
         res.status(200).json({
           success: true,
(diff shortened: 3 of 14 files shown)


apps/api/src/controllers/agent.controller.ts
```diff
@@ -205,7 +205,7 @@ export function makeAgentController(services: ServiceRegistry) {
 
         const sanitizedAgent = services.agentManager.sanitizeAgent(agent);
         const computedAgent =
-          services.agentManager.attachAgentMetrics(sanitizedAgent);
+          await services.agentManager.attachAgentMetrics(sanitizedAgent);
```
joewagner, line 208: 👍

apps/api/src/controllers/user.controller.ts
```diff
@@ -172,12 +172,16 @@ export function makeUserController(services: ServiceRegistry) {
         );
 
         // Remove sensitive fields, but add back the email and deactivation since the user should see them
-        const sanitizedAgents = agents.map((agent) => ({
-          ...services.agentManager.sanitizeAgent(agent),
-          email: agent.email,
-          deactivationReason: agent.deactivationReason,
-          deactivationDate: agent.deactivationDate,
-        }));
+        const sanitizedAgents = await Promise.all(
```
joewagner, line 175: 👍

apps/api/e2e/tests/agent.test.ts
```diff
@@ -932,6 +934,12 @@ describe("Agent API", () => {
     expect(Array.isArray(agentData.agent.trophies)).toBe(true);
     expect(agentData.agent.hasUnclaimedRewards).toBe(false);
     expect(agentData.agent.stats).toBeDefined();
+    expect(agentData.agent.stats?.completedCompetitions).toBe(0);
+    expect(agentData.agent.stats?.totalTrades).toBe(0);
```
dtbuchholz, line 938: this adds "summary" metrics for an agent's competitions, trades, and votes to these agent endpoints: `/user/agents`, `/user/agents/:id`, and `/agents/:id`

a follow up was created in #566 to handle the `bestPlacement`, `score`, and `rank` since these depend on some of the new tables (`agent_rank`, `competitions_leaderboard`) that are a WIP.

apps/api/openapi/openapi.json
```diff
@@ -4144,6 +4164,10 @@
                               "type": "number",
                               "description": "Total volume of trades in USD"
                             },
+                            "totalVotes": {
```
dtbuchholz, line 4167: note: this was previously a level above `stats`. since `stats` already has similar metrics, it's been nested accordingly

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -527,3 +529,40 @@ export async function findByStatus({
     throw error;
   }
 }
+
+/**
+ * Find the best placement of an agent across all competitions
+ * @param agentId The agent ID
+ * @returns The agent best placement
+ */
+export async function findBestPlacementForAgent(agentId: string) {
```
dtbuchholz, line 538: this is where we use the new `competitions_leaderboard` table. this new table doesnt have any logic that writes to it, yet, so the tests in the PR operate with this assumption.

the `findBestPlacementForAgent` will query the `competitions_leaderboard` table for the top "rank" by agent ID. then, it gets the total # of agents that were in that competitions so that we can show the agent placed _N_ of _M_.

apps/api/src/routes/agents.routes.ts
```diff
@@ -168,25 +168,37 @@ export function configureAgentsRoutes(
...
-   *                           type: array
-   *                           items:
-   *                             type: string
    *                         bestPlacement:
    *                           type: object
    *                           properties:
    *                             competitionId:
    *                               type: string
-   *                             postition:
+   *                             rank:
```
dtbuchholz, line 182: just calling out that the `bestPlacement` object's field were altered to better align with existing terminology. (i.e., the `comps` app API client / types will need to adjust for this.)

apps/api/src/services/agent-manager.service.ts
```diff
@@ -958,11 +986,28 @@ export class AgentManager {
    * @param sanitizedAgent The sanitized agent object
    * @returns The agent object with metrics attached
    */
-  attachAgentMetrics(
+  async attachAgentMetrics(
     sanitizedAgent: ReturnType<AgentManager["sanitizeAgent"]>,
   ) {
     const metadata = sanitizedAgent.metadata as AgentMetadata;
-    const stats = metadata?.stats || {};
+    const completedCompetitions = await countAgentCompetitionsForStatus(
```
dtbuchholz, line 993: the `completedCompetitions`, `totalVotes`, `totalTrades`, and `bestPlacement` have all been added to the response. `bestPlacement` will always be undefined until the new logic mentioned above is implemented.

the follow up issue for `rank` and `score` should address populating the remaining values.

joewagner (approved):
🚀

apps/api/src/types/index.ts
```diff
@@ -187,11 +187,34 @@ export interface AdminMetadata {
...
 }
 
+/**
+ * Agent's stats schema
+ */
+export const AgentStatsSchema = z.object({
+  completedCompetitions: z.number(),
+  totalTrades: z.number(),
+  totalVotes: z.number(),
+  // TODO: make these non-optional once we have the data
```
brunocalza, line 197: @dtbuchholz I think they would still be optional because until an agent compete and the competition finishes, these values are `undefined`, right?

dtbuchholz: @brunocalza hm yeah i suppose you're right. makes sense

joewagner: $0.02 here.  maybe they can be `nullable` instead of optional. https://zod.dev/api?id=nullables
`nullable` maps closer to postgres' use of null if there is no value.

dtbuchholz: yeah i'm not sure what the best approach is, but it's something we need to resolve wrt API types etc #524

the code is laden with duplicates due to null vs undefined definitions caused by drizzle (uses `null`) and typical TS optional types (`undefined`). e.g., should we transform drizzle nulls to be interpreted as undefined to better align with TS optionals? or do the inverse, like what you proposed?

joewagner: Zod also has `nullish`, which can be either null or undefined.  For now, I guess just leave the TODO

PR #567, opened 2025-06-11
feat: Add object index table

islamaliev:
This PR adds the database schema required for the Recall Data Synchronizer to function. The synchronizer reads from the object_index table to identify S3 objects that need to be synced to the Recall Testnet. 

The schema follows existing js-recall conventions and includes proper indexes for efficient querying by the synchronizer. Foreign key relationships are established with the competitions and agents tables to maintain data integrity.

### DIFF
--- a/apps/api/drizzle/0009_lush_the_hunter.sql
+++ b/apps/api/drizzle/0009_lush_the_hunter.sql
@@ -11,12 +11,12 @@ CREATE TABLE "object_index" (
 	"event_timestamp" timestamp with time zone,
 	"object_last_modified_at" timestamp with time zone NOT NULL,
 	"created_at" timestamp with time zone DEFAULT now() NOT NULL,
-	"updated_at" timestamp with time zone DEFAULT now() NOT NULL
+	"updated_at" timestamp with time zone DEFAULT now() NOT NULL,
+	CONSTRAINT "object_index_object_key_key" UNIQUE("object_key")
 );
 --> statement-breakpoint
 ALTER TABLE "object_index" ADD CONSTRAINT "object_index_competition_id_fkey" FOREIGN KEY ("competition_id") REFERENCES "public"."competitions"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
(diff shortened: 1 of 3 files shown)


apps/api/drizzle/0008_add_object_index.sql
dtbuchholz, line 1: @islamaliev the `drizzle/*` files are autogenerated with the following command:
```
pnpm db:gen-migrations
```
to fix the CI/e2e tests issue: delete this file, merge `main` into this branch (it has some new drizzle/schema changes), and then run the command.

or, you can merge my PR into this branch, if you want: https://github.com/recallnet/js-recall/pull/568

apps/api/src/database/schema/syncing/defs.ts
```diff
@@ -0,0 +1,68 @@
...
+ */
+export const objectIndex = pgTable(
+  "object_index",
+  {
+    id: uuid().primaryKey().notNull(),
+    objectKey: text("object_key").notNull(),
+    bucketName: varchar("bucket_name", { length: 255 }).notNull(),
+    competitionId: uuid("competition_id").notNull(),
+    agentId: uuid("agent_id").notNull(),
+    dataType: varchar("data_type", { length: 100 }).notNull(),
```
dtbuchholz, line 27: note: something we'll want to align on is what this field is. it'll probably need to be a `pgEnum` that includes predefined types: `trade | ...` etc. 

since all trades _include_ reasoning, we'll want to figure out if we want to split every trade into 2 objects (one for the trade, and another for the `trade.reasoning` field), or just store them as one object.

islamaliev: You want me to change it now? I would need all predefined types for this.

Or I can merge this one and we will see what to do with later.

dtbuchholz: > Or I can merge this one and we will see what to do with later.

yeah that's what i was thinking. we'll just want to make sure we define the enum before we write any data (to make the migration easier)

Issue #472, opened 2025-06-02
Implement new Competitions Hub design


ramirotw:
**Problem Statement**
The current Competitions page was based on the old designs. Implement the new design as described in  https://github.com/recallnet/js-recall/issues/409.

**Priority**
High

This ticket is for the frontend implementation.  UI Design is complete.  Backend changes will be handled in  https://github.com/recallnet/js-recall/issues/497

PR #570, opened 2025-06-11
feat(comps): add competitions hub

ramirotw:
Fixes https://github.com/recallnet/js-recall/issues/472

localhost_3001_competitions


apps/comps/components/featured-competition/top-leaders-list.tsx
```diff
@@ -0,0 +1,84 @@
...
+        <div
+          key={agent.id}
+          className="hover:bg-card col-span-4 grid grid-cols-[70px_1fr_auto_auto] gap-4"
+        >
+          <RankBadge position={agent.position} />
+          <Link
+            href={`/agents/${agent.id}`}
+            className="flex items-center gap-3"
+          >
+            <AgentAvatar agent={agent} size={32} />
```
dtbuchholz, line 60: this is causing a hydration error: `In HTML, <a> cannot be a descendant of <a>.
This will cause a hydration error.`
Screenshot 2025-06-12 at 3 07 29 PM

dtbuchholz:
### FeaturedCompetition
 
- the `Reward`'s left margin/padding isnt consistent with the `Duration`
- we should include only the top 3 agents in the leaderboard (bottom right), which will might help prevent the text under `Join. Vote. Win.` getting cut off on standard laptop viewports.
- also, it looks like the backend might have an issue with the `sort` param here? because it's not showing the top 3 agents
- maybe the video should use some `radial-gradient` css to match the design? idk, we can keep things as-is. it's hard to tell if that's what the design expects vs what the mocked video looks like.
Screenshot 2025-06-12 at 2 39 55 PM

### Join competition

- if i click "join" and select an agent, i can choose one that's already in that competition. thus, we see the error modal pop up. 
   - i _dont think_ the designs addressed this, so we can probably keep it as-is. e.g., when you join, the agent dropdown could show your agents but disable+tooltip if one is already in that competition. cc @jovirecall 
Screenshot 2025-06-12 at 2 48 37 PM

dtbuchholz:
@ramirotw looking good! a couple of small tweaks but lmk if you saw a similar API issue on the featured comp leaderboard, and i can dig into it

apps/comps/components/featured-competition/top-leaders-list.tsx
ramirotw: fixed in [9ed2821](https://github.com/recallnet/js-recall/pull/570/commits/9ed28212509c2fcb81ea4d84635a74b302c8a86b)

ramirotw:
> it looks like the backend might have an issue with the sort param here?

yes, sorting for `/api/competitions/{competitionId}/agents` seems quite basic, no sorting based on rank/position

https://github.com/recallnet/js-recall/blob/main/apps/api/src/database/repositories/agent-repository.ts#L271-L291

joewagner:
> > it looks like the backend might have an issue with the sort param here?
> 
> yes, sorting for `/api/competitions/{competitionId}/agents` seems quite basic, no sorting based on rank/position
> 
> https://github.com/recallnet/js-recall/blob/main/apps/api/src/database/repositories/agent-repository.ts#L271-L291

This is kind of confusing, but a good callout.  The current value is not stored in the database, so it can't be sorted via the code you're linking to.  The api does sort after the fact, which is not ideal. IIUC, it does require that you request all agents, i.e. limit >= number of competitors. If you set limit to somthing like `50`, and sort to `position`, you should get the agents sorted by position. 

Code is here: 
https://github.com/recallnet/js-recall/blob/main/apps/api/src/controllers/competition.controller.ts#L756

Once the ranking database work is done we will need to re-visit this.

For now, we can just do this https://github.com/recallnet/js-recall/pull/608

PR #584, opened 2025-06-11
refactor: use agent rank for global leaderboard

### LINKED ISSUE #550: Update global leaderboard to use agent rank
There is already a `/leaderboard` API endpoint that is used to display the leaderboard.  This ticket will update that endpoint to pull data from the new ` agent_rank` table.  No frontend changes should be necessary

brunocalza:
Closes #550


apps/api/src/services/leaderboard.service.ts
```diff
@@ -83,145 +75,31 @@ export class LeaderboardService {
 
   /**
    * Calculate global metrics for all agents across multiple competitions
-   * @param competitionIds List of competition IDs to process
    * @returns Map of agent ID to accumulated metrics
    */
-  private async calculateGlobalMetrics(
-    competitionIds: string[],
-  ): Promise<Map<string, LeaderboardAgent>> {
+  private async calculateGlobalMetrics(): Promise<
```
brunocalza, line 80: If it helps, better looking at the final code instead of the diff

```typescript
  /**
   * Calculate global metrics for all agents across multiple competitions
   * @returns Map of agent ID to accumulated metrics
   */
  private async calculateGlobalMetrics(): Promise<
    Map<string, LeaderboardAgent>
  > {
    const agentMetricsMap = new Map<string, LeaderboardAgent>();

    // TODO: this is not scalable. It should be paginated.
    const agentRanks = await getAllAgentRanks();

    for (const { id, name, score } of agentRanks) {
      const voteCount = await countTotalVotesByAgent(id);
      const numCompetitions = await countAgentCompetitions(id);

      agentMetricsMap.set(id, {
        rank: 0,
        id: id,
        name: name,
        score: score,
        numCompetitions: numCompetitions,
        voteCount: voteCount,
      });
    }

    return agentMetricsMap;
  }
 ```

File: apps/api/src/database/schema/ranking/defs.ts [hunk 2]
```diff
@@ -17,12 +18,9 @@ export const agentRank = pgTable(
   {
     id: uuid().primaryKey().notNull(),
     agentId: uuid("agent_id").notNull(),
-    mu: numeric("mu", { precision: 6, scale: 2, mode: "number" }).notNull(),
-    sigma: numeric("sigma", {
-      precision: 6,
-      scale: 2,
-      mode: "number",
-    }).notNull(),
+    mu: doublePrecision("mu").notNull(),
```
brunocalza, line 21: - Changed to **double precision** which is more aligned with what's used in OpenSkill library and we don't have to worry about decimal places, we store whatever the library give to us.

- Added ordinal based on https://recalllabs.slack.com/archives/C08TEDYB5E1/p1749670458666889

apps/api/src/database/repositories/leaderboard-repository.ts
```diff
@@ -50,10 +34,7 @@ export async function getGlobalStats(type: CompetitionType): Promise<{
     .where(
       and(
         eq(competitions.type, type),
-        inArray(competitions.status, [
-          COMPETITION_STATUS.ACTIVE,
-          COMPETITION_STATUS.ENDED,
-        ]),
+        eq(competitions.status, COMPETITION_STATUS.ENDED),
```
brunocalza, line 37: We grab data only for competition that ended. See https://recalllabs.slack.com/archives/C08TEDYB5E1/p1749586581497879

apps/api/src/database/repositories/leaderboard-repository.ts
```diff
@@ -99,7 +68,7 @@ export async function getGlobalStats(type: CompetitionType): Promise<{
     .where(inArray(votes.competitionId, relevantCompetitionIds));
 
   return {
-    activeAgents: activeAgentsResult[0]?.count ?? 0,
```
brunocalza, line 102: Since active agents only make sense on ongoing competitions, this information does not make sense now. Keeping it 0 so not to break any clients

.github/workflows/api-ci.yml
```diff
@@ -47,6 +47,21 @@ jobs:
       - name: Build everything
         run: pnpm build
 
+      - name: Check for uncommitted database migrations
```
brunocalza, line 50: added this check to avoid this problem: https://recalllabs.slack.com/archives/C08TV9K3XS5/p1749678705705319

apps/api/e2e/tests/leaderboard.test.ts
```diff
@@ -80,12 +81,35 @@ describe("Leaderboard API", () => {
       reason: "Test trade",
     });
 
+    await adminClient.endCompetition(startResponse.competition.id);
```
brunocalza, line 84: All leaderboard tests were refactored along these lines:
- all competitions are finished (because data from in-flight competitions don't make it to the global leaderboard anymore)
- a fake update on agent rank ( this will be removed later when we have the agent rank update logic)
- the check for active agents

dtbuchholz (approved):
🤘

Issue #579, opened 2025-06-11
[FEATURE] Add `isVerified` field to per-agent response


dtbuchholz:
**Problem Statement**
An agent is "verified" if it proved ownership of a wallet via the `/auth/verify` API call. This will populate the `walletAddress` field, allowing you to infer verification happened. 

**Proposed Solution**
We should add an `isVerified` field to make it semantically obvious. 

**Impact**
Makes it easer to work with the APIs and understand an agent's current state.

**Priority**
High.

dtbuchholz:
Unrelated. 

This is about the verified status of an agent: have they proved ownership of a _wallet_? If so, the response includes an `isVerified` field to make their "verified" wallet ownership explicit in API responses. e.g., comps app uses it when rendering the agent profile.

dtbuchholz:
Although, maybe there's some overlap with what's described in https://github.com/recallnet/js-recall/issues/416.

If an agent is verified (has a wallet), that can give us info about the owning user...and marketing emails would target that user differently that a user that has zero verified agents.

stbrody:
got it.  Do we need to introduce some different words here to distinguish the different forms of verification?  ie wallet vs email verification?

dtbuchholz:
`isVerified` is just an API field (not db column), so it's something we can easily change if needed. 

fwiw, i dont think `isWalletVerified` and `isEmailVerified` will be db fields for the same reason as above. if a user or agent db row has a wallet address, their wallet has been verified; that's the only way it gets written to the table. 

it's not clear what email verification will be, so we can figure that out once it happens. e.g., SSO verifies => creates db entry => no need to track `isEmailVerified`

Issue #392, opened 2025-05-22
Move to spec-first development pattern


dtbuchholz:
**Problem Statement**
We’re currently developing our OpenAPI spec metadata and API server implementation in tandem, and then generating a final OpenAPI spec file. This approach is okay, but focuses on implementation first. If we can use the spec as the place where we collaborate to encode API design thinking and decisions, that information will filter down to client and server implementations and database schema design. 

**Proposed Solution**
1. Move the OpenAPI spec to new package, like `packages/api-spec`
2. Use some tool like [`openapi-generator-cli` ](https://openapi-generator.tech/docs/installation/) to generate TS requirements
3. Let `apps/api` and `packages/api-sdk` use this dep to drive logic

**Impact**
- Improved engineering practices
- E.g., now that we're going from designs<>frontend<>backend, it'll be easier to design API/table changes when the spec is isolated & leads vs follows

**Priority**
Medium? Features are more important, right now.

stbrody:
Want to do this soon after the app launch

PR #586, opened 2025-06-12
feat(api): add isVerified to agent response object

joewagner:
## Overview

This adds a calculated field to the agent response object called `isVerified`.  

## Details

The logic to determine that an agent is verified is simply, if they have a walletAddress they are verified.  This works because we now require users to sign a message and prove they control the wallet address of an agent.  It is worth noting that an admin can still add a wallet address without a signed message.  If this happens the agent is considered to be verified,

The PR also includes some de-duplication of the Agent schema in the openapi spec.

fixes https://github.com/recallnet/js-recall/issues/579

### DIFF
--- a/apps/api/src/routes/agents.routes.ts
+++ b/apps/api/src/routes/agents.routes.ts
@@ -93,6 +93,8 @@ export function configureAgentsRoutes(
    *                       walletAddress:
    *                         type: string
    *                         example: "0x1234567890abcdef1234567890abcdef12345678"
+   *                       isVerified:
+   *                         type: boolean
    *                       name:
    *                         type: string
    *                         example: "Trading Bot Alpha"
@@ -155,6 +157,8 @@ export function configureAgentsRoutes(
    *                     name:
    *                       type: string
    *                       example: "Trading Bot Alpha"
+   *                     isVerified:
+   *                       type: boolean
    *                     imageUrl:
    *                       type: string
    *                       example: "https://example.com/bot-avatar.jpg"
--- a/apps/api/openapi/openapi.json
+++ b/apps/api/openapi/openapi.json
@@ -2024,6 +2024,9 @@
                           "type": "string",
                           "example": "0x1234567890abcdef1234567890abcdef12345678"
                         },
+                        "isVerified": {
+                          "type": "boolean"
+                        },
                         "name": {
                           "type": "string",
                           "example": "Trading Bot Alpha"
@@ -2171,6 +2174,9 @@
                         },
                         "walletAddress": {
                           "type": "string"
+                        },
+                        "isVerified": {
+                          "type": "boolean"
                         },
                         "name": {
                           "type": "string"
@@ -2681,6 +2687,9 @@
                             "type": "string",
                             "example": "0x1234567890abcdef1234567890abcdef12345678"
                           },
+                          "isVerified": {
+                            "type": "boolean"
+                          },
                           "name": {
                             "type": "string",
                             "example": "Trading Bot Alpha"
@@ -2769,6 +2778,9 @@
                         "name": {
                           "type": "string",
                           "example": "Trading Bot Alpha"
+                        },
+                        "isVerified": {
+                          "type": "boolean"
                         },
                         "imageUrl": {
                           "type": "string",
@@ -5940,6 +5952,9 @@
                             "type": "string",
                             "nullable": true
                           },
+                          "isVerified": {
+                            "type": "boolean"
+                          },
                           "name": {
                             "type": "string"
                           },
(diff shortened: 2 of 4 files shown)


apps/api/src/config/swagger.ts
```diff
@@ -185,6 +185,86 @@ For convenience, we provide an API client that handles authentication automatica
             },
           },
         },
+        Agent: {
```
joewagner: This is the Agent schema. We had declared this in several different places, and in some cases they were different.

apps/api/src/database/schema/syncing/defs.ts
```diff
@@ -65,4 +65,4 @@ export const objectIndex = pgTable(
       name: "object_index_agent_id_fkey",
     }).onDelete("cascade"),
   ],
-);
```
joewagner, line 68: This was a result of running `pnpm format` I'm not sure how it made it to main without end of file return

apps/api/src/routes/agents.routes.ts
```diff
@@ -82,35 +82,7 @@ export function configureAgentsRoutes(
...
-   *                       status:
-   *                         type: string
-   *                         enum: [active, suspended, deleted]
-   *                       createdAt:
-   *                         type: string
-   *                         format: date-time
-   *                       updatedAt:
-   *                         type: string
-   *                         format: date-time
+   *                     $ref: '#/components/schemas/Agent'
```
joewagner: You'll see all the different places an agent was spec'ed out now have a `$ref` to the one spec.

apps/api/src/types/index.ts
```diff
@@ -469,7 +469,9 @@ export type Agent = z.infer<typeof AgentSchema>;
 /**
  * Pulic Agent information Object, omits apiKey
  */
-export const AgentPublicSchema = AgentSchema.omit({ apiKey: true });
+export const AgentPublicSchema = AgentSchema.omit({ apiKey: true }).extend({
```
joewagner, line 473: We now have one type and one matching spec schema.  This is assuming we keep the spec up to date with the type.

apps/api/src/services/agent-manager.service.ts
```diff
@@ -983,7 +983,10 @@ export class AgentManager {
    * @returns The sanitized agent object
    */
   sanitizeAgent(agent: SelectAgent) {
-    return AgentPublicSchema.parse(agent);
+    return AgentPublicSchema.parse({
+      ...agent,
+      isVerified: !!agent.walletAddress,
```
joewagner, line 988: This is the only business logic change. In general as long as our controllers use the `sanitizeAgent` method, the responses will match the spec schema

apps/api/src/database/schema/syncing/defs.ts
dtbuchholz: hm odd. there are some chore-related repo things we need to look at...e.g., `format` is a root-level command, and it isnt enforced in CI (or pre-commit hooks, if we want to get more restrictive).

apps/api/src/config/swagger.ts
dtbuchholz: nice! this is something we've been needing to clean up. the spec _and_ general types throughout the `api` app need to be more reusable and extensible.

a few callouts:
- this "base" `Agent` is missing: `isVerified`, `walletAddress`, `description`, `createdAt`, `updatedAt`, and `ownerId` fields. 
- there are some responses below that extend this and include the `deactivationReason` and `deactivationDate`. e.g., when a [user gets their agents](https://github.com/recallnet/js-recall/blob/a0c823a7d497c6ceeb9b97977c11bdd51ce23345/apps/api/src/controllers/user.controller.ts#L181), the response is technically different from the `Agent` schema. (imo, these "deactivation" fields aren't needed tho.)
- agent profile response also includes the `email`
- the agent creation API also includes the `apiKey`

to your point—it's confusing and isn't obvious if the spec incorrectly defined a response, or if the response has a one-off field that extends `Agent`. as you've outlined here, we should only need one `Agent` schema in (basically) all of these responses.

fwiw, it's clearly messy. i'll often see slight differences in the spec response vs. what i see when testing w/ cURL locally.

...maybe we can prioritize #392 / #524.

joewagner: Great points.  I didn't spend much time trying to make sure this was 100% correct since as you mention most of the endpoints aren't quite correct anyway, but I'll cleanup the base schema.  
As for the one-off responses, i.e. apiKey, deactivationDate, etc..., maybe we move them out of the agent object?

For example, the agent creation response could be { success: ..., agent: ..., apiKey: ... }

dtbuchholz: Agreed, good idea; moving them out of the base object makes sense to me.

joewagner:
@dtbuchholz I may have underestimated how much work is involved with cleaning up this part of the openapi spec.  I got the openapi schema to match the type, but I didn't get into the work of updating the actual response objects.
Maybe we can leave that work for #392 and #524?

PR #589, opened 2025-06-12
fix(api): fix openapi spec for /user/competitions

joewagner:
The spec for the `/api/user/competitions` endpoint did not have the available query filters listed.  This updates the spec, which in turn updates the swagger docs stie.

### DIFF
--- a/apps/api/src/services/index.ts
+++ b/apps/api/src/services/index.ts
@@ -71,9 +71,7 @@ class ServiceRegistry {
     this._voteManager = new VoteManager();
 
     // Initialize LeaderboardService with required dependencies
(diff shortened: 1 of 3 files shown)


apps/api/src/services/index.ts
```diff
@@ -71,9 +71,7 @@ class ServiceRegistry {
     this._voteManager = new VoteManager();
 
     // Initialize LeaderboardService with required dependencies
-    this._leaderboardService = new LeaderboardService(
-      this._agentManager,
-    );
+    this._leaderboardService = new LeaderboardService(this._agentManager);
```
joewagner, line 74: This is unrelated to the PR. It is an outcome of running format locally.

PR #592, opened 2025-06-12
api: allow for legacy data to be loaded pre-drizzle migration in docker setup, dev, and prod

mzkrasner:
# Baseline Migration & Production Database Improvements

## Summary
This PR introduces robust support for initializing production databases with a legacy baseline SQL file, followed by Drizzle migrations. It also ensures the Docker production environment applies both the baseline and all migrations automatically, and updates documentation for clarity.

## Key Changes
- **Baseline Migration System**: Ensures a large legacy SQL file is applied on fresh databases before running Drizzle migrations.
- **Production Script**: `pnpm db:prepare-production` now applies both the baseline and all migrations.
- **Docker Integration**: Docker startup script (`startup.sh`) uses `pnpm db:prepare-production` for first-time database setup, supporting both baseline and migrations.
- **Progress Feedback**: Adds terminal progress indicator for long-running baseline imports.
- **Documentation**: Updates to `README.md` and Docker `instructions.md` to clearly explain the production database preparation process and baseline system.

## How to Use
1. Place your legacy SQL file at `apps/api/baseline/baseline.sql`.
2. Run `pnpm db:prepare-production` (or start the Docker container) to initialize a fresh production database.
3. Both the baseline and all Drizzle migrations will be applied automatically.

## Why?
- Supports onboarding legacy data and schema in production.
- Ensures all environments (local, CI, Docker) are consistent and up-to-date.
- Provides clear, step-by-step documentation for future maintainers.

### DIFF
--- a/apps/api/src/database/db.ts
+++ b/apps/api/src/database/db.ts
@@ -11,8 +11,6 @@ import { fileURLToPath } from "url";
 import { config } from "@/config/index.js";
 import schema from "@/database/schema/index.js";
 
-import { applyBaselineIfNeeded } from "../../scripts/baseline.js";
-
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
@@ -135,9 +133,6 @@ export async function dropAll() {
 }
 
 export async function migrateDb() {
-  // Apply baseline if needed before running migrations
-  await applyBaselineIfNeeded();
-
   // Run normal Drizzle migrations
   await migrate(db, {
     migrationsFolder: path.join(__dirname, "../../drizzle"),


apps/api/src/database/baseline.ts
dtbuchholz, line 1: since the legacy db/baseline migration is a one-time thing, i think we want to keep this code isolated in the `scripts` folder? not a big deal tho. i just can't imagine we'd go through this process again in the future since drizzle migrations will handle it for us....unless you see another possible use case.

apps/api/src/database/db.ts
```diff
@@ -133,6 +135,10 @@ export async function dropAll() {
 }
 
 export async function migrateDb() {
```
dtbuchholz, line 135: this isn't needed, right?
1. it looks like `baseline.ts` will call `applyBaselineIfNeeded` before it calls this file's `migrateDb`
2. plus, i dont think we want to alter the core logic (even if it's small) for this one-time, manual migration

apps/api/trade-simulator-docker/startup.sh
```diff
@@ -16,9 +16,9 @@ done
 
 # Check if we need to run migrations
 if ! check_db_initialized; then
-    echo "Database not initialized, running migrations..."
-    cd /workdir/apps/api && pnpm db:migrate
-    echo "Migrations completed"
+    echo "Database not initialized, running baseline and migrations..."
+    cd /workdir/apps/api && pnpm db:prepare-production
```
dtbuchholz: and same thing here...i think the container should maintain whatever flow a "normal" production setup (non-legacy, manual migration) will use, so it should stay as `db:migrate`. 

or, we could set some env var like `RUN_BACKFILL=true`, and if that's set, it'll execute `pnpm:prepare-production ` ; else, `pnpm db:migrate`

apps/api/src/database/baseline.ts
mzkrasner: Yeah makes sense

apps/api/src/database/db.ts
mzkrasner: Ah yes - forgot to take that out when trying to ensure consistent behavior between prepping the production environment and using in dev

dtbuchholz: @mzkrasner we still wanna remove this from `db.ts`, right?

apps/api/src/database/db.ts
```diff
@@ -11,6 +11,8 @@ import { fileURLToPath } from "url";
 import { config } from "@/config/index.js";
 import schema from "@/database/schema/index.js";
 
+import { applyBaselineIfNeeded } from "../../scripts/baseline.js";
```
dtbuchholz: ^specifically, this and `await applyBaselineIfNeeded();` down below

mzkrasner: ah - apologies. Spending the past two days on data normalization seems to have fried my brain. Just ammended

dtbuchholz: lol i cant even imagine

Issue #539, opened 2025-06-09
[IMPLEMENTATION] Create a more explicit "Voting Window"


stbrody:
## Definition of Done:

- [x] Database schema update to add voting start and stop date columns to competition data
- [x] Update admin creation API endpoint to accept start and stop date, and create a new "update" endpoint
- [x] Update competitions-related endpoints with the new voting start/end date in the responses
- [x] Update vote endpoint to reject votes outside of voting window
- [ ] New banners to display when voting is open/closed
- [ ] New competition page countdown and "your vote" information; also, competition hub countdown

## Notes

- Figma designs: [here](https://www.figma.com/design/WqQqcsg1g66EdMjjM8yfIM/Competitions-App---v.1.0--06-25-?node-id=8-50549&t=zAmUz88x6Mn1LYVN-1)
- The voting window is flexible, so there is no hardcoded / programmatic rule; it is per-competition.
- Admins need to be able to set the voting window the create/update competition endpoints to set these fields.
- All relevant competition endpoints will return these new fields in API responses.
- We'll need to make some new APIs and alter underlying voting logic.

### Backend:

**Database changes:**

- Alter the [`competitions` table](https://github.com/recallnet/js-recall/blob/aaf4716f54f54e7928d44ae798bbbb6dbdee25e8/apps/api/src/database/schema/core/defs.ts#L161) to include two new columns: `vote_start_date`, `vote_end_date`
- We'll need to adjust our logic for "can i vote on this competition?" e.g., [here](https://github.com/recallnet/js-recall/blob/aaf4716f54f54e7928d44ae798bbbb6dbdee25e8/apps/api/src/services/vote-manager.service.ts#L308)

**API changes:**
- `/admin/competition/create`: Update to include the (optional) voting start/end date.
- `/admin/competition/update`: We need to be able to update an existing competition that doesn't have vote windows.
- We'll need to ensure all of these include the new fields: `/competitions/*` (where relevant), `/agents/{agentId}/competitions`, `/user/competitions`
- The `/vote` endpoints should handle the proper rejections

### Frontend: 

**Relevant pages**:
   - Competitions page: individual competitions pages will need to display the voting countdown and banner
   - Competitions Hub: each competition in the hub needs to show the voting countdown (see [original designs](https://www.figma.com/design/WqQqcsg1g66EdMjjM8yfIM/Competitions-App---v.1.0--06-25-?node-id=3-14043&m=dev); different from figma above)
   - Page root: the banner should be on every page)

[...]

PR #593, opened 2025-06-12
feat(api): add voting window dates backend

joewagner:
part of: https://github.com/recallnet/js-recall/issues/539

This is the backend part of implementing the Voting window work.  I'm opening for review while I'm working on the frontend in a branch that is based off the tip of this branch.


apps/api/e2e/utils/server.ts
```diff
@@ -31,17 +31,21 @@ export async function startServer(): Promise<Server> {
...
-      serverProcess = spawn("npx", ["tsx", "src/index.ts"], {
-        env: {
-          ...process.env,
-          NODE_ENV: "test",
-          PORT: testPort,
-          HOST: testHost,
-          TEST_MODE: "true",
+      serverProcess = spawn(
+        "npx",
+        ["tsx", /*"--inspect-brk",*/ "src/index.ts"],
```
dtbuchholz: leftover comment?

apps/api/e2e/utils/server.ts
```diff
@@ -66,7 +70,7 @@ export async function startServer(): Promise<Server> {
       });
 
       // Wait for the server to be ready
-      waitForServerReady(30, 500) // 30 retries, 500ms interval = 15 seconds max
+      waitForServerReady(30_000, 500) // 30 retries, 500ms interval = 15 seconds max
```
dtbuchholz: this should be `30` retries, right?

apps/api/src/controllers/vote.controller.ts
```diff
@@ -63,12 +63,16 @@ export function makeVoteController(services: ServiceRegistry) {
           if (voteError instanceof Error && "type" in voteError) {
             const error = voteError as VoteError;
             switch (error.type) {
+              // TODO: it's a bit cumbersome to have to keep all these cases in
```
dtbuchholz, line 66: good idea https://github.com/recallnet/js-recall/issues/632

apps/api/src/services/vote-manager.service.ts
```diff
@@ -177,15 +177,40 @@ export class VoteManager {
       }
 
       // Check if voting is allowed based on competition status
-      const votingAllowed = this.checkCompetitionVotingStatus(competition);
-      if (!votingAllowed) {
+      const votingEnabledStatuses: Array<SelectCompetition["status"]> = [
```
dtbuchholz: since this and the method below both reuse this logic, should we keep the `this.checkCompetitionVotingStatus(...)` method usage? the method returns a boolean, which can be used in the subsequent `if` condition.

(optional: maybe we rename `checkCompetitionVotingStatus` -> `isVotingAllowedByStatus`)

apps/api/src/services/vote-manager.service.ts
```diff
@@ -276,9 +301,13 @@ export class VoteManager {
       throw error;
     }
 
-    // Check if competition allows voting based on status
-    const votingAllowed = this.checkCompetitionVotingStatus(competition);
-    if (!votingAllowed) {
+    // Check if competition status allows voting
```
dtbuchholz, line 284: here's the other place to keep the `checkCompetitionVotingStatus` method call

apps/api/src/services/vote-manager.service.ts
```diff
@@ -287,6 +316,29 @@ export class VoteManager {
       throw error;
     }
 
+    // Check voting dates if they are set
+    const now = new Date();
+
+    // If voting start date is set and we haven't reached it, voting is not allowed
+    if (competition.votingStartDate && now < competition.votingStartDate) {
```
dtbuchholz: optional: since both methods make a similar date check, it might be helpful for a helper like `isVotingAllowedByDates`...not essential tho.

dtbuchholz:
@joewagner left a couple of minor comments, but the backend stuff is lookin good! that drizzle-zod will def make things easier

joewagner:
> @joewagner left a couple of minor comments, but the backend stuff is lookin good! that drizzle-zod will def make things easier

Awesome! Good catches, I'll update this with the suggestions.

dtbuchholz (approved):
🤘

PR #594, opened 2025-06-13
Rafael/global leaderboard

Rcontre360:


apps/comps/types/agent.ts
```diff
@@ -22,6 +22,10 @@ export interface Agent {
   imageUrl: string;
   description?: string;
   status: string;
+  stats: {
```
dtbuchholz: note: the `bestPlacement` response shape has slightly changed ([here](https://github.com/recallnet/js-recall/blob/59f1d59f7d99a46781c74414e2c4a6c38630ce6a/apps/api/openapi/openapi.json#L2812))
- position => rank
- participants => totalAgents
- and there's also a `score` and `competitionId` field

apps/comps/components/leaderboard-table.tsx
```diff
@@ -21,32 +20,55 @@ import { LeaderboardAgent } from "@/types/agent";
...
-  onExtend: () => void;
   loaded?: boolean;
+
+  handleSortChange: (field: string) => void;
+  sortState: Record<string, SortState>;
+  onLoadMore: () => void;
+  total?: number;
 }) {
-  const toRender = agents;
+  const hasMore = total > (agents?.length || 0);
```
dtbuchholz: is there a reason we don't use the API's response `pagination.hasMore`?
```json
{
  "success": true,
  "stats": {
    "activeAgents": 20,
    "totalCompetitions": 3,
    "totalTrades": 26,
    "totalVolume": 2003.19816,
    "totalVotes": 1
  },
  "agents": [],
  "pagination": {
    "total": 20,
    "limit": 50,
    "offset": 0,
    "hasMore": false
  }
}

dtbuchholz:
@Rcontre360 all of the following column sort fields are available via the API:
- `rank` or `-rank` => the "Rank" column
- `name` or `-name` => the "Agent" column
- `competitions` or `-competitions` => the "Competitions" column
- `votes` or `-votes` => the "Votes" column

but, it doesn't look like they're working in the UI. is there a specific reason why? (i know the spec is a bit messy and it's hard to validate which fields should be used.)

we're going to review/clean up sorting logic via #637, but i think we can implement all of these now.

dtbuchholz (approved):
we can address the issues with "show more" in a follow up https://github.com/recallnet/js-recall/issues/640

PR #595, opened 2025-06-13
feat: Add 'data' column to 'object_index' table

islamaliev:


dtbuchholz:
> @dtbuchholz it's supposed to hold the data generated by agents. Whatever we planned to store in S3. You think it should have some other type?

@islamaliev gotcha. nah i think what you have is fine...i just wanted to double check what it's used for. i suppose `text` is more flexible than `jsonb`, in case we want to write data other than objects

PR #629, opened 2025-06-16
fix (api): correct pagination and sorting in /user/competitions endpoint

mzkrasner:
# Summary

### Issue Overview
Fixed bugs in `/user/competitions` endpoint where:
1. **Sorting threw 400 errors** when using valid competition fields (name, startDate, etc.)
2. **Pagination metadata was incorrect** - `total` and `hasMore` values didn't match actual returned data

### Root Cause
- **Layer violation**: Competition sort parameters were incorrectly passed to agent repository layer
- **Pagination mismatch**: Database pagination applied to raw JOIN results, but application layer performed deduplication, causing count inconsistencies

### Solution Implemented
**Two-Query Repository Pattern** with proper layer separation:

1. **Fixed Sort Format & Validation** (`helpers.ts`)
   - Standardized sort format: `"field"` (asc), `"-field"` (desc)
   - Added early validation with clear error messages

2. **Fixed Layer Separation** (`agent-manager.service.ts`)
   - Removed competition sort from agent queries
   - Added service-layer validation for competition sort fields

3. **Fixed Pagination Logic** (`agent-repository.ts`)
   - **Step 1**: Get unique competition IDs with proper sorting/pagination
   - **Step 2**: Fetch full data for those specific competitions
   - **Step 3**: Count total unique competitions for accurate metadata
   - Preserves sort order using SQL CASE statements

### Files Changed
- `apps/api/src/database/repositories/helpers.ts` - Sort format standardization
- `apps/api/src/database/repositories/agent-repository.ts` - Two-query pagination pattern
- `apps/api/src/services/agent-manager.service.ts` - Layer separation and validation

**Tests**: All existing tests pass, comprehensive test coverage validates the fixes.

### DIFF
--- a/apps/api/src/database/repositories/helpers.ts
+++ b/apps/api/src/database/repositories/helpers.ts
@@ -3,6 +3,15 @@ import { PgSelect } from "drizzle-orm/pg-core";
 
 import { ApiError } from "@/middleware/errorHandler.js";
 
+/**
+ * Apply sorting to a Drizzle query based on sort string
+ * @param query The Drizzle query to apply sorting to
+ * @param sortString Comma-separated sort fields with optional direction.
+ *                   Use "-field" for descending, "field" for ascending.
+ *                   Examples: "name", "-createdAt", "name,-createdAt"
+ * @param orderByOptions Available fields that can be sorted by
+ * @returns The query with sorting applied
+ */
 export function getSort<T extends PgSelect>(
   query: T,
   sortString: string,
@@ -20,22 +29,28 @@ export function getSort<T extends PgSelect>(
     );
   }
 
+  // Collect all ordering criteria
+  const orderByCriteria = [];
+
   for (let i = 0; i < parts.length; i++) {
-    let part = parts[i];
+    const part = parts[i];
     if (typeof part !== "string") {
       throw new ApiError(400, "cannot build sort with undefined");
     }
 
+    // Standard format: "-field" for desc, "field" for asc
     const isDesc = part.startsWith("-");
-    const column = isDesc ? (part = part.slice(1)) : part;
+    const column = isDesc ? part.slice(1) : part;
 
     const orderBy = orderByOptions[column];
     if (typeof orderBy === "undefined") {
-      throw new ApiError(400, `cannot sort by field: '${part}'`);
+      throw new ApiError(400, `cannot sort by field: '${column}'`);
     }
 
-    query = isDesc ? query.orderBy(desc(orderBy)) : query.orderBy(asc(orderBy));
+    // Add to criteria array instead of applying immediately
+    orderByCriteria.push(isDesc ? desc(orderBy) : asc(orderBy));
   }
 
-  return query;
+  // Apply all ordering criteria in a single orderBy call
+  return query.orderBy(...orderByCriteria);
 }
--- a/apps/api/src/services/agent-manager.service.ts
+++ b/apps/api/src/services/agent-manager.service.ts
@@ -864,16 +864,42 @@ export class AgentManager {
         params,
       );
 
+      // Service layer validates business inputs
       const {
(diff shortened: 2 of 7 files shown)


apps/api/src/services/agent-manager.service.ts
```diff
@@ -864,33 +864,71 @@ export class AgentManager {
...
+            throw new ApiError(400, `cannot sort by field: '${fieldName}'`);
+          }
+        }
+      }
+
+      // Get user agents (no competition sort - we're sorting competitions, not agents)
+      const userAgents = await this.getAgentsByOwner(userId, {
+        sort: "",
+        limit: 100,
+        offset: 0,
```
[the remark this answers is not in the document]
mzkrasner: The suggested fix would break the architectural pattern we use throughout the codebase. 

This follows the same pattern as `getUpcomingCompetitions()` (line 493) and other internal queries that use `limit: 100` for safety. This is an **internal query** to get agent IDs for filtering - applying competition pagination parameters here would violate layer separation.

Our two-tier architecture is intentional:
- **User-facing endpoints**: Accept pagination parameters (max 100)  
- **Internal service queries**: Use hardcoded safety limits

The current implementation maintains consistency with established patterns across the codebase.

apps/api/src/services/agent-manager.service.ts
```diff
@@ -864,33 +864,71 @@ export class AgentManager {
...
-        throw new ApiError(500, `cannot parse paging: ${error}`);
+        throw new ApiError(400, `Invalid pagination parameters: ${error}`);
       }
-      // Get all agents owned by this user
-      const userAgents = await this.getAgentsByOwner(userId, pagingParams);
+
+      // Validate sort parameter early (before checking if user has agents)
+      // This ensures consistent error handling regardless of user's agent count
+      if (validatedParams.sort) {
+        const validSortFields = [
```
dtbuchholz, line 919: could we try to add the remaining sort fields needed by the [user profile page](https://www.figma.com/design/WqQqcsg1g66EdMjjM8yfIM/Competitions-App---v.1.0--06-25-?node-id=4-16982&t=1pMftXX3iHtV7M4U-4)?
- `name` ✅  (comp name for the "Competitions" column)
- `status` ✅ (comp "Status" column)
- `agentName` ❌ ("Agent" column)
- `startDate`/`endDate` ✅  ("Date" column)
- `rank` ❌  (per-competition "Rank"...aka not global rank)
 
Screenshot 2025-06-16 at 8 51 43 PM

for the `rank` sorting, it can reuse some of the patterns you implemented in https://github.com/recallnet/js-recall/pull/633 wrt "computed sort fields" query logic.

apps/api/src/database/repositories/helpers.ts
```diff
@@ -20,22 +29,28 @@ export function getSort<T extends PgSelect>(
...
 
+    // Standard format: "-field" for desc, "field" for asc
     const isDesc = part.startsWith("-");
-    const column = isDesc ? (part = part.slice(1)) : part;
+    const column = isDesc ? part.slice(1) : part;
 
     const orderBy = orderByOptions[column];
     if (typeof orderBy === "undefined") {
-      throw new ApiError(400, `cannot sort by field: '${part}'`);
+      throw new ApiError(400, `cannot sort by field: '${column}'`);
```
dtbuchholz: we should keep the error as `'${part}'` because we want to maintain the `-` prefix in logging

dtbuchholz:
@mzkrasner this is looking solid as well. just a small request on adding two more sort fields...i figure we might as well tackle it in this PR vs a separate follow up

mzkrasner:
> @mzkrasner this is looking solid as well. just a small request on adding two more sort fields...i figure we might as well tackle it in this PR vs a separate follow up

Took a little more time than I had hoped for but just updated

Issue #630, opened 2025-06-16
[FEATURE] Add fields to `/agents/:agentId/competitions` and ensure sorting


dtbuchholz:
**Problem Statement**
The API `/agents/:agentId/competitions` doesn't support the fields needed by the UI's table sorting component on the app's `agents/:id` route/page.

![Image](https://github.com/user-attachments/assets/0a7fe264-b393-4ad4-a359-2a95960e29d7)

**Proposed Solution**
- We effectively need to add a new method similar to [ `attachAgentMetrics`](https://github.com/recallnet/js-recall/blob/313a4976bbb7a4c90617aebf066376bdd131cbfb/apps/api/src/services/agent-manager.service.ts#L998), but on a per-competition basis. 
- The method would extend the `/agents/:agentId/competitions` response by adding the following fields to the response and also as sorting options:
  - `name`: Name of the competition
  - `portfolioValue`: The agents portfolio value in the competition
  - `pnl`: The agent's PnL in the competition
  - `totalTrades`: The total number of trades (see `countAgentTrades` for an example, but this will need to be competition-specific trades, and not aggregate across all competitions)
  - `bestPlacement`: Needs the `totalAgents` and `rank`, but internally, the sorting logic should only need to apply to the `rank` field
- Also, make sure the UI table sorting works. You can see that table on NextJS page at `/agents/:id`, and this is the component: [`CompetitionTable`](https://github.com/recallnet/js-recall/blob/main/apps/comps/components/agent-profile/index.tsx#L237)

Note: `skills` doesnt need to be sortable; it's an array of string values. Also, take a look at `calculateAgentMetrics`, `getVoteCountsByCompetition`, and `countAgentTrades` for similar implementations).

**Impact**
UI needs this feature to properly sort data on the `agents/:id` pages.

**Priority**
High/highest

Issue #621, opened 2025-06-16
[BUG] Fix sorting with `/competitions/:id/agents`


dtbuchholz:
**Problem Statement**
The `/competition/:id/agents` doesnt sort properly

**Proposed Solution**
Fix the sorting logic and add more comprehensive tests.

**Impact**
- Table UI sorting

**Priority**
High. App v1 needs this

dtbuchholz:
closed by https://github.com/recallnet/js-recall/pull/633

Issue #620, opened 2025-06-16
[BUG] Fix computed fields post-processing sort logic


dtbuchholz:
**Bug Description**
We allow you to sort competition agents by:
- `agents` table fields
- "stats" fields that extend the base agent response _after_ the db query

If you, e.g., request the "bottom 10" agents by rank, the only way to achieve this is by getting _all_ agents, computing their stats, and then applying the sorting to the subset of agents that fit within request limit/offset params. 

Some API queries apply this logic, which is inefficient but meets the user's expectation. Other queries do not, which means the sorting logic is only applied to the "window" of agents returned by the initial database query.

**Steps to Reproduce**
1. You need to seed the agents table with some number > the `limit` you apply
2. Make a query to `/competitions/agents` with sort + limit

See the `getCompetitionAgents` in `competition.controller.ts`. 

The fix is easy: "get _all_ rows, and sort in-memory returning an API response." See https://github.com/recallnet/js-recall/pull/633 as an example for handling db vs "computed" sort fields.

This will work, for now, but there are performance considerations. We should do one of the following:
- WIndow function
- Materialized view
- Add new columns to track this in `competitions_leaderboard` (#559)

**Additional context**
As a follow up to #509, we need to continue to fix the pagination/query logic.

dtbuchholz:
See this PR for the logic we want to implement: https://github.com/recallnet/js-recall/pull/643

Here's the logic we want to follow:
- controllers: handle all parsing with zod validation; if some "sort" field is invalid, it should never hit a service layer.
- services: handle interfacing with the controller<>repositories, and enhance data with computation if needed. they need to handle "splitting up" a sort field and pass it to either the database query, or use the field within the service itself for post-query computed fields.
- repositories: should purely execute database queries, and (mostly) shouldnt be "aware" of the any non-database things like field parsing or post-query compute logic.

See the `getCompetitionAgentsWithMetrics` for a reference impl. Here's the usage pattern:
1. Create our new types and zod schemas in `types/sort/<name>.ts`; see `types/sort/agent.ts` as a reference.
2. Use `lib/sort.ts` and `makeSortFieldSchema` to create unified types for controller-level validation. This auto-creates `-`-prefixed descending sort fields.
3. In the service, use `splitSortField` to figure out if post-processing sort is required; see `services/competition-manager.service.ts` and `getCompetitionAgentsWithMetrics` as a reference.
[...]

PR #633, opened 2025-06-16
feat (api): fields added to /agents/:agentId/competitions and ensure sorting

mzkrasner:
# Summary

### Issue Overview
GitHub Issue #630 identified that the `/agents/:agentId/competitions` API endpoint was missing critical fields needed by the UI's table sorting component. 

### Required Fields
The UI needed the following agent-specific metrics for each competition:
- `portfolioValue`: Agent's current portfolio value in the competition
- `pnl`: Agent's profit/loss amount and percentage  
- `totalTrades`: Number of trades made by the agent in that competition
- `bestPlacement`: Agent's rank and total participants in the competition

### Solution Implemented
Following the existing `attachAgentMetrics` pattern, we implemented a per-competition metrics enhancement:

**Repository Layer:**
- Added `countAgentTradesInCompetition()` for competition-specific trade counting
- Added `getAgentCompetitionRanking()` for accurate ranking calculations
- Extended sorting support to include computed fields

**Service Layer:**
- Added `attachCompetitionMetrics()` method to enhance competitions with agent-specific data
- Modified `getCompetitionsForAgent()` to apply metrics to each competition
- Implemented proper error handling with sensible defaults

**API Enhancement:**
- Enhanced response includes all required fields with proper TypeScript types
- Full sorting support for new computed fields (`portfolioValue`, `pnl`, `totalTrades`, `rank`)
- Maintained backward compatibility with existing clients

### Testing & Quality
- **37/37 E2E tests passing** covering functionality, edge cases, and multi-agent scenarios
- Comprehensive test coverage including cross-competition ranking accuracy
- Updated OpenAPI documentation with field descriptions and examples

### DIFF
--- a/apps/api/openapi/openapi.json
+++ b/apps/api/openapi/openapi.json
@@ -3,7 +3,7 @@
   "info": {
     "title": "Trading Simulator API",
     "version": "1.0.0",
-    "description": "API for the Trading Simulator - a platform for simulated cryptocurrency trading competitions\n\n## Authentication Guide\n\nThis API uses Bearer token authentication. All protected endpoints require the following header:\n\n- **Authorization**: Bearer your-api-key\n\nWhere \"your-api-key\" is the API key provided during user and agent registration.\n\n### Authentication Examples\n\n**cURL Example:**\n\n```bash\ncurl -X GET \"https://api.example.com/staging/api/account/balances\" \\\n  -H \"Authorization: Bearer abc123def456_ghi789jkl012\" \\\n  -H \"Content-Type: application/json\"\n```\n\n**JavaScript Example:**\n\n```javascript\nconst fetchData = async () => {\n  const apiKey = 'abc123def456_ghi789jkl012';\n  const response = await fetch('https://api.example.com/staging/api/account/balances', {\n    headers: {\n      'Authorization': `Bearer ${apiKey}`,\n      'Content-Type': 'application/json'\n    }\n  });\n\n  return await response.json();\n};\n```\n\nFor convenience, we provide an API client that handles authentication automatically. See `docs/examples/api-client.ts`.\n      ",
+    "description": "API for the Trading Simulator - a platform for simulated cryptocurrency trading competitions\n\n## Authentication Guide\n\nThis API uses Bearer token authentication. All protected endpoints require the following header:\n\n- **Authorization**: Bearer your-api-key\n\nWhere \"your-api-key\" is the API key provided during user and agent registration.\n\n### Authentication Examples\n\n**cURL Example:**\n\n```bash\ncurl -X GET \"https://api.example.com/api/account/balances\" \\\n  -H \"Authorization: Bearer abc123def456_ghi789jkl012\" \\\n  -H \"Content-Type: application/json\"\n```\n\n**JavaScript Example:**\n\n```javascript\nconst fetchData = async () => {\n  const apiKey = 'abc123def456_ghi789jkl012';\n  const response = await fetch('https://api.example.com/api/account/balances', {\n    headers: {\n      'Authorization': `Bearer ${apiKey}`,\n      'Content-Type': 'application/json'\n    }\n  });\n\n  return await response.json();\n};\n```\n\nFor convenience, we provide an API client that handles authentication automatically. See `docs/examples/api-client.ts`.\n      ",
     "contact": {
       "name": "API Support",
       "email": "support@example.com"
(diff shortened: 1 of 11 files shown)


apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -429,6 +429,53 @@ export async function getAgentPortfolioSnapshots(
...
+    );
+
+    // Find the agent's position (1-based ranking)
+    const agentIndex = sortedSnapshots.findIndex(
+      (snapshot) => snapshot.agentId === agentId,
+    );
+
+    // If agent not found in snapshots, return 0 rank
+    if (agentIndex === -1) {
+      return { rank: 0, totalAgents: sortedSnapshots.length };
```
dtbuchholz: the "rank 0" is problematic because a typical query is: "give my all agents and sort by rank". rank `0` would come before rank `1` and muddy the results.

~~so, we should just make it nullable/undefined (whichever makes it easier wrt our return types). the client can define handling the `undefined / totalAgents` logic.~~

edit: actually, we should just make `bestPlacement` nullable/optional, so the `rank` & `totalAgents` is only supplied if both exist. that way, it's easier to handle on the frontend for rendering: `bestPlacement ? "${rank} / ${totalAgents}" : "N/A"`

apps/api/src/services/agent-manager.service.ts
```diff
@@ -832,13 +842,42 @@ export class AgentManager {
         params,
       );
 
-      // Get agents from repository
+      // Get competitions from repository
       const results = await findAgentCompetitions(agentId, params);
 
+      // Attach metrics to each competition
```
dtbuchholz: if we do this:
```ts
const competitions = results.competitions.filter(
  (comp) => comp !== null,
);
const enhancedCompetitions = await Promise.all(     
  results.competitions.map(...)
  // ...
)
```
then the `null` checks via `if (!competition) ...` and `validCompetitions` aren't needed.

apps/api/src/services/agent-manager.service.ts
```diff
@@ -1030,6 +1069,123 @@ export class AgentManager {
...
+    }
+  }
+
+  /**
+   * Sort competitions by computed fields
+   * @param competitions Array of competitions with metrics
+   * @param sortString Sort string (e.g., "portfolioValue", "-pnl")
+   * @returns Sorted competitions array
+   */
+  private sortCompetitionsByComputedField(
```
dtbuchholz, line 1149: note to self: revisit this logic in https://github.com/recallnet/js-recall/issues/620 

(what you have is good / necessary. we just need to figure out a solution for "load _all_ results in the db query because the post-processing needs to sort the fields, too")

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -429,6 +429,53 @@ export async function getAgentPortfolioSnapshots(
...
+export async function getAgentCompetitionRanking(
+  agentId: string,
+  competitionId: string,
+): Promise<{ rank: number; totalAgents: number }> {
+  try {
+    // Get all latest portfolio snapshots for the competition
+    const snapshots = await getLatestPortfolioSnapshots(competitionId);
+
+    if (snapshots.length === 0) {
+      return { rank: 0, totalAgents: 0 };
```
dtbuchholz: oh and same thing here re: the point about rank zero & totalAgents zero....if there's no `bestPlacement`, then we should return `undefined` because handling "zero rank" can potentially be misleading for the client to handle.

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -429,6 +429,53 @@ export async function getAgentPortfolioSnapshots(
...
+      rank: agentIndex + 1, // Convert to 1-based ranking
+      totalAgents: sortedSnapshots.length,
+    };
+  } catch (error) {
+    console.error(
+      "[CompetitionRepository] Error in getAgentCompetitionRanking:",
+      error,
+    );
+    // Return default values on error to prevent breaking the flow
+    return { rank: 0, totalAgents: 0 };
```
dtbuchholz: (and here)

dtbuchholz:
@mzkrasner my only qualm is the `bestPlacement` values for edge cases but everything else looks solid.

i noticed a similar query sort problem while fixing the other comps API in #621. i like the repository-level approach you used here where the response tells you "is this a computed sort field?"...i tried a zod-based one at the service layer and it got a bit messy.

apps/api/openapi/openapi.json
```diff
@@ -15,15 +15,15 @@
   },
   "servers": [
     {
-      "url": "https://api.competitions.recall.network",
+      "url": "https://api.competitions.recall.network/staging",
```
dtbuchholz: is there a reason these got changed and include the `staging` prefix and define different ports?

mzkrasner: that's odd - let me resolve that

mzkrasner: Ah - still had staging assigned to `API_PREFIX` in my local .env which was getting picked up by swagger

dtbuchholz (approved):

Issue #458, opened 2025-05-29
[IMPLEMENTATION] Add per-competition agent status


dtbuchholz:
**Problem Statement**
Currently, agent status is tracked globally.  We want to update this to be per competition so an agent can have different status' simultaneously in different competitions.

Track per-competition agent status. via https://github.com/recallnet/js-recall/pull/455
> i wonder if a _hard_ delete in [`competition_agents`](https://github.com/recallnet/js-recall/blob/8deb5fc482accc10776a38fef6fce8addaa82678/apps/api/src/database/schema/core/defs.ts#L183) is the right approach—vs adding a new column that tracks the agent's status within the competition. e.g., we probably need the posterity tracking to ensure agents dont just leave halfway through if theyre performing poorly.

**Proposed Solution**

there are implications elsewhere because our [`agents`](https://github.com/recallnet/js-recall/blob/8deb5fc482accc10776a38fef6fce8addaa82678/apps/api/src/database/schema/core/defs.ts#L88) table uses the `status` field to indicate if an agent is active in a competition. e.g., once a competition ends, the agent is [effectively deactivated](https://github.com/recallnet/js-recall/blob/8deb5fc482accc10776a38fef6fce8addaa82678/apps/api/src/database/repositories/agent-repository.ts#L380) globally (via [`endCompetition`](https://github.com/recallnet/js-recall/blob/21a9c760185c9daa6277096de3438109f418c8eb/apps/api/src/services/competition-manager.service.ts#L237)). this logic should probably exist in the `competition_agents`, whereas the `agents` state is more of a global view, irrespective of an active competition.

backend needs to add new cols to the tables above and rework POST/DELETE APIs for joining or leaving competitions.

**Definition of Done**
- [ ] Database schema changes
- [ ] Expose agent status via the appropriate endpoints
- [ ] 

**Impact**
Without this:
- agent cannot update profile in between competitions
- if agent is deactivated, they lose the ability to do some basic global operations that aren't related to a specific competition.
- Cannot support multiple simultaneous competitions
- 
**Priority**
Medium/high. We don't want to let agents just leave competitions and have no record of it because the db deleted the row in `competition_agents`

dtbuchholz:
@stbrody we might want to prioritize this due prevent a confusing global leaderboard page.

the example below is the global leaderboard under these conditions:
- 20 agents participate in competition 1
- then, we end competition 1
- 2 agents participate in competition 2; keep it in an "active" competition status
[...]

PR #634, opened 2025-06-16
fix(api): count total active agents

dtbuchholz:
- Populate the "active" agents in the leaderboards `activeAgents` instead of hardcoding as zero.
- Use the `agent_rank` table to give us an accurate representation of "active" agents (see #458 for more info re: agent "status" and why this query works well for our needs).

### DIFF
--- a/apps/api/e2e/tests/leaderboard.test.ts
+++ b/apps/api/e2e/tests/leaderboard.test.ts
@@ -87,7 +87,7 @@ describe("Leaderboard API", () => {
       (await agentClient1.getGlobalLeaderboard()) as GlobalLeaderboardResponse;
     expect(leaderboard.success).toBe(true);
 
-    expect(leaderboard.stats.activeAgents).toBe(0);
+    expect(leaderboard.stats.activeAgents).toBe(2);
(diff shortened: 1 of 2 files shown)


brunocalza (approved):
Looks good! So, the global stats always show the current number of agents that are competing?

dtbuchholz:
buuut...you made me realize a better approach for this is: query the `agent_rank` table for unique agent IDs, and join it with the `agents` table to find all "active" agents.

the only problem with the latter half of that query is that the current logic is a bit problematic (see #458), which :
- a new agent is created in an "active" status (a column in the `agents` table, not `competitions` / `competition_agents`)
- an agent enters a competition -> there's no per-comp agent status, so we update the `agents` table's global status
- you query for all "active" agents -> this is all agents participating in the competition _plus_ any newly created agent
- a competition ends -> each entry in `agents` who was participating is switched from "active" to inactive
- you query for all "active" agents -> any agent who was in the recently ended comp will _not_ be included

lmk if that makes sense to you (i just pushed this change).

brunocalza:
Yeah, this changes the meaning of total active agents to how many agents have participated in a competition, which I think makes more sense than the previous for the global leaderboard page.

PR #636, opened 2025-06-17
fix(comps): fix hydration mismatch in siwe

joewagner:
## Overview

The SIWE component has been generating hydration mismatch errors for me during local dev.  This ensures the server and initial client rendering match.

## Details

The issue comes from the `useUserSession` hook, which is behaving differently on the server vs. the client.  Specifically, the value of `user` on line 27 is available for the server's prerender, but it is null during the initial client render.  I _think_ this is due to the tanstack `useQuery` doing a fetch, which has a variable response between client and server.

What I did here simply leverages useState to only render after mounting, so there is no way there can be a hydration mismatch.
We could alternatively refactor the conditional rendering logic, but this seemed less complicated. I'm open to alternatives.

### DIFF
--- a/apps/comps/components/siwe/index.tsx
+++ b/apps/comps/components/siwe/index.tsx
@@ -4,7 +4,7 @@ import { ConnectButton } from "@rainbow-me/rainbowkit";
 import { LogOut } from "lucide-react";
 import Image from "next/image";
 import { useRouter } from "next/navigation";
-import React from "react";
+import React, { useEffect, useState } from "react";
 
 import { Button } from "@recallnet/ui2/components/button";
 import {
@@ -25,10 +25,15 @@ export const SIWEButton: React.FunctionComponent<
   const router = useRouter();
   const logout = useLogout();
   const { user } = useUserSession();
+  const [isMounted, setIsMounted] = useState(false);
+
+  useEffect(() => setIsMounted(true), []);
 
   const handleLogout = async () => {
     logout.mutate();
   };
+
+  if (!isMounted) return <div></div>;
 
   if (user) {
     return (


dtbuchholz:
@joewagner hm dang, i still see the error when i run this branch locally.

another option is to do something like this https://github.com/recallnet/js-recall/pull/638/files with dynamic imports & explicit non-ssr declarations. i'd have to dig in a little more to see if there's a better fix, though, but it works.

curious if you have any better ideas @ramirotw

joewagner:
> @joewagner hm dang, i still see the error when i run this branch locally.
> 
> another option is to do something like this https://github.com/recallnet/js-recall/pull/638/files with dynamic imports & explicit non-ssr declarations. i'd have to dig in a little more to see if there's a better fix, though, but it works.
> 
> curious if you have any better ideas @ramirotw

hmm, that's odd.  Does the error show you a diff of the two renders if you click on the little nextjs logo in the bottom right?

dtbuchholz:
> hmm, that's odd. Does the error show you a diff of the two renders if you click on the little nextjs logo in the bottom right?

yeah...but after double checking, the error is for a different component: `components/profile-skeleton.tsx`. this error happens if i'm on the "My Profile" page and refresh it:
```
  ...
    <RenderFromTemplateContext>
      <ScrollAndFocusHandler segmentPath={[...]}>
        <InnerScrollAndFocusHandler segmentPath={[...]} focusAndScrollRef={{apply:false, ...}}>
          <ErrorBoundary errorComponent={undefined} errorStyles={undefined} errorScripts={undefined}>
            <LoadingBoundary loading={null}>
              <HTTPAccessFallbackBoundary notFound={undefined} forbidden={undefined} unauthorized={undefined}>
                <RedirectBoundary>
                  <RedirectErrorBoundary router={{...}}>
                    <InnerLayoutRouter url="/profile" tree={[...]} cacheNode={{lazyData:null, ...}} segmentPath={[...]}>
                      <ClientPageRoot Component={function ProfilePage} searchParams={{}} params={{}}>
                        <ProfilePage params={Promise} searchParams={Promise}>
                          <AuthGuard skeleton={<ProfileSkeleton>}>
                            <ProfileSkeleton>
+                             <div className="mb-5 flex items-center gap-4">
-                             <Suspense>
                              ...
                      ...
```
in `components/profile-skeleton.tsx`:
```
   7 |     <>
   8 |       {/* Header: Back button + Breadcrumb */}
>  9 |       <div className="mb-5 flex items-center gap-4">
     |       ^
  10 |         <Skeleton className="h-4 w-40" />
  11 |         <Skeleton className="h-4 w-40" />
  12 |       </div>
```

Issue #517, opened 2025-06-05
Implement voting flow in comps UI


dtbuchholz:
Frontend impl only

see #349; figma [here](https://www.figma.com/design/uuMfHT5O7zW4dfTgj6ryGG/Wednesday-s-Designs?node-id=10798-32680&p=f&m=dev)

dtbuchholz:
cc @Rcontre360 / @ramirotw after #511 (finalized comps components / flows), we should start implementing the updated voting flows described in the figma. note that we created [the following APIs](https://github.com/recallnet/js-recall/blob/main/apps/api/src/routes/vote.routes.ts) for user-specific voting actions:
- `POST /api/user/vote`: cast a vote
- `GET /api/user/votes`: get a user's vote
- `GET /api/user/votes/{competitionId}/state`: get a user's status for a specific competition

lmk if you see any gaps, such as path params or other fields. as for tallying up votes at a competition or leaderboard level, we're adding those to certain APIs. for example, `/api/competitions/:id` includes a `voteCount` per agent.

PR #639, opened 2025-06-17
feat(comps): add voting integration

ramirotw:
Fixes https://github.com/recallnet/js-recall/issues/517


dtbuchholz:
@ramirotw lgtm. gave it a test locally...voting logic and animations look great!

there are a few minor things that i think we can clean up as part of the QA process—mainly because idk if the designs made it explicit, and we should just land the feature asap.

---

(i'm making some notes for myself)

this problem might be due to some of the hydration errors we're aware of...not sure:
- after i initially connect my account in the comps hub, the "Vote" button isn't clickable unless i refresh the page

and these are design-related questions we'll need answered: 
- on the comps hub, the "vote" button default bg doesn't make it obvious that it's clickable. the style diff between an unauthenticated vs connected user isn't noticeable. 
- on the competition page, after i cast a vote, the "up arrow" should not change bg on hover. it needs to somehow be disabled. else, i try to click the vote button again, and (as expected) nothing happens.

dtbuchholz:
note: github seems to be having issues right now, so i can't approve...but it's effectively approved
Screenshot 2025-06-17 at 3 51 39 PM

ramirotw:
> * after i initially connect my account in the comps hub, the "Vote" button isn't clickable unless i refresh the page

Needed to invalidate the competitions cache on login so the app requests the competitions again with the voting data. Fixed in [9ab7a2e](https://github.com/recallnet/js-recall/pull/639/commits/9ab7a2e9f5f5fc38f93d3063aee60ae59cd9c50d)


> * on the comps hub, the "vote" button default bg doesn't make it obvious that it's clickable. the style diff between an unauthenticated vs connected user isn't noticeable. 

Added a border color to make it more noticeable. Regarding the overall app design, now that we’ve more or less implemented the MVP features, we should probably ask Jovi to create a design document outlining things like typography, color palette, spacing, etc., and ensure everything is covered through our CSS variables. Right now, the app has a lot of custom styles that should be standardized.
Fixed in [026f071](https://github.com/recallnet/js-recall/pull/639/commits/026f071a713b0c66637fb7f0f3b72bd48ac3c23b)

> * on the competition page, after i cast a vote, the "up arrow" should not change bg on hover. it needs to somehow be disabled. else, i try to click the vote button again, and (as expected) nothing happens.

Fixed in [c03ff83](https://github.com/recallnet/js-recall/pull/639/commits/c03ff834462c12cd2a3b90f7bc7493ac41bf543d)

dtbuchholz:
> Regarding the overall app design, now that we’ve more or less implemented the MVP features, we should probably ask Jovi to create a design document outlining things like typography, color palette, spacing, etc., and ensure everything is covered through our CSS variables

great idea. i'll if i can get something asap https://github.com/recallnet/js-recall/issues/641

PR #642, opened 2025-06-18
feat(comp): implement voting banner

joewagner:
fixes: https://github.com/recallnet/js-recall/issues/539


apps/comps/components/clock/index.tsx
```diff
@@ -5,25 +5,50 @@ import { cn } from "@recallnet/ui2/lib/utils";
...
+  const pluralize = (amount: number, word: string) =>
+    `${word}${amount > 1 ? "s" : ""}`;
   useEffect(() => {
     const updateCountdown = () => {
       const now = new Date();
       const difference = targetDate.getTime() - now.getTime();
 
       if (difference > 0) {
+        const days = Math.floor(difference / (1000 * 60 * 60 * 24));
+        if (showDuration === true) {
```
joewagner: To match the design I added a `showDuration` prop that will show more human readable info until the time left is less than 48 hours.

apps/comps/app/competitions/[id]/page.tsx
```diff
@@ -89,16 +90,17 @@ export default function CompetitionPage({
   }
 
   return (
-    <>
+    <div style={{ marginTop: "-40px" }}>
```
Comment (joewagner) [inline apps/comps/app/competitions/[id]/page.tsx:93]: The padding and margin rules that exist currently make it a little tricky to get the banner to be a child component and still push up against the nav, as per the design, this negative margin combined with a matching positive margin for the breadcrums work afaict.

apps/comps/components/competition-voting-banner.tsx
```diff
@@ -0,0 +1,121 @@
+"use client";
```
joewagner, line 1: This is the new component to show voting states

apps/comps/components/competition-voting-config.tsx
```diff
@@ -0,0 +1,108 @@
+import { Competition, CompetitionStatus } from "@/types";
```
joewagner, line 1: This is also new, and modeled after `getCompetitionStatusConfig` in `competition-status-banner.tsx`

apps/comps/components/competition-voting-config.tsx
```diff
@@ -0,0 +1,108 @@
...
+    : null;
+  const votingStart = competition.votingStartDate
+    ? new Date(competition.votingStartDate)
+    : null;
+  const votingEnd = competition.votingEndDate
+    ? new Date(competition.votingEndDate)
+    : null;
+  const now = new Date();
+
+  if (hasVoted) {
```
joewagner, line 30: there's a **lot** of conditionals here, but as best I can tell the correctly map the state to the design.  The only thing I'm not really sure on is the state where registration is not open yet.  See the TODO comment below.

apps/comps/components/siwe/index.tsx
```diff
@@ -23,11 +23,16 @@ export const SIWEButton: React.FunctionComponent = () => {
   const router = useRouter();
   const logout = useLogout();
   const { user } = useUserSession();
+  const [isMounted, setIsMounted] = useState(false);
+
+  useEffect(() => setIsMounted(true), []);
```
joewagner: This is left over from https://github.com/recallnet/js-recall/pull/636

I can remove it if that doesn't make sense, but I added it here just so I didn't see the error in the console.

ramirotw: @dtbuchholz @joewagner about the hydration error:

The problem isn't with the components themselves, but with Jotai's userAtom. The getOnInit function makes the localStorage read synchronous so that user and auth data from localStorage are already available on the first render. So any component that uses this atom through a hook will see user as null during pre-rendering on the server (since there's no localstorage there), but on the first client-side render, after reading from localstorage, user is not null (if the user is logged in), which causes the mismatch.
I'm going to look into the best way to fix this without monkey-patching every component

```js
export const userAtom = atomWithStorage<UserStorage>(
  "user",
  {
    user: null,
    status: "unauthenticated",
  },
  undefined,
  {
    getOnInit: true,
  },
);
```

dtbuchholz: @ramirotw makes sense...created an issue to track this https://github.com/recallnet/js-recall/issues/645

@joewagner given that, let's just revert the fix

joewagner: Sound good to me.  I'll take this out of this PR.

apps/comps/components/clock/index.tsx
dtbuchholz: nice! one small tweak: it looks like the designs specify 24 hours as the cutoff, and not 48 hours?

Screenshot 2025-06-18 at 10 37 12 AM

apps/comps/components/clock/index.tsx
```diff
@@ -5,25 +5,50 @@ import { cn } from "@recallnet/ui2/lib/utils";
...
+            } else {
+              const months = Math.floor(
+                difference / (1000 * 60 * 60 * 24 * 7 * 30),
+              );
+              setDuration(`${months} ${pluralize(months, "month")}`);
+            }
+
+            return;
+          } else {
+            setDuration("");
```
dtbuchholz: should we have a default like "TBA" so the UI renders something?

apps/comps/components/clock/index.tsx
```diff
@@ -34,14 +59,15 @@ export const CountdownClock: React.FC<CountdownClockProps> = ({
       } else {
         setTimeLeft({ hours: 0, minutes: 0, seconds: 0 });
         setIsCountdownFinished(true);
+        setDuration("");
```
dtbuchholz: same thing here...maybe "TBA"? (this might be related to the points below.)

apps/comps/components/competition-voting-config.tsx
dtbuchholz: agreed. i created a follow up to revisit the voting API logic: https://github.com/recallnet/js-recall/issues/647. i think the API's response for `votingEnable` is causing some of the problems below.

a couple of examples where this becomes problematic, and i think point (2) is the only one that requires the issue above:
1. for status "pending" + when a competition has no comp start / end date, we should default to "TBA". else, the banner will say "Registration opens in ..." and doesnt render anything thereafter. i _think_ this is would be fixed via the `setDuration` comment above.
Screenshot 2025-06-18 at 10 32 39 AM

2. for status "pending" + voting windows are null, the banner shows "Voting is open"—i think this would be fixed by issue i noted above.
Screenshot 2025-06-18 at 10 43 34 AM

3. for status "completed", we should not show any banners. (fwiw, i dont think this was clear in the designs.)
Screenshot 2025-06-18 at 11 21 35 AM

apps/comps/components/competition-voting-config.tsx
```diff
@@ -0,0 +1,108 @@
...
+    if (status === CompetitionStatus.Pending) {
+      return {
+        subTitle: "Counting Votes!",
```
dtbuchholz: since the other CTAs are sentence case, we should maintain that: `Counting votes!`

apps/comps/components/competition-voting-config.tsx
```diff
@@ -0,0 +1,108 @@
...
+        description: compStartDate ? "Competition starts in..." : "",
+        variant: "gray",
+        untilTime: compStartDate,
+        phase: null,
+      };
+    }
+    if (status === CompetitionStatus.Active) {
+      return {
+        subTitle: "Counting Votes!",
```
dtbuchholz: sentence case

apps/comps/components/competition-voting-config.tsx
```diff
@@ -0,0 +1,108 @@
...
+    };
+  }
+
+  // all states below here are for competitions that the user has not voted on yet
+
+  // TODO: what defines registration start time?  I'm assuming that if there
+  //  isn't a comp start time, then registration is not open.
+  if (!compStartDate) {
+    return {
+      subTitle: "Get Ready!",
```
dtbuchholz: correct—registration start time == competition start date. it's a bit confusing imo because the banner hold state for two different types of users:
- a dev who "registers" an agent in the comp
- a user who only cares about voting

(also: sentence case for `Get Ready!`)

apps/comps/components/competition-voting-config.tsx
```diff
@@ -0,0 +1,108 @@
...
+      subTitle: "Get ready!",
+      description: "Voting opens in...",
+      variant: "blue",
+      untilTime: votingStart,
+      phase: "voting",
+    };
+  }
+  if (votingEnd && now > votingEnd) {
+    return {
+      subTitle: "Counting Votes!",
```
dtbuchholz: sentence case

apps/comps/components/competition-voting-config.tsx
dtbuchholz: follow up—we _do_ want allow for competitions _without_ voting enabled. i.e., i think https://github.com/recallnet/js-recall/issues/646 will help with this an ensure `votingEnabled` is `false` when the voting dates are not set.

apps/comps/components/clock/index.tsx
joewagner: Good call, I'll change this

apps/comps/components/competition-voting-config.tsx
joewagner: Good catch!

apps/comps/components/clock/index.tsx
joewagner: This is confusing.  I'll add a comment here, but when `duration` is equal to `""` the render logic below only shows the countdown timer, or if the countdown is finished it shows the html on line 77.

apps/comps/components/clock/index.tsx
joewagner: Related to the above response, when the countdownFinished value is `true` the duration and clock are hidden.
I was following the existing impl here by setting the time and duration back to their initial values.

apps/comps/components/competition-voting-config.tsx
joewagner: This has kind of a lot going on, I'm happy to jump into a call if you want to talk through.

For `#1`, as far as I can tell from the notes in the issue, the only time status will come into consideration is if the user has already voted. So maybe if this is the case, it can say "Counting votes! Competition starts soon?" Thoughts?

For `#2`, If there is no voting start date or voting end date the api allows voting as long as the status isn't "ended".  Which makes me think the banner should indicate voting is open.  If that buisness logic isn't correct, we would need to change the api before the front-end.

For `#3`, I didn't realize that, I'll make that change.

apps/comps/components/competition-voting-config.tsx
joewagner: Good call, fixing this now.

apps/comps/components/competition-voting-config.tsx
dtbuchholz: hm yeah these banners are getting confusing. i shouldve caught that in the design review.

For `#1`—ah iirc this is what you brought up in standup. the main issue is our backend only considers competitions as: pending, active, or ended. "registration" becomes ambiguous and confuses the possible states. maybe we just stick with what you implemented and clean it up during the QA process since that's when product will take a deeper look.

For `#2`
> if there is no voting start date or voting end date the api allows voting as long as the status isn't "ended"

hm yeah i think that's what we should fix in the issue above. i.e., if there's no voting window, then voting isnt technically enabled because there's no window established for voting. related—i asked product today if we should allow for non-voting comps. the response was _yes_. so, that aligns with the business logic change, too.

i.e., if we make the API change, then _i think_ the banner change will automatically use the correct logic because voting will be disabled for null start/end voting windows? aka we dont need to make any frontend changes in this PR?

joewagner: This all sounds good, and after the changes I pushed a few minutes ago this should be resolved.

dtbuchholz (approved):
ran it locally, lgtm. (will approve so you can merge after the linting fix)

PR #643, opened 2025-06-18
fix(api): `compititions/:id/agents` query sorting and pagination

### LINKED ISSUE #621: [BUG] Fix sorting with `/competitions/:id/agents`

dtbuchholz:
- Closes #621 
- Related to #620 (and sorting _with_ computed value in post-query logic)

### DIFF
--- a/apps/api/e2e/tests/competition.test.ts
+++ b/apps/api/e2e/tests/competition.test.ts
@@ -34,6 +34,7 @@ import {
   startTestCompetition,
 } from "@/e2e/utils/test-helpers.js";
 import { wait } from "@/e2e/utils/test-helpers.js";
+import { ServiceRegistry } from "@/services/index.js";
 
 describe("Competition API", () => {
   let adminApiKey: string;
@@ -1799,6 +1800,9 @@ describe("Competition API", () => {
       adminApiKey,
       agentName: "Sort Test Client Agent",
     });
+    // Force a snapshot directly
+    const services = new ServiceRegistry();
+    await services.portfolioSnapshotter.takePortfolioSnapshots(competitionId);
 
     // Test sorting by name (ascending)
     const nameAscResponse = (await client.getCompetitionAgents(competitionId, {
@@ -1815,7 +1819,7 @@ describe("Competition API", () => {
 
     // Test sorting by name (descending)
     const nameDescResponse = (await client.getCompetitionAgents(competitionId, {
-      sort: "name_desc",
+      sort: "-name",
     })) as CompetitionAgentsResponse;
 
     expect(nameDescResponse.success).toBe(true);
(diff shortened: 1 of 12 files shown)


apps/api/e2e/tests/competition.test.ts
```diff
@@ -1827,16 +1831,202 @@ describe("Competition API", () => {
     expect(nameDescOrder[2]).toBe("Alpha Sort Agent");
 
     // Test sorting by position (default)
-    const positionResponse = (await client.getCompetitionAgents(competitionId, {
-      sort: "position",
+    const positionAscResponse = (await client.getCompetitionAgents(
```
dtbuchholz, line 1834: added tests for ascending/descending: rank (aka position), name, portofolio, pnl%, 24h pnl%, and votes
Screenshot 2025-06-18 at 1 15 48 AM

apps/api/src/controllers/competition.controller.ts
```diff
@@ -678,13 +681,14 @@ export function makeCompetitionController(services: ServiceRegistry) {
...
         // Get competition ID from path parameter
-        const competitionId = req.params.competitionId;
-        if (!competitionId) {
-          throw new ApiError(400, "Competition ID is required");
-        }
+        const competitionId = ensureUuid(req.params.competitionId);
 
         // Parse query parameters
-        const queryParams = CompetitionAgentsParamsSchema.parse(req.query);
+        const parseQueryParams = AgentQuerySchema.safeParse(req.query);
```
dtbuchholz, line 691: our `types` barrel file is massive. we know we need to do some refactoring.

with that in mind, i created net-new types in `types/sort/...`. now that the APIs are mostly wrapped up, i figure we can follow this pattern whenever we clean things up. totally up for discussion tho on what i put together is how we wanna handle zod + controller-level sorts.

apps/api/src/controllers/competition.controller.ts
```diff
@@ -695,73 +699,16 @@ export function makeCompetitionController(services: ServiceRegistry) {
 
         // Get agents for the competition with pagination
         const { agents, total } =
-          await services.agentManager.getAgentsForCompetition(
+          await services.competitionManager.getCompetitionAgentsWithMetrics(
             competitionId,
             queryParams,
           );
 
-        // Get leaderboard data for the competition to get scores and positions
```
dtbuchholz, line 703: i realized there are a decent amount of methods in the controller that handle "compute some data" after db queries. (i likely contributed to some of this code, hah.)

this is probably obvious, but i figured i'd mention (in case anyone disagrees):
1. controllers: handle all parsing with zod validation; if some "sort" field is invalid, it should never hit a service layer.
2. services: handle interfacing with the controller<>repositories, and enhance data with computation if needed. they need to handle "splitting up" a sort field and pass it to either the database query, or use the field within the service itself for post-query computed fields.
3. repositories: should purely execute database queries, and (mostly) shouldnt be "aware" of the any non-database things like field parsing or post-query compute logic.

apps/api/src/controllers/request-helpers.ts
```diff
@@ -58,3 +96,12 @@ export function ensureCompetitionUpdate(req: Request) {
 
   return data;
 }
+
+/**
+ * Check if the request is authenticated as an admin
+ * @param req Express request
+ * @returns True if the request is authenticated as an admin, false otherwise
+ */
+export function checkIsAdmin(req: Request) {
```
dtbuchholz, line 105: mostly used in logging aka doesnt prevent execution

apps/api/src/database/repositories/agent-repository.ts
```diff
@@ -201,11 +198,13 @@ export async function findAgentCompetitions(
...
  * @param competitionId Competition ID
  * @param params Query parameters for filtering, sorting, and pagination
+ * @param isComputedSort Whether computed sorting is handled at the service layer (needs "full" results)
  * @returns Object containing agents array and total count
  */
 export async function findByCompetition(
   competitionId: string,
-  params: CompetitionAgentsParams,
+  params: AgentQueryParams,
+  isComputedSort: boolean = false,
```
dtbuchholz, line 207: for now, we expect the service layer to tell us if it needs "all" of the database results via `isComputedSort`. if true, then we dont apply the limit/offset, and leave it to the service layer to handle.

oh, and since the service layer "splits" sort fields and only passes database-related sort columns to the repository, it allows the `getSort` method to work w/o issues.

apps/api/src/lib/sort.ts
```diff
@@ -0,0 +1,111 @@
+import { ZodEffects, ZodOptional, ZodString, z } from "zod";
+
+/**
+ * Build a Zod schema that accepts
+ *   "field" | "-field"
+ * for any field name in the supplied list.
+ * @param fields - The list of fields to accept
+ * @returns A Zod schema that accepts "field" | "-field" for any field name in the supplied list
+ */
+export function makeSortFieldSchema<const FIELDS extends readonly string[]>(
```
dtbuchholz, line 10: this basically just lets us pass a `["field"]`, and the result will transform it to `["field","-field"]` so that zod can validate the different sort asc/desc values.

there are _many_ ways to do this with zod, though.

apps/api/src/lib/sort.ts
```diff
@@ -0,0 +1,111 @@
...
+ * at the database level or post-processing level
+ * @template TDbSort Database-level sort fields type
+ * @template TPostProcessSort Post-processing sort fields type
+ * @param sortField The sort field to categorize
+ * @param dbSortSchema Zod schema for database-level sort fields (can use defaults)
+ * @param computeSortSchema Zod schema for post-processing sort fields (can use defaults)
+ * @returns Object with dbSort and postProcessSort fields
+ */
+// TODO: this considers only one sort field at a time
+export function splitSortField<
```
dtbuchholz, line 40: i figured a generic "split the db sort vs compute sort" fields could be useful. you pass the sort field, the destination db sort schema, and the post-query computed sort schema. the output will just split the field into two groups.

note: for now, i ignored multi-value sorting.

apps/api/src/lib/sort.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/**
+ * Helper function to apply post-processing sorting to an array of objects
+ * @template T The type of objects being sorted
+ * @param items Array of items to sort
+ * @param sortField Sort field string (e.g., "name", "-createdAt")
+ * @returns void (sorts in place)
+ */
+export function applySorting<T extends Record<string, unknown>>(
```
dtbuchholz, line 74: helper to apply post-query compute sorting logic. 

we should be able to use this in a handful of other methods. there's a bit of overlap with the repository `getSort` usage, but they're used at different levels and requires different types.

apps/api/src/services/agent-manager.service.ts
```diff
@@ -1384,10 +1384,26 @@ export class AgentManager {
...
-            const aName = a.agents?.[0]?.name || "";
-            const bName = b.agents?.[0]?.name || "";
-            comparison = aName.localeCompare(bName);
+            // Sort by the lexicographically FIRST agent name that belongs to the
+            // authenticated user within each competition. This guarantees a
+            // deterministic "primary" agent name irrespective of the order in
+            // which agents were joined or returned from the database. Without
+            // this, the primary agent could change between requests causing
+            // flaky ordering in the UI and in automated tests.
+            const primaryNameA =
```
dtbuchholz, line 1393: note: this was a bug that led to flaky tests. claude's fix seemed to work.

apps/api/src/types/sort/agent.ts
dtbuchholz, line 1: here's where you'll see the "new" patterns for zod types and sorting. totally up for discussion. we can:
- define fields for db-only sorting, and use `makeSortFieldSchema` to prefix them with `-`
- define fields for compute-only sorting, and use `makeSortFieldSchema`
- create a shared type that combines both of these, which is used in the controller to validate sort fields

i separated this file out from the `types/index.ts`, so there's a bit of duplication. probs easier to just start from scratch and move things out of the barrel file, too.

apps/api/src/types/sort/shared.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from "zod";
+
+/**
+ * Shared paging schema for all endpoints
+ */
+export const PagingSchema = z.object({
```
dtbuchholz, line 6: and this is another dupe from the `types/index.ts` that i recreated

apps/api/src/database/repositories/agent-repository.ts
```diff
@@ -159,21 +157,20 @@ export async function findAgentCompetitions(
...
     }
 
     // For computed sorting, we'll need to get all results and sort at service layer
     // So we don't apply limit/offset here if sorting by computed fields
     if (!isComputedSort) {
-      query = query.limit(params.limit).offset(params.offset);
+      query = query.limit(limit).offset(offset);
     }
 
     const results = await query;
```
mzkrasner, line 176: Like you mentioned - these changes are just an interim solution and eventually we will replace what's happening here (fetching all agents with `const results = await query` instead of applying pagination at the DB level

apps/api/src/services/competition-manager.service.ts
```diff
@@ -327,6 +338,90 @@ export class CompetitionManager {
...
+        };
+      }),
+    );
+
+    // Apply post-processing sorting, if needed
+    if (isComputedSort) {
+      applySorting(competitionAgents, computedSort);
+    }
+
+    return {
```
mzkrasner, line 425: I think this would return all agents when computed sorting is used - if you add in:

```
test("PAGINATION BUG: computed sorting ignores pagination limits", async () => {
    // Setup admin client
    const adminClient = createTestClient();

    // Register 6 agents to test pagination
    const agents = [];
    for (let i = 1; i <= 6; i++) {
      const { agent } = await registerUserAndAgentAndGetClient({
        adminApiKey,
        agentName: `Pagination Test Agent ${i}`,
      });
      agents.push(agent);
    }

    // Start a competition with all agents
    const competitionName = `Pagination Bug Test Competition ${Date.now()}`;
    const startResult = await startTestCompetition(
      adminClient,
      competitionName,
      agents.map((a) => a.id),
    );
    const competitionId = startResult.competition.id;

    // Create a client for testing
    const { client } = await registerUserAndAgentAndGetClient({
      adminApiKey,
      agentName: "Pagination Bug Test Client Agent",
    });

    // Test 1: Database sorting (works correctly)
    console.log("Testing database sorting pagination (should work correctly)...");
    const dbSortResponse = (await client.getCompetitionAgents(competitionId, {
      sort: "name",  // Database field - pagination should work
      limit: 3,
      offset: 0,
    })) as CompetitionAgentsResponse;

    expect(dbSortResponse.success).toBe(true);
    expect(dbSortResponse.agents.length).toBe(3); // ✅ Should respect limit
    expect(dbSortResponse.pagination.limit).toBe(3);
    expect(dbSortResponse.pagination.offset).toBe(0);
    expect(dbSortResponse.pagination.total).toBe(6);
    expect(dbSortResponse.pagination.hasMore).toBe(true);

    console.log(`✅ DB sorting: Got ${dbSortResponse.agents.length} agents (expected 3)`);

    // Test 2: Computed sorting (BROKEN - returns all agents)
    console.log("Testing computed sorting pagination (THIS IS THE BUG)...");
    const computedSortResponse = (await client.getCompetitionAgents(competitionId, {
      sort: "position",  // Computed field - pagination is BROKEN
      limit: 3,
      offset: 0,
    })) as CompetitionAgentsResponse;

    expect(computedSortResponse.success).toBe(true);

    // 🐛 THE BUG: This should be 3, but it returns ALL 6 agents
    console.log(`🐛 COMPUTED sorting: Got ${computedSortResponse.agents.length} agents (expected 3, but got all agents!)`);
    console.log(`🐛 Pagination metadata claims limit=${computedSortResponse.pagination.limit}, but actual response has ${computedSortResponse.agents.length} agents`);

    // These assertions expect CORRECT behavior (will FAIL, demonstrating the bug)
    expect(computedSortResponse.agents.length).toBe(3); // ✅ SHOULD return only 3 agents (will fail because bug returns 6)
    expect(computedSortResponse.pagination.limit).toBe(3); // ✅ Metadata is correct
    expect(computedSortResponse.pagination.offset).toBe(0); // ✅ Metadata is correct  
    expect(computedSortResponse.pagination.total).toBe(6); // ✅ Metadata is correct
    expect(computedSortResponse.pagination.hasMore).toBe(true); // ✅ Should be true since 3 < 6

    // Test 3: Try different computed fields to confirm the bug affects all computed sorting
    const testFields = ["score", "pnl", "pnlPercent", "change24h", "change24hPercent", "voteCount"];

    for (const field of testFields) {
      console.log(`Testing computed field: ${field}`);
      const response = (await client.getCompetitionAgents(competitionId, {
        sort: field,
        limit: 2,  // Request only 2 agents
        offset: 0,
      })) as CompetitionAgentsResponse;

      expect(response.success).toBe(true);
      console.log(`🐛 Field ${field}: Got ${response.agents.length} agents (expected 2)`);

      // Expect CORRECT behavior (will fail due to bug)
      expect(response.agents.length).toBe(2); // ✅ SHOULD return only 2 agents (will fail because bug returns 6)
      expect(response.pagination.limit).toBe(2); // ✅ Metadata claims limit=2
    }

    // Test 4: Demonstrate that offset is also ignored
    console.log("Testing offset with computed sorting...");
    const offsetResponse = (await client.getCompetitionAgents(competitionId, {
      sort: "position",
      limit: 2,
      offset: 3,  // Should skip first 3 agents
    })) as CompetitionAgentsResponse;

    expect(offsetResponse.success).toBe(true);
    console.log(`🐛 With offset=3, limit=2: Got ${offsetResponse.agents.length} agents (expected 2)`);

    // Expect CORRECT behavior (will fail due to bug)
    expect(offsetResponse.agents.length).toBe(2); // ✅ SHOULD return only 2 agents (will fail because bug returns 6)
    expect(offsetResponse.pagination.offset).toBe(3); // ✅ Metadata claims offset=3
    expect(offsetResponse.pagination.limit).toBe(2); // ✅ Metadata claims limit=2

    console.log("🐛 PAGINATION BUG CONFIRMED: Computed sorting returns all agents regardless of limit/offset");
  });
  ```

apps/api/e2e/tests/competition.test.ts
mzkrasner: Nice. Left a related comment, but would be good to have a test case that tests pagination with computed sorting

mzkrasner (commented):
Look great overall! Just left a few comments that ultimately point to the same thing (pagination with computed sorting)

apps/api/src/services/competition-manager.service.ts
dtbuchholz: ahh nice catch. and thanks for the snippet...just fixed this

dtbuchholz:
> I'd vote to put a somewhat high priority to replacing the full scan plus computed sort and pagination with a rubust solution.

sg. i made a note in https://github.com/recallnet/js-recall/pull/643 to use this PR as a reference

Issue #640, opened 2025-06-17
[BUG] Fix "Show more" logic in Global Leaderboard UI


dtbuchholz:
**Bug Description**
Our existing pagination logic in the comps UI has a "show more" button at the global leaderboard page. If you click "show more", it'll cycle to the next 10 results but hide the original set. You can't go back and retrieve them w/o reloading the page.

Note: Our future leaderboard designs via https://github.com/recallnet/js-recall/issues/572 will implement "Page 1, 2, 3,..." rendering. The "Show more" is an older design, but coincidentally, the bug itself better aligns with the future-state logic.

**Steps to Reproduce**

- Global Leaderboard page: clicking "show more" loads the next batch of 10 into the context, and flushes the original 10 away. They're irrecoverable unless you refresh the page.

**Proposed solution**
- _Only_ on the Global Leaderboard page, we can more easily fix this by using the new designs. 
- Ensure the pagination aligns correctly with the batch of 10 that gets loaded
   - i.e., you should be able to select page 1 -> page 2 -> page 1

**Additional context**
- See [Figma](https://www.figma.com/design/WqQqcsg1g66EdMjjM8yfIM/Competitions-App---v.1.0--06-25-?node-id=4-11575&t=JC93n4yUrnh4VuB5-4)
- We _only_ need to implement the pagination bug fix / "Page 1, 2, etc" logic, and the rest of the features in https://github.com/recallnet/js-recall/issues/572 can be implemented later

<img width="249" alt="Image" src="https://github.com/user-attachments/assets/14572303-0031-4237-88aa-4b2e693b2e15" />

PR #649, opened 2025-06-18
feat: pagination component

Rcontre360:
#640

### DIFF
--- a/apps/comps/components/leaderboard-table.tsx
+++ b/apps/comps/components/leaderboard-table.tsx
@@ -39,7 +39,10 @@ export function LeaderboardTable({
   };
   onPageChange: (page: number) => void;
 }) {
-  const page = pagination.offset / pagination.limit + 1;
+  const page =
+    pagination.limit > 0
+      ? Math.floor(pagination.offset / pagination.limit) + 1
+      : 1;
 
   return (
     <div className="flex flex-col items-end">


apps/comps/components/leaderboard-table.tsx
```diff
@@ -18,24 +16,33 @@ import { cn } from "@recallnet/ui2/lib/utils";
...
-  loaded,
   handleSortChange,
   sortState,
-  onLoadMore,
-  hasMore = false,
+  onPageChange,
+  pagination = {
+    total: 0,
+    limit: 0,
+    offset: 0,
```
dtbuchholz, line 29: note: the `pagination` response also includes `hasMore`, in case it's heplful

apps/comps/components/leaderboard-table.tsx
```diff
@@ -18,24 +16,33 @@ import { cn } from "@recallnet/ui2/lib/utils";
...
-  onLoadMore: () => void;
-  hasMore?: boolean;
+  pagination?: {
+    total: number;
+    limit: number;
+    offset: number;
+  };
+  onPageChange: (page: number) => void;
 }) {
+  const page = pagination.offset / pagination.limit + 1;
```
dtbuchholz: here's the fix for a `NaN` showing up as a pagination option:
Screenshot 2025-06-18 at 1 16 43 PM
```ts
  const page =
    pagination.limit > 0
      ? Math.floor(pagination.offset / pagination.limit) + 1
      : 1;
```
also, if you hover over the right arrow while you're on page one, it doesnt use a "pointer" but keeps it as a default cursor value.

dtbuchholz (approved):

Issue #572, opened 2025-06-11
Small changes to Leaderboard UI


stbrody:
Based on design updates from https://github.com/recallnet/js-recall/issues/410

dtbuchholz:
note: i added some DoD callouts...this is also good to go

PR #656, opened 2025-06-18
Rafael/leaderboard v1.2

Rcontre360:
#572

### DIFF
--- a/apps/comps/components/leaderboard-table/index.tsx
+++ b/apps/comps/components/leaderboard-table/index.tsx
@@ -133,7 +133,7 @@ export function LeaderboardTable({
                           <span>3rd</span>
                         </div>
                       ) : (
-                        <div className="ml-7 flex items-center justify-center rounded bg-gray-800 p-2">
+                        <div className="flex w-20 items-center justify-center rounded bg-gray-800 p-2">
                           {agent.rank}
                         </div>
                       )}
--- a/apps/comps/components/leaderboard-table/index.tsx
+++ b/apps/comps/components/leaderboard-table/index.tsx
@@ -1,4 +1,4 @@
-import { AwardIcon, ExternalLink, Trophy } from "lucide-react";
+import {AwardIcon, ExternalLink, Trophy} from "lucide-react";
 import Image from "next/image";
 import Link from "next/link";
 
@@ -12,11 +12,11 @@ import {
   TableHeader,
   TableRow,
 } from "@recallnet/ui2/components/table";
-import { cn } from "@recallnet/ui2/lib/utils";
+import {cn} from "@recallnet/ui2/lib/utils";
 
-import { LeaderboardAgent } from "@/types/agent";
+import {LeaderboardAgent} from "@/types/agent";
 
-import { Pagination } from "./pagination/index";
+import {Pagination} from "@/components/pagination/index";
 
 export function LeaderboardTable({
   agents,
--- a/apps/comps/components/leaderboard-table/index.tsx
+++ b/apps/comps/components/leaderboard-table/index.tsx
@@ -133,7 +133,7 @@ export function LeaderboardTable({
                           <span>3rd</span>
                         </div>
                       ) : (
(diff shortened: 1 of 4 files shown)


apps/comps/components/leaderboard-table/index.tsx
```diff
@@ -48,14 +47,20 @@ export function LeaderboardTable({
...
             <SortableTableHeader
-              className="pl-10 text-white"
+              className="mr-15 pl-20 text-white"
               onToggleSort={() => handleSortChange("rank")}
               sortState={sortState["rank"]}
             >
               Rank
             </SortableTableHeader>
+            <SortableTableHeader
+              onToggleSort={() => handleSortChange("name")}
```
dtbuchholz: should probably be: `handleSortChange("score")}` and `sortState={sortState["score"]}`

dtbuchholz:
@Rcontre360 lookin good. a few things:
1. for the `AgentRank` column, we can round to the nearest whole integer
2. the `Agent` column values aren't aligned. the point above should fix this, but keep an eye out in case additional styles needed to be applied to align the column
3. along the same lines, the `Competitions` and `Votes` column values aren't aligned horizontally, but this is also probably fixed by point (1)
Screenshot 2025-06-18 at 4 26 08 PM
4. if i click the sorting on the `AgentRank` or `Agent` columns, the sort toggle highlights _both_ of them. the inline comment i made should fix this.
Screenshot 2025-06-18 at 4 31 07 PM

5. unrelated to this PR—i noticed the font in `Volume Traded` (the top summary section) doesn't match the other two boxes. maybe we just fix that in this PR, too, since it's a tiny but related change?

apps/comps/components/leaderboard-table/index.tsx
```diff
@@ -44,143 +46,157 @@ export function LeaderboardTable({
...
+                          {agent.rank}
+                        </div>
+                      )}
+                    </TableCell>
 
-              <TableCell className="flex items-center justify-start pr-10 text-gray-500">
-                {agent.score}
-              </TableCell>
+                    <TableCell className="flex items-center justify-start pr-10 text-gray-500">
+                      <BigNumberDisplay
```
dtbuchholz, line 143: for the AgentRank, we want to display it as the whole integer/rounded value. but, we _do not_ want to use `BigNumberDisplay`. 

the scores will be anywhere from `0` to some number in the low thousands like `1500`. we want to display the full number and truncate decimals so that it's `0` or `100` or `1543`
Screenshot 2025-06-18 at 5 56 37 PM

apps/comps/components/leaderboard-table/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+                        <div
+                          className={cn(
+                            "flex w-20 items-center justify-center gap-1 rounded bg-[#1A0E05] p-2 text-[#C76E29]",
+                          )}
+                        >
+                          <AwardIcon size={17} />
+                          <span>3rd</span>
+                        </div>
+                      ) : (
+                        <div className="ml-7 flex items-center justify-center rounded bg-gray-800 p-2">
```
dtbuchholz:
```suggestion
                        <div className="flex w-20 items-center justify-center rounded bg-gray-800 p-2">
```

dtbuchholz: the quicket fix is to do what the competitions leaderboard page does and keep the div width the same for all numbers. this doesnt perfectly match the design, but it'll fix the spacing issue that i'm still seeing

dtbuchholz: - without this change
Screenshot 2025-06-19 at 9 54 40 AM

- with the recommended change
Screenshot 2025-06-19 at 9 54 12 AM

dtbuchholz (approved):
@Rcontre360 once you accept/commit that suggested fix i just made, this should be good to go!

dtbuchholz:
lgtm

Issue #645, opened 2025-06-18
[BUG] Fix hydration errors due to Jotai


dtbuchholz:
**Bug Description**
There's a hydration error in the comps app. The problem is with Jotai's `useAtom`.

The `getOnInit` function makes the `localStorage` read synchronous so that user and auth data from `localStorage` are already available on the first render. So any component that uses this atom through a hook will see user as null during pre-rendering on the server (since there's no `localStorage` there), but on the first client-side render, after reading from `localStorage`, user is not null (if the user is logged in), which causes the mismatch.
```ts
export const userAtom = atomWithStorage<UserStorage>(
  "user",
  {
    user: null,
    status: "unauthenticated",
  },
  undefined,
  {
    getOnInit: true,
  },
);
```

**Steps to Reproduce**
Startup the `apps/comps`, and you'll see the error on load.

**Screenshots**
<img width="1123" alt="Image" src="https://github.com/user-attachments/assets/cc17c4a8-ebfe-41cb-9380-a4a7d991f3e5" />

PR #657, opened 2025-06-18
fix(comps): prevent hydration error on LS read

ramirotw:
Fixes https://github.com/recallnet/js-recall/issues/645

### DIFF
--- a/apps/comps/hooks/useAuth.ts
+++ b/apps/comps/hooks/useAuth.ts
@@ -1,7 +1,8 @@
 import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
+import { useIsClient } from "@uidotdev/usehooks";
 import { useAtom } from "jotai";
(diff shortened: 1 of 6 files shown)


apps/comps/hooks/useAuth.ts
```diff
@@ -121,10 +143,25 @@ export const useUserSession = (): UserSessionState => {
...
+    isClient,
+    authState.user,
     isAuthenticated,
     isProfileUpdated,
-    isLoading: profileIsLoading,
-  };
+    profileIsLoading,
+  ]);
+
+  return sessionState;
```
ramirotw, line 166: This solves the problem for all calls to useUserSession, although I'm not completely convinced by it. I also considered persisting the auth/user state in a cookie, that would allow SSR to access it, although it could introduce other issues. Plus, we’d need to modify which user fields are persisted, since it wouldn’t be feasible to send all of them in a cookie

Issue #664, opened 2025-06-19
[BUG] Fix `/competitions/:id/agents`sort by portfolio value and wrong default


dtbuchholz:
**Bug Description**
- Portfolio value is not included in sorting
- Default agent `position` in competition is not applied

PR #665, opened 2025-06-19
fix(comps): comps leaderboard sort

dtbuchholz:
closes https://github.com/recallnet/js-recall/issues/664

### DIFF
--- a/apps/comps/components/agents-table/index.tsx
+++ b/apps/comps/components/agents-table/index.tsx
@@ -104,7 +104,7 @@ export const AgentsTable: React.FC<AgentsTableProps> = ({
   const columns = useMemo<ColumnDef<AgentCompetition>[]>(
     () => [
       {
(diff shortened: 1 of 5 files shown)


apps/comps/components/agents-table/index.tsx
```diff
@@ -203,7 +203,8 @@ export const AgentsTable: React.FC<AgentsTableProps> = ({
         size: 100,
       },
       {
-        id: "votes",
+        id: "voteCount",
+        accessorKey: "voteCount",
```
dtbuchholz, line 207: note: the API sorting handlers don't have any aliasing set up, so in this case, it needs to match the exact value in the API response (which is `voteCount`).

there's inconsistency across APIs that we want to (eventually) clean up and make this less of a headache

ramirotw (approved):
tested locally, works great

Issue #659, opened 2025-06-19
[IMPL] Trophy component frontend updates


dtbuchholz:
## Definition of done

- [x] Trophies for 1st, 2nd, 3rd place should have a special border colors
- [x] Trophies for 1st, 2nd, 3rd place should have special "shine" behavior
- [x] All non-top-3 competitions should have a normal trophy / border color, and do not shine
- [x] Hover over trophy badge should produce a tooltip for the placement + competition name
- [x] All trophies should be clickable -> go to the Competition Page they're corresponding to
- [x] Badges should be clustered in order (gold > silver > bronze > normal) and by "earned" date (least to most recent)

# Notes

See [FIgma](https://www.figma.com/design/WqQqcsg1g66EdMjjM8yfIM/Competitions-App---v.1.0--06-25-?node-id=647-41087&t=BRG0JgtC3Ck67gU9-0). There's a lot more detail for the various features above.

Issue #573, opened 2025-06-11
Agent profile page updates (design v1.3)


stbrody:
## Definition of Done

- [x] Fix the `/agents/:id` route's table and sorting/filtering (the table headers are stacked into a single column)
- [x] Change Hover Text: "API key" to "Agent API Key"
- [x] Change Hover Text: "Wallet Address" to "Agent Wallet Address"
- [x] Change to "Developed by" subtag => no longer appears clickable
- [x] Agent Skills are now editable (open modal with selects)
- [x] Remove "Hide" option from Agent Wallet Address
- [x] For Agent Wallet Address, changed supporting button: Copy Wallet Address instead of opening External Link
- [x] Add Verified Badge to Agents
   - If verified: On hover, show tooltip "Verified agent"
   - If not verified: On hover, show tooltip "Learn how to verify", clickable link 

Nice to haves:
- [x] Create "share" social modals that pop up when you click on an agent profile's share button
- [x] Include clickable "trophies" that bring you to the competition page. Note: it needs https://github.com/recallnet/js-recall/issues/613, but you can expect a nullable `trophies` field. The objects will look like `trophies: [{"competitionId":"..."}]`

## Notes

- Depends on design updates from https://github.com/recallnet/js-recall/issues/411
- Impacts public agent profiles and components
- Note: #571 handles the user profile page update and adds the "verified agent" feature here, so just a heads up in case of any overlap

PR #667, opened 2025-06-19
[WIP] Rafael/agent profile v1.3

Rcontre360:


Rcontre360:
closes #659 and #573

dtbuchholz:
## public agent profile

1. the agent's skills button _always_ show the "edit" button for agents that are _not_ owned by the user; this should never happen on the public agent profile page
2. the agent verification badge isnt centered vertically
3. "Proven Skills" should say "Agent Skills"
4. if you hover over a verified agent badge, it needs to show "Verified Agent". the designs don't account for the opposite case, so let's have the tooltip show "Agent Not Verified" if the agent is unverified
5. it should read as `Developed by [<name>]` with the brackets enclosing the name; also, the spacing looks odd
6. the designs didnt account for this—but while we're at it, we should update "ELO" to "AgentRank"
Screenshot 2025-06-20 at 4 46 02 PM

## private agent profile page

1. let's make the text say `Learn how to verify your agent`, and we should make this link and a noreferer / external redirect to the `socialLinks.docs.url` (like [here](https://github.com/recallnet/js-recall/blob/010bfe67684d97ee11c84cd7c9af33eb537566a0/apps/comps/components/agent-created.tsx#L74))
Screenshot 2025-06-20 at 4 54 00 PM

2. similar to the create agent flow, we should validate image URLs (something like [here](https://github.com/recallnet/js-recall/pull/689/files#diff-acdf6942d473104f821a60915ea738e502a13dbee0fa321d6f46841b9d4fe202R37) or [here](https://github.com/recallnet/js-recall/pull/689/files#diff-0a2788d5887bce016530c354d68fa8419e5413c8d627d35d73d75de2eec5e95fR27)) to prevent the app from breaking or trying to save something that wont be rendered properly
Screenshot 2025-06-20 at 5 34 02 PM

## trophies

i realize the issue itself doesnt capture all of these points in detail, and the figma has the more in-depth requirements:

1. the `/agents:id` page is still using the old trophy design in the table. note: the designs didnt account for this, but since this PR is already updating the trophies, we might as well render this table properly.
Screenshot 2025-06-20 at 5 24 11 PM

2. the trophy expands outside the bounds and gets cut off at the edges by the surrounding div
3. the trophy tooltip should show on hover, and there should be an "increased shine" for 1st/2nd/3rd on hover

https://github.com/user-attachments/assets/c8f5339c-f6d0-413b-997d-67c61546dbf5

apps/api/src/types/index.ts
```diff
@@ -737,6 +737,14 @@ export const UpdateAgentProfileBodySchema = z
       .min(1, { message: "Description must be at least 1 character" })
       .optional(),
     imageUrl: z.url("Invalid image URL format").optional(),
+    metadata: z
```
dtbuchholz: this change isn't needed. if you remove it, a user is able to successfully update the agent skills w/o issue. (i tested this locally.)

PR #678, opened 2025-06-20
Add ObjectIndexService

islamaliev:
This PR adds the ObjectIndexService to populate the object_index table with data from various sources across the trading simulator. The primary focus is storing trade data with embedded chain-of-thought reasoning, while also supporting other data types like agent rank history, competition leaderboards, portfolio snapshots, and agent ranks.

The implementation provides three ways to populate the object_index table: a CLI script for manual population with configurable options, an admin API endpoint for on-demand synchronization, and automatic population when competitions end. The service processes data in batches for efficiency and stores complete records as JSON, maintaining a generic and data-agnostic approach.

A new POST /api/admin/object-index/sync endpoint has been added to the admin routes, allowing administrators to manually trigger data synchronization with optional competition filtering and data type selection. The service is fully integrated into the application's service registry and competition lifecycle.

### DIFF
--- a/apps/api/src/controllers/admin.controller.ts
+++ b/apps/api/src/controllers/admin.controller.ts
@@ -682,11 +682,74 @@ export function makeAdminController(services: ServiceRegistry) {
         const leaderboard =
           await services.competitionManager.getLeaderboard(competitionId);
 
+        // Populate object_index with competition data
+        try {
+          await services.objectIndexService.populateTrades(competitionId);
+          await services.objectIndexService.populateAgentRankHistory(competitionId);
+          await services.objectIndexService.populateCompetitionsLeaderboard(competitionId);
+          console.log(`Successfully populated object_index for competition ${competitionId}`);
+        } catch (error) {
+          console.error(`Failed to populate object_index for competition ${competitionId}:`, error);
+          // Don't fail the request if object_index population fails
+        }
+
         // Return the ended competition with leaderboard
         res.status(200).json({
           success: true,
           competition: endedCompetition,
           leaderboard,
+        });
+      } catch (error) {
+        next(error);
+      }
+    },
+
+    /**
+     * Manually trigger object index population
+     * @param req Express request
+     * @param res Express response
+     * @param next Express next function
+     */
+    async syncObjectIndex(req: Request, res: Response, next: NextFunction) {
+      try {
+        const { competitionId, dataTypes } = req.body;
+
+        const typesToSync = dataTypes || ['trade', 'agent_rank_history', 'competitions_leaderboard'];
+        
+        console.log(`Starting object index sync for types: ${typesToSync.join(', ')}`);
+        
+        for (const dataType of typesToSync) {
+          try {
+            switch (dataType) {
+              case 'trade':
+                await services.objectIndexService.populateTrades(competitionId);
+                break;
+              case 'agent_rank_history':
+                await services.objectIndexService.populateAgentRankHistory(competitionId);
+                break;
+              case 'competitions_leaderboard':
+                await services.objectIndexService.populateCompetitionsLeaderboard(competitionId);
+                break;
+              case 'portfolio_snapshot':
+                await services.objectIndexService.populatePortfolioSnapshots(competitionId);
+                break;
+              case 'agent_rank':
+                await services.objectIndexService.populateAgentRank();
+                break;
+              default:
+                console.warn(`Unknown data type: ${dataType}`);
+            }
+          } catch (error) {
+            console.error(`Error syncing ${dataType}:`, error);
+            throw error;
+          }
+        }
+
+        res.status(200).json({
+          success: true,
+          message: 'Object index sync initiated',
+          dataTypes: typesToSync,
+          competitionId: competitionId || 'all'
         });
       } catch (error) {
         next(error);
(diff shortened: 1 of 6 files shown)


apps/api/src/services/objectIndex.service.ts
dtbuchholz, line 1: i think all of this logic should exist in the `src/database/repositories` as a new file called `object-index.services.ts`. the app is basically split up like this:
- controllers handle API requests
- services are the middleman b/w controllers and the database, and 
- repositories interact with the database

we should rename this file to a hyphenated `object-index.service.ts` just to keep things consistent with the rest of the codebase. then, the service will, more or less, just mediate calls between the controllers and repositories.

apps/api/src/services/objectIndex.service.ts
```diff
@@ -0,0 +1,301 @@
...
+
+    const tradesToSync = await query;
+    
+    if (tradesToSync.length === 0) {
+      console.log('No trades to sync');
+      return;
+    }
+
+    const entries = tradesToSync.map(trade => ({
+      id: crypto.randomUUID(),
```
dtbuchholz: we should `import { v4 as uuidv4 } from "uuid";` and use `uuidv4()` instead of `crypto.randomUUID()`

apps/api/src/scripts/populateObjectIndex.ts
dtbuchholz, line 1: nit: rename to a hyphenated format so that it's consistent with the rest of the repo

apps/api/src/scripts/populateObjectIndex.ts
```diff
@@ -0,0 +1,91 @@
...
+      options.batchSize = parseInt(args[i + 1]!, 10);
+      i++;
+    }
+  }
+
+  return options;
+}
+
+// Execute the script if run directly
+const envOptions: PopulateOptions = {
```
dtbuchholz: since this is a one-off script, i think we should just remove the env vars here.

the script already allows these values to be passed as args above. if they're not provided, we can set defaults like you've done here.

(i want to avoid adding env vars for anything that isnt part of the core app logic)

apps/api/src/controllers/admin.controller.ts
```diff
@@ -693,6 +704,58 @@ export function makeAdminController(services: ServiceRegistry) {
...
 
+    /**
+     * Manually trigger object index population
+     * @param req Express request
+     * @param res Express response
+     * @param next Express next function
+     */
+    async syncObjectIndex(req: Request, res: Response, next: NextFunction) {
+      try {
+        const { competitionId, dataTypes } = req.body;
```
dtbuchholz, line 728: given you've outlined some concrete `dataType` values, we should codify these values as zod types and postgres enum values.

for example, something like this:
```ts
// src/types/index.ts
export const SYNC_DATA_TYPE_VALUES = [
  "trade",
  "agent_rank_history",
  "agent_rank",
  "competitions_leaderboard",
  "portfolio_snapshot"
] as const;
export const SYNC_DATA_TYPE = {
  TRADE: "trade",
  // ...
} as const;
export const SyncDataTypeSchema = z.enum(SYNC_DATA_TYPE_VALUES);
```
and in the postgres definition:
```ts
// src/database/schemas/syncing/defs.ts
export const syncDataType = pgEnum("sync_data_type", SYNC_DATA_TYPE_VALUES);

export const objectIndex = pgTable(
  "object_index",
  {
     // ...
    dataType: syncDataType("data_type").notNull(),
  }
)
// ...
```
and in the controller logic, we should parse this with zod. there are better examples [here](https://github.com/recallnet/js-recall/blob/main/apps/api/src/controllers/request-helpers.ts) or scattered through the other controllers with dedicated helper methods.
```ts
// src/controllers/admin.controller.ts`
async syncObjectIndex(...) {
   // ...
  const { competitionId, dataTypes } = req.body;
  const competitionId = ensureUuid(competitionId);
  const { success, data, error } = SyncDataTypeSchema.safeParse(dataTypes)
  // handle error etc
  // ...
}
```

fwiw, this is a pattern that isnt implemented across all of our APIs, yet. also, my example is probably a bit imperfect since it looks like the API can accept multiple `dataTypes` values at one, but hopefully, what i described makes sense.

apps/api/src/services/objectIndex.service.ts
islamaliev: done

apps/api/src/services/objectIndex.service.ts
islamaliev: done

apps/api/src/scripts/populateObjectIndex.ts
islamaliev: done

apps/api/src/scripts/populateObjectIndex.ts
islamaliev: done

apps/api/src/controllers/admin.controller.ts
islamaliev: done

dtbuchholz (commented):
@islamaliev this lgtm! however, one more thing: i realized there aren't any tests for this logic; sorry i didn't catch it sooner.

here's what you'll need to do: https://github.com/recallnet/js-recall/issues/717

given we're doing a lot of infra deployment activities this week, i want to make sure these are written _before_ we merge this PR...just in case something unexpected comes up. fwiw, i manually tested the logic locally, and it all seems to work as expected!

Issue #646, opened 2025-06-18
[BUG] User voting info shown for agent response


dtbuchholz:
**Bug Description**
The `/competitions:id` response includes user information when called via a curl request that includes bearer authorization + agent API key.

**Steps to Reproduce**
For an active comp, make a request like this:
```sh
curl "http://localhost:3000/api/competitions/$COMPETITION_ID" -H "Authorization: Bearer $AGENT_API_KEY" | jq
```
and the response incorrectly includes `userVotingInfo`:
```json
{
  "success": true,
  "competition": {
    // ...
    "votingEnabled": true,
    "userVotingInfo": {
      "canVote": true,
      "info": {
        "hasVoted": false
      }
    }
  }
}
```

deepanshu2711:
Hi @dtbuchholz  I want to work n this issue assign this to me if possible

PR #681, opened 2025-06-20
fix(api): add test for bug in hasVoted value calc

joewagner:
wip: fixing issue with voting computed values and voting start stop dates.

fixes: https://github.com/recallnet/js-recall/issues/646

### DIFF
--- a/apps/api/e2e/tests/voting.test.ts
+++ b/apps/api/e2e/tests/voting.test.ts
@@ -542,6 +542,59 @@ describe("Voting API", () => {
         (stateResponse as VotingStateResponse).votingState.info.hasVoted,
       ).toBe(true);
     });
+
+    test("should maintain hasVoted=true after competition closes", async () => {
+      // Setup admin client
+      const adminClient = createTestClient();
+      await adminClient.loginAsAdmin(adminApiKey);
+
+      // Register agents
+      const { agent: agent1 } = await registerUserAndAgentAndGetClient({
+        adminApiKey,
+        agentName: "Agent Romeo",
+      });
+      const { agent: agent2 } = await registerUserAndAgentAndGetClient({
+        adminApiKey,
+        agentName: "Agent Sierra",
+      });
+
+      // Start a competition
+      const competitionName = `Voting Persistence Test ${Date.now()}`;
+      const competitionResponse = await startTestCompetition(
+        adminClient,
+        competitionName,
+        [agent1.id, agent2.id],
+      );
+      const competition = competitionResponse.competition;
+
+      // Create SIWE authenticated user
+      const { client: userClient } = await createSiweAuthenticatedClient({
+        adminApiKey,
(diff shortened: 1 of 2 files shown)


apps/api/drizzle/meta/_journal.json
```diff
@@ -108,4 +108,4 @@
       "breakpoints": true
     }
   ]
-}
```
joewagner, line 111: The files in the drizzle directory did not pass formatting.  I updated them with `pnpm format`, but they don't have anything to do with this PR.

apps/api/e2e/tests/voting.test.ts
```diff
@@ -66,6 +66,11 @@ describe("Voting API", () => {
       const competitionResponse = await createTestCompetition(
         adminClient,
         competitionName,
+        undefined, // description
+        undefined, // externalUrl
+        undefined, // imageUrl
+        undefined, // competition type
+        new Date().toISOString(), // start date to allow voting
```
joewagner, line 73: This test is asserting that voting works, so we need to enable it for the competition.  You'll see this type of change in a few places, as well as a change to the test helper to support setting a voting start date.

apps/api/e2e/tests/voting.test.ts
```diff
@@ -542,6 +552,59 @@ describe("Voting API", () => {
         (stateResponse as VotingStateResponse).votingState.info.hasVoted,
       ).toBe(true);
     });
+
+    test("should maintain hasVoted=true after competition closes", async () => {
```
joewagner, line 556: This wasn't called out explicitly in the ticket #646, but the value of `hasVoted` was always set to false when a competition status equaled "ended".  This seemed related so I'm fixing here.

apps/api/src/services/vote-manager.service.ts
```diff
@@ -176,19 +176,19 @@ export class VoteManager {
...
+      // Check if user has already voted
+      const userVote = await getUserVoteForCompetition(userId, competitionId);
+      const hasVoted = !!userVote;
+
       // Check if voting is allowed based on competition status
       if (!this.checkCompetitionVotingStatus(competition)) {
         return {
           canVote: false,
           reason: `Competition status does not allow voting (${competition.status})`,
-          info: { hasVoted: false },
```
joewagner, line 184: This is the fix for the hard coded `hasVoted` field.

apps/api/src/services/vote-manager.service.ts
```diff
@@ -338,6 +338,14 @@ export class VoteManager {
   private checkCompetitionVotingDates(competition: SelectCompetition) {
     const now = new Date();
 
+    // If at least one voting date is not set, voting for the comp is disabled
+    if (!competition.votingStartDate && !competition.votingEndDate) {
```
joewagner, line 342: This is the change that disables voting if at least one voting date is not set.

Issue #614, opened 2025-06-13
Create agent / registration updates (design v1.2)


dtbuchholz:
## Definition of Done

- [x] Split registration into three screens:
   - Basic: name (required), description, repo URL, skills
   - Socials: profile pic, email, twitter, telegram
   - API key: mostly, the same as the current confirmation screen but with css (e.g., justify left; show agent card on right) and text sizes
- [x] Remove "Wallet Address" from the agent profile form
- [x] If the "Skills" + "Other skill" is selected, label the form field that appears as "Other Skill"
- [x] Create "lose progress" model that will pop up if you try to exist the registration process on "Basic" or "Socials" screens
- [x] Update to use responsive breakpoints (note: lose the supporting image at >1280 width) 

## Notes

- Depends on https://github.com/recallnet/js-recall/issues/411; see [Figma](https://www.figma.com/design/WqQqcsg1g66EdMjjM8yfIM/Competitions-App---v.1.0--06-25-?node-id=303-6642&p=f&m=dev)
- Affects the `CreateAgentView` page and components

PR #689, opened 2025-06-20
feat(comps): create agent v1.2 changes

ramirotw:
Fixes https://github.com/recallnet/js-recall/issues/614


apps/comps/components/create-agent/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+import { SocialsStep } from "./socials-step";
+import { Steps } from "./steps";
+
+const formSchema = z
+  .object({
+    name: z.string().min(1, "Agent name is required"),
+    imageUrl: asOptionalStringWithoutEmpty(
+      z.string().url({ message: "Must be a valid URL" }),
+    ),
+    repositoryUrl: z.string().url({ message: "Must be a valid URL" }),
```
dtbuchholz: let's make the `repositoryUrl` optional. (this change was buried in some comments.)

dtbuchholz:
@ramirotw this is mostly good to go; just make the repo URL optional.

i _did_ discover a bug in the backend API. if you create an agent with a duplicate name, the server throws a 500—and then you get stuck at the "Submit" screen. we could either:
- make a check on the initial "Basics" screen and ensure we validate the agent's name is unique relative to the user's existing set of agents
- on the submit screen, catch this to-be-implemented error response https://github.com/recallnet/js-recall/issues/691

apps/comps/components/create-agent/index.tsx
ramirotw: Fixed in [fac2220](https://github.com/recallnet/js-recall/pull/689/commits/fac222008b1fc5439e92f4697a8ea7e4136adf33)

ramirotw:
> @ramirotw this is mostly good to go; just make the repo URL optional.
> i did discover a bug in the backend API. if you create an agent with a duplicate name, the server throws a 500—and then you get stuck at the "Submit" screen. we could either:
> make a check on the initial "Basics" screen and ensure we validate the agent's name is unique relative to the user's existing set of agents
on the submit screen, catch this to-be-implemented error response https://github.com/recallnet/js-recall/issues/691

Got it. I improved the error handling so it automatically focuses on the correct input and displays the appropriate error message. In this specific case, I set up the 409 error to target the agent's name, and for other types of errors, a toast notification is shown

apps/comps/components/user-agents/index.tsx
```diff
@@ -27,7 +27,7 @@ import { formatCompactNumber, toOrdinal } from "@/utils/format";
 import { VerificationBadge } from "../verification-badge";
 
 export default function UserAgentsSection() {
-  const { data: agentsData, isLoading } = useUserAgents();
+  const { data: agentsData, isLoading } = useUserAgents({ limit: 100 });
```
ramirotw, line 30: `/user/agents` is paginated and this was only fetching 10 agents. Btw, the openapi spec and the response needs to be updated to include the pagination and sort params.

dtbuchholz (approved):
lgtm, great ideas

PR #690, opened 2025-06-20
fix: apply unauthenticated rate limiting at the ip address level

mzkrasner:
## Summary
Replaces the shared "anonymous" rate limiting pattern with IP-based rate limiting to prevent abuse and improve security for unauthenticated API requests.

### Changes Made

_Core Implementation:_

- Express Configuration: Added app.set('trust proxy', true) to properly extract client IP addresses behind proxies/load balancers
- Rate Limiter Enhancement: Updated rate-limiter.middleware.ts to use ip:${clientIP} instead of "anonymous" for unauthenticated requests
- IP Extraction Function: Added getClientIp() helper function with fallback logic for reliable IP detection

_Testing:_

- New E2E Test: Added comprehensive test for IP-based rate limiting using /api/agents endpoint
- Rate Limit Validation: Test verifies proper rate limiting behavior (30 requests/minute account limit)
- IP Structure Verification: Confirms server logs show ip:127.0.0.1 format for unauthenticated requests

### Technical Details

- Maintains existing authenticated request patterns (no breaking changes)
- Uses Express's req.ip which respects the trust proxy setting
- Fallback logic ensures IP detection works in various deployment scenarios
- Test coverage validates both successful requests and rate limit enforcement


apps/api/src/middleware/rate-limiter.middleware.ts
```diff
@@ -58,10 +58,59 @@ function getRateLimiter(
   return agentLimiters.get(type)!;
 }
 
+/**
+ * Get the client IP address from the request
+ * Handles Cloudflare + Load Balancer setup properly
+ *
+ * Priority order:
```
mzkrasner, line 65: @sam701 will this approach work with our load balancer + cloudflare setup?

sam701: Yes, the general approach should work for the CloudFlare setup. But only for the CloudFlare setup 🙂 , what will be the case anyway.

In general, using IP addresses for rate limiting is not very reliable:
- an attacker can use IPv6 and get millions of IPs for a single machine,
- `x-forwarded-for` can be explicitly set in the request.

I'd suggest to use the user/agent ID for this purpose where it's possible and fallback on the IPs where it's not possible like `/auth` endpoints.

apps/api/src/middleware/rate-limiter.middleware.ts
```diff
@@ -58,10 +58,59 @@ function getRateLimiter(
...
+ * 2. First IP in X-Forwarded-For chain (excluding private IPs)
+ * 3. Express req.ip (fallback)
+ *
+ * @param req - Express request object
+ * @returns Client IP address or "unknown" if unable to determine
+ */
+function getClientIp(req: Request): string {
+  // Priority 1: Cloudflare connecting IP (if from CF)
+  // CF-Ray header indicates the request came through Cloudflare
+  if (req.headers["cf-ray"] && req.headers["cf-connecting-ip"]) {
```
sam701, line 76: 👍 this should work, because `cf-connecting-ip` will be correctly set by CloudFlare.

apps/api/src/middleware/rate-limiter.middleware.ts
```diff
@@ -58,10 +58,59 @@ function getRateLimiter(
...
+    const cfIp = req.headers["cf-connecting-ip"] as string;
+    if (cfIp && cfIp !== "unknown") {
+      return cfIp;
+    }
+  }
+
+  // Priority 2: First IP in X-Forwarded-For (excluding private ranges)
+  const xForwardedFor = req.headers["x-forwarded-for"] as string;
```
sam701: This part is not very reliable because the IP can be IPv6 and `x-forwarded-for` can be explicitly set by the client. 

I'd suggest to use `x-real-ip` for this use case and trust the load balancer that it correctly sets this header.

apps/api/src/middleware/rate-limiter.middleware.ts
```diff
@@ -77,12 +126,38 @@ export const rateLimiterMiddleware = async (
       return next();
     }
 
-    // Get agent ID from request (set by auth middleware)
-    const agentId = req.agentId || "anonymous";
+    // Get agent ID from request (set by auth middleware) or use IP address as fallback
+    let agentId = req.agentId || `ip:${getClientIp(req)}`;
```
sam701, line 121: Just noticed this line. That's great! It already implements the idea to use the IP only as a fallback. 👍

apps/api/src/middleware/rate-limiter.middleware.ts
mzkrasner: Right - the issue right now is that all unauthenticated traffic defaults identity to `anonymous` which uniformly gets rate limited, which is currently an issue (for example, more than 30 users accessing the frontend at one time on the `leaderboard` page

apps/api/src/middleware/rate-limiter.middleware.ts
mzkrasner: @sam701 great feedback - i just pushed up that change, so let me know what you think

sam701 (approved):
great work 🚀

apps/api/e2e/tests/rate-limiter.test.ts
```diff
@@ -457,4 +457,142 @@ describe("Rate Limiter Middleware", () => {
       "Successfully verified health endpoint exemption from rate limits",
     );
   });
+
+  test("unauthenticated requests are rate limited by IP address", async () => {
+    // Test IP-based rate limiting for unauthenticated requests
+    // Using /api/agents endpoint which is public and has account rate limit (30/min)
```
dtbuchholz, line 463: @mzkrasner the overall IP-based logic lgtm. however, which endpoints should this apply to? it should be _anything_ that can be called publicly, correct?

if i test this locally with the `/competitions` endpoint, i _dont_ see the rate limiter error. e.g., all of these succeed:
```
for i in {1..100}; do curl "http://localhost:3000/api/competitions/"; done
```

dtbuchholz: ah or i guess this is expected because the "global" limit is 3000/min https://github.com/recallnet/js-recall/blob/main/apps/api/src/middleware/rate-limiter.middleware.ts#L24

Issue #693, opened 2025-06-23
[BUG] User profile agents list are shown incorrectly


Rcontre360:
Many agents should be shown horizontally.

![Image](https://github.com/user-attachments/assets/d4229c84-41d6-43be-9e15-e611196d2100)

PR #706, opened 2025-06-23
fix(comps): fixed wrong agent list layout

Rcontre360:
Fixes bug #693


ramirotw (changes requested):
1. input misalignment 
image

2. Your Competitions and Your Agents separation
image

3. Agent Card width, name was truncated before
image

ramirotw (commented):
nice! the issue with the agent card is fixed, now the only thing is what the Cursor's bot commented about the `overflow-none` class which is invalid

_about the forced pushed: I mistakenly pushed to this branch_

ramirotw (approved):

Issue #691, opened 2025-06-20
[BUG] `/user/agents` API throws 500 instead of 409 for dupe agent name


dtbuchholz:
**Bug Description**
- When you create an agent via the comps app, you define an agent name. It must be unique.
- If this agent name already exists, the `agents` table's contraint will fail and throw a 500:
- The API server will show an error message that notes this:
```
Key (owner_id, name)=(3e6ac61b-a797-4bba-865d-60b7bd6934bc, asdf) already exists.
```

We have logic to catch the 409 and display a nice error message explaining that there's a duplicate agent name.  Currently though, the user gets a generic 500 error with no context and explanation of what went wrong.

**Steps to Reproduce**
- Create an agent in the comps app, and then create another with the same name

**Proposed Solution**
Before the we create the agent [here](https://github.com/recallnet/js-recall/blob/main/apps/api/src/controllers/user.controller.ts#L134), we should check if it exists with that name. If so, we need to throw a 409 error

PR #716, opened 2025-06-23
fix(api): return correct http status code when user attempts to create duplicated agent name

joewagner:
fixes: #691

### DIFF
--- a/apps/api/e2e/tests/user.test.ts
+++ b/apps/api/e2e/tests/user.test.ts
@@ -940,6 +940,36 @@ describe("User API", () => {
     expect(emptySortAgents).toHaveLength(3);
   });
 
+  test("user cannot create agents with duplicate names", async () => {
+    // Create a SIWE-authenticated client
+    const { client: siweClient } = await createSiweAuthenticatedClient({
+      adminApiKey,
+      userName: "Duplicate Agent Name Test User",
+      userEmail: "duplicate-agent-name-test@example.com",
+    });
+
+    const agentName = "Duplicate Agent Name";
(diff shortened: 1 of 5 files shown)


apps/api/e2e/tests/user.test.ts
```diff
@@ -940,6 +940,36 @@ describe("User API", () => {
...
+    );
+    expect(firstAgentResponse.success).toBe(true);
+
+    // Try to create a second agent with the same name - should fail with 409
+    const secondAgentResponse = await siweClient.createAgent(
+      agentName,
+      agentDescription,
+    );
+    expect(secondAgentResponse.success).toBe(false);
+    expect((secondAgentResponse as ErrorResponse).status).toBe(409);
```
joewagner, line 967: assert the 409

apps/api/src/services/agent-manager.service.ts
```diff
@@ -157,6 +157,26 @@ export class AgentManager {
     } catch (error) {
       if (error instanceof Error) {
         console.error("[AgentManager] Error creating agent:", error);
+
+        // Check if this is a unique constraint violation for agent name
+        // PostgreSQL throws various error formats depending on the driver and ORM
+        const errorMessage = error.message.toLowerCase();
```
joewagner, line 163: This is kind of messy, but seems to work and be relatively future proof.

dtbuchholz (approved):
nice—and this is how it works with the UI (when you try to submit, it'll catch the 409 and tell you to fix the name)
Screenshot 2025-06-23 at 8 37 17 PM

Issue #695, opened 2025-06-23
[FEATURE] Add admin API for creating an agent


dtbuchholz:
**Problem Statement**
We have an existing admin API to create a user _and_ agent in one call. But, we need a way to do these in independent steps.

**Proposed Solution**
- Ensure existing `POST /admin/users` has all of the fields needed
- Add new `POST /admin/agents`

**Impact**
Our new plan for the app requires a two step process with the "production" server and "sandbox" server. The sandbox server will use the admin API.

**Priority**
Very high.

PR #718, opened 2025-06-24
Dtb/admin agents api

dtbuchholz:
- closes https://github.com/recallnet/js-recall/issues/695

### DIFF
--- a/apps/api/e2e/tests/agent.test.ts
+++ b/apps/api/e2e/tests/agent.test.ts
@@ -389,16 +389,14 @@ describe("Agent API", () => {
     const agentDescription = "Agent for testing profile retrieval";
 
     // Register user and agent with metadata
-    const registerResponse = await client.registerUser(
-      generateRandomEthAddress(),
-      userName,
-      userEmail,
-      undefined,
-      agentName,
-      agentDescription,
-      undefined,
-      metadata,
-    );
+    const registerResponse = await client.registerUser({
+      walletAddress: generateRandomEthAddress(),
+      name: userName,
+      email: userEmail,
+      agentName: agentName,
+      agentDescription: agentDescription,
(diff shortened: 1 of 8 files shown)


apps/api/e2e/tests/agent.test.ts
```diff
@@ -389,16 +389,14 @@ describe("Agent API", () => {
...
-      generateRandomEthAddress(),
-      userName,
-      userEmail,
-      undefined,
-      agentName,
-      agentDescription,
-      undefined,
-      metadata,
-    );
+    const registerResponse = await client.registerUser({
```
dtbuchholz, line 392: note: i updated the admin method for `registerUser` to make this less error prone to forgetting fields

apps/api/src/types/index.ts
```diff
@@ -969,3 +969,30 @@ export const UserVotesParamsSchema = z.object({
 });
 
 export type UserVotesParams = z.infer<typeof UserVotesParamsSchema>;
+
+/**
+ * Admin create agent schema
+ */
+export const AdminCreateAgentSchema = z.object({
+  user: z
```
dtbuchholz, line 977: we only really need "create an agent for a user wallet", but i figured keeping/allowing the "user ID flow" could be useful.

apps/api/src/routes/admin.routes.ts
```diff
@@ -695,6 +695,10 @@ export function configureAdminRoutes(
    *                 type: string
    *                 description: URL to the user's profile image
    *                 example: "https://example.com/user-image.jpg"
+   *               userMetadata:
```
dtbuchholz, line 698: the frontend needs to be able to add user metadata in order to keep the user creation APIs 1:1 for SIWE:admin

PR #722, opened 2025-06-24
Dtb/repo chores

dtbuchholz:


.dockerignore
dtbuchholz, line 1: this was in the root, which seems like the wrong place? do we need this, or should we delete it?

.prettierignore
```diff
@@ -1,11 +1,6 @@
 packages/contracts/contracts
 **/pnpm-lock.yaml
-packages/contracts/src/generated.ts
```
dtbuchholz, line 3: not sure why these were ignored; removed

apps/api/e2e/tests/object-index.test.ts
dtbuchholz, line 1: there's a handful of `apps/api`, `packages/ui`, and some other files that got formatted. for reviewing purposes, ignore.

apps/api/package.json
```diff
@@ -6,6 +6,8 @@
   "description": "Trading Simulator Server for hosting trading competitions",
   "main": "dist/src/index.js",
   "scripts": {
+    "format": "prettier --write . --ignore-path=../../.prettierignore",
```
dtbuchholz, line 9: the way to handle prettier and eslint is a bit of a debate. i used [vercel's turbo monorepo](https://github.com/vercel/turborepo/blob/main/packages/turbo-workspaces/package.json#L25) as a reference:
- they install eslint and prettier at a global level
- each package defines the `lint` script
- each pacakge defines the `format` script (and passes the root prettier ignore file, unless overridden within the package's own prettier ignore file)
- the `turbo.json` and root `package.json` lets you run each subpackage's `lint` or `format` commands

apps/api/package.json
```diff
@@ -47,7 +49,8 @@
     "scripts:populate-object-index": "tsx src/scripts/populateObjectIndex.ts",
     "generate-openapi": "tsx -e \"const { swaggerSpec } = require('./src/config/swagger'); require('fs').writeFileSync('./openapi/openapi.json', JSON.stringify(swaggerSpec, null, 2));\"",
     "generate-markdown": "openapi-markdown -i ./openapi/openapi.json -o ./openapi/API.md",
-    "generate-docs": "pnpm generate-openapi && pnpm generate-markdown"
+    "generate-docs": "pnpm generate-openapi && pnpm generate-markdown",
+    "generate": "pnpm generate-docs && pnpm format"
```
dtbuchholz, line 53: added `pnpm generate` so that we can ensure OpenAPI spec changes are also formatted by default

package.json
```diff
@@ -3,11 +3,11 @@
   "version": "0.0.0",
   "license": "MIT AND Apache-2.0",
   "scripts": {
-    "build": "turbo build --filter='!@recallnet/api-sdk'",
+    "build": "turbo build",
     "dev": "turbo dev --concurrency=20",
     "lint": "turbo lint",
-    "format": "prettier --write \"**/*.{ts,tsx,md,json}\"",
-    "format:check": "prettier --check \"**/*.{ts,tsx,md}\"",
+    "format": "turbo format",
```
dtbuchholz, line 9: e.g., this matches vercel's [turborepo script](https://github.com/vercel/turborepo/blob/main/package.json#L9).

package.json
```diff
@@ -21,6 +21,7 @@
     "@recallnet/eslint-config": "workspace:*",
     "@recallnet/typescript-config": "workspace:*",
     "@trivago/prettier-plugin-sort-imports": "^5.2.1",
+    "eslint": "^9.22.0",
```
dtbuchholz, line 24: note: maybe `eslint` isnt needed as a root dep since our custom eslint package handles this. however, iirc we had CI issues with missing `eslint` dev deps, so i figured i'd include it here vs defining it on a per-package basis. (it's still a bit murky as to how turbo expects these repo-wide deps to work.)

packages/api-sdk/.speakeasy/gen.yaml
```diff
@@ -20,6 +20,9 @@ generation:
     oAuth2ClientCredentialsEnabled: false
     oAuth2PasswordEnabled: false
 typescript:
+  compileCommand:
```
dtbuchholz, line 23: the Speakeasy team said this should allow us to use `pnpm` instead of `npm`. i'm going to handle updating Speakeasy sometime this week. reference docs: [here](https://www.speakeasy.com/guides/sdks/override-compile#overriding-compile-commands-for-sdk-generation).

we previously saw problems when i deleted the `package-lock.json` file, so i'll keep an eye out and can revert this, if needed. i gave it a test locally, and it seemed to work fine.

packages/ui/src/components/shadcn/avatar.tsx
dtbuchholz, line 1: a bunch of `ui` package updates due to the new formatting changes

turbo.json
```diff
@@ -18,6 +18,24 @@
     "lint": {
       "dependsOn": ["^lint"]
     },
+    "format": {
```
dtbuchholz, line 21: new turbo rules for prettier. afaik, these should be the file formats we apply it to?: js,jsx,ts,tsx,json,md,mdx

dtbuchholz:
@mzkrasner @joewagner the diff looks worse than the actual changes here. it's basically:
- per-package and root `package.json` files, `turbo.json`, and their new scripts for linting/formatting
- small Speakeasy tweaks in the `gen.yaml` and root build script

as a follow up, i'm going to look into better CI linting and/or pre-commit/push hooks. PRs like https://github.com/recallnet/js-recall/pull/678 should've been caught and failed due to linting problems, but for whatever reason, that's not happening.

mzkrasner (approved):
makes sense

joewagner (approved):
Works for me, It will be good to keep main up to date with formatting via CI

joewagner:
> as a follow up, i'm going to look into better CI linting and/or pre-commit/push hooks. PRs like #678 should've been caught and failed due to linting problems, but for whatever reason, that's not happening.

fwiw, If CI ensures lint and format devs who like pre-commit hooks and setup `.git/hooks/pre-commit` with:
```

Issue #632, opened 2025-06-16
[FEATURE] Unify error types and messages


dtbuchholz:
**Problem Statement**
Error messages are handled a bit differently throughout the stack. There's duplication or deviation that could benefit from a single `types/errors.ts` (or similar). See comment [here](https://github.com/recallnet/js-recall/pull/593/files#diff-25360bf099884a014ffbd92a2751103dae4ebfd5e39f03caddba78b0fc5dee16R66).

**Proposed Solution**
- Create `types/errors.ts` and/or per-context error files
- Create standardized error message and codes—e.g., extend [`ApiError`](https://github.com/recallnet/js-recall/blob/313a4976bbb7a4c90617aebf066376bdd131cbfb/apps/api/src/middleware/errorHandler.ts#L6) into standard auth errors: `class UnauthorizedError extends ApiError ...`

**Impact**
Simplify API error responses and make the error codes/messages more predictable across different APIs

**Priority**
Low

PR #724, opened 2025-06-24
feat: adds email verification

### LINKED ISSUE #416: [FEATURE] Send 'verified' status to loops email list for segmentation
**Problem Statement**
We're verifying agents who have executed trades and adding a badge to their profile, but it'd be great to also send a tag to the email list so we can easily segment and send drip campaigns to the verified users. 

**Proposed Solution**
When a user is verified and we signal that in the database, we could also send a quick update to Loops so that their User Group can be updated from 'Registered' to 'Verified'

**Impact**
Marketing is impacted, making it easier for us to separate the spam registrations from the ones that can actually compete.

**Priority**
Low and nice to have. I can currently sniff out the verified competitors by pinging the leaderboard and ignoring anyone with the starting balance. It's just not timely enough to set up a drip campaign.

### LINKED ISSUE #437: Add email verification logic

brunocalza:
## Summary
Closes #437 
Closes #416 

## Implementation

- When a user is registered or updated, we check for the email and send a verification email
- When an agent is created or updated, we check for the email and send a verification email
- A token is stored on a new table called `email_verification_tokens`
- A new endpoint is created `GET /verify-email?token=${token}` for verifying the user (or agent)
- The user (or agent) is marked as verified (flag `is_email_verified` is set to `true`)
- That information is sent to Loops

## Implement details
- Two new envs are added: `LOOPS_API_KEY` and `LOOPS_TRANSACTIONAL_ID`. `LOOPS_TRANSACTIONAL_ID` is the id of the email template created on Loops web app
- A migration for  `email_verification_tokens` table and `is_email_verified` flag is added
- `EmailVerificationService` is responsible for verifying the user (or agent) token
- `EmailService` is responsible for sending transactional emails

## Notes
- Before deploying, we must create a transaction email on Loops and set `LOOPS_API_KEY` and `LOOPS_TRANSACTIONAL_ID`
- This was tested with automated tests (`email-verification.test.ts`) that ignores Loops by not setting the envs
- This was also manually tested by using the scripts `register:user` and `register:agent`

### DIFF
--- a/apps/api/src/controllers/email-verification.controller.ts
+++ b/apps/api/src/controllers/email-verification.controller.ts
@@ -1,0 +1,75 @@
+import { NextFunction, Request, Response } from "express";
+
+import { config } from "@/config/index.js";
+import { ServiceRegistry } from "@/services/index.js";
+
+export function makeEmailVerificationController(services: ServiceRegistry) {
+  /**
+   * Email Verification Controller
+   * Handles email verification endpoints
+   */
+  return {
+    /**
+     * Verify an email verification token
+     * @param req Request object
+     * @param res Response object
+     * @param next Next function
+     */
+    async verifyEmail(req: Request, res: Response, next: NextFunction) {
+      try {
+        const { token } = req.query;
+
+        if (!token || typeof token !== "string") {
+          return res.redirect(
+            `${config.app.url}/profile?success=false&message=${encodeURIComponent("Token is required")}`,
+          );
+        }
+
+        const result =
+          await services.emailVerificationService.verifyToken(token);
+
+        if (!result.success) {
+          let message = "Verification failed";
+
+          switch (result.error.type) {
+            case "InvalidToken":
+              message = "Invalid verification token";
+              break;
+            case "TokenHasBeenUsed":
+              message = "Token has already been used";
+              break;
+            case "TokenHasExpired":
+              message = "Token has expired";
+              break;
+            case "UserNotFound":
+              message = "User not found";
+              break;
+            case "AgentNotFound":
+              message = "Agent not found";
+              break;
+            case "NoAssociation":
+              message = "Token is not associated with a user or agent";
+              break;
+            case "SystemError":
+              message = result.error.message;
+              break;
+          }
+
+          return res.redirect(
+            `${config.app.url}/profile?success=false&message=${encodeURIComponent(message)}`,
+          );
+        }
+
+        return res.redirect(
+          `${config.app.url}/profile?success=true&message=${encodeURIComponent(result.value.message)}`,
+        );
+      } catch (error) {
+        next(error);
+      }
+    },
+  };
+}
+
+export type EmailVerificationController = ReturnType<
+  typeof makeEmailVerificationController
+>;
--- a/apps/api/src/controllers/email-verification.controller.ts
+++ b/apps/api/src/controllers/email-verification.controller.ts
@@ -41,12 +41,6 @@ export function makeEmailVerificationController(services: ServiceRegistry) {
             case "TokenHasExpired":
               message = "Token has expired";
               break;
-            case "UserNotFound":
-              message = "User not found";
-              break;
-            case "AgentNotFound":
-              message = "Agent not found";
-              break;
             case "NoAssociation":
               message = "Token is not associated with a user or agent";
               break;
--- a/apps/api/src/database/repositories/email-verification-repository.ts
+++ b/apps/api/src/database/repositories/email-verification-repository.ts
@@ -2,19 +2,19 @@ import { eq } from "drizzle-orm";
 
 import { db } from "@/database/db.js";
 import { emailVerificationTokens } from "@/database/schema/core/defs.js";
+import {
+  InsertEmailVerificationToken,
+  SelectEmailVerificationToken,
+} from "@/database/schema/core/types.js";
 
 /**
  * Creates a new email verification token
  * @param token The token data to insert
  * @returns The inserted token
  */
-export async function createEmailVerificationToken(token: {
-  id: string;
-  userId?: string;
-  agentId?: string;
-  token: string;
-  expiresAt: Date;
-}): Promise<typeof emailVerificationTokens.$inferSelect> {
+export async function createEmailVerificationToken(
+  token: InsertEmailVerificationToken,
+): Promise<SelectEmailVerificationToken> {
   const [insertedToken] = await db
     .insert(emailVerificationTokens)
     .values(token)
@@ -34,7 +34,7 @@ export async function createEmailVerificationToken(token: {
  */
 export async function findEmailVerificationTokenByToken(
   tokenString: string,
-): Promise<typeof emailVerificationTokens.$inferSelect | undefined> {
+): Promise<SelectEmailVerificationToken | undefined> {
   const tokens = await db
     .select()
     .from(emailVerificationTokens)
@@ -51,7 +51,7 @@ export async function findEmailVerificationTokenByToken(
  */
 export async function markTokenAsUsed(
   tokenId: string,
-): Promise<typeof emailVerificationTokens.$inferSelect | undefined> {
+): Promise<SelectEmailVerificationToken | undefined> {
   const [updatedToken] = await db
     .update(emailVerificationTokens)
     .set({ used: true })
--- a/apps/api/src/database/repositories/email-verification-repository.ts
+++ b/apps/api/src/database/repositories/email-verification-repository.ts
@@ -1,0 +1,62 @@
+import { eq } from "drizzle-orm";
+
+import { db } from "@/database/db.js";
+import { emailVerificationTokens } from "@/database/schema/core/defs.js";
+
+/**
+ * Creates a new email verification token
+ * @param token The token data to insert
+ * @returns The inserted token
+ */
+export async function createEmailVerificationToken(token: {
+  id: string;
+  userId?: string;
+  agentId?: string;
+  token: string;
+  expiresAt: Date;
+}): Promise<typeof emailVerificationTokens.$inferSelect> {
+  const [insertedToken] = await db
+    .insert(emailVerificationTokens)
+    .values(token)
+    .returning();
+
+  if (!insertedToken) {
+    throw new Error("Failed to create email verification token");
+  }
+
+  return insertedToken;
+}
+
+/**
+ * Finds an email verification token by its token string
+ * @param tokenString The token string to search for
+ * @returns The token if found, undefined otherwise
+ */
+export async function findEmailVerificationTokenByToken(
+  tokenString: string,
+): Promise<typeof emailVerificationTokens.$inferSelect | undefined> {
+  const tokens = await db
+    .select()
+    .from(emailVerificationTokens)
+    .where(eq(emailVerificationTokens.token, tokenString))
+    .limit(1);
+
+  return tokens[0];
+}
+
+/**
+ * Marks a token as used
+ * @param tokenId The ID of the token to mark as used
+ * @returns The updated token
+ */
+export async function markTokenAsUsed(
+  tokenId: string,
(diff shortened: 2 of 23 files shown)


apps/api/src/controllers/email-verification.controller.ts
```diff
@@ -0,0 +1,75 @@
...
+   * Handles email verification endpoints
+   */
+  return {
+    /**
+     * Verify an email verification token
+     * @param req Request object
+     * @param res Response object
+     * @param next Next function
+     */
+    async verifyEmail(req: Request, res: Response, next: NextFunction) {
```
brunocalza, line 18: Handler for `/verify-email` endpoint. The result of this is a redirect to `/profile/profile?success=false&message=${message}`. Let me know if this is ok for now, or we need adjustments.

apps/api/src/database/schema/core/defs.ts
```diff
@@ -63,6 +63,7 @@ export const users = pgTable(
     email: varchar({ length: 100 }).unique(),
     imageUrl: text("image_url"),
     metadata: jsonb(),
+    isEmailVerified: boolean("is_email_verified").default(false),
```
brunocalza, line 67: adds flag to users table

apps/api/src/database/schema/core/defs.ts
```diff
@@ -97,6 +98,7 @@ export const agents = pgTable(
     imageUrl: text("image_url"),
     apiKey: varchar("api_key", { length: 400 }).notNull(),
     metadata: jsonb(),
+    isEmailVerified: boolean("is_email_verified").default(false),
```
brunocalza, line 102: adds flag to agents table

apps/api/src/database/schema/core/defs.ts
```diff
@@ -349,3 +351,47 @@ export const competitionsLeaderboard = pgTable(
     }).onDelete("cascade"),
   ],
 );
+
+/**
+ * Email verification tokens for users and agents
+ */
+export const emailVerificationTokens = pgTable(
```
brunocalza, line 359: new table to store email verification tokens

apps/api/src/services/email-verification.service.ts
```diff
@@ -0,0 +1,201 @@
...
+      await markTokenAsUsed(tokenRecord.id);
+
+      // Check if it's associated with user
+      if (tokenRecord.userId) {
+        const user = await findUserById(tokenRecord.userId);
+        await this.userManager.markEmailAsVerified(tokenRecord.userId);
+
+        // Update Loops with "Verified" userGroup if user exists
+        if (user && user.email && user.name) {
+          await this.updateLoopsWithVerifiedStatus(user.email, user.name);
```
brunocalza, line 94: solves #416

apps/api/src/services/email-verification.service.ts
```diff
@@ -0,0 +1,201 @@
...
+      }
+
+      // Check if it's associated with agent
+      if (tokenRecord.agentId) {
+        const agent = await findAgentById(tokenRecord.agentId);
+        await this.agentManager.markEmailAsVerified(tokenRecord.agentId);
+
+        // Update Loops with "Verified" userGroup if agent exists
+        if (agent && agent.email && agent.name) {
+          await this.updateLoopsWithVerifiedStatus(agent.email, agent.name);
```
brunocalza, line 113: solves #416

apps/api/src/services/email-verification.service.ts
```diff
@@ -0,0 +1,201 @@
...
+    }
+  }
+
+  /**
+   * Update Loops with "Verified" userGroup status
+   * @param email The email address to update
+   * @param name The name to update
+   * @returns Promise resolving to the API response, or null if API key is missing
+   */
+  private async updateLoopsWithVerifiedStatus(
```
brunocalza, line 151: this was tested with a free account that I created

apps/api/src/services/email.service.ts
```diff
@@ -0,0 +1,127 @@
...
+    if (!this.isEmailEnabled()) {
+      console.log(
+        "[EmailService] Email sending skipped - API key or transactional ID not provided",
+      );
+      return null;
+    }
+
+    const verificationLink = this.formatVerificationLink(token);
+
+    const payload: Payload = {
```
brunocalza, line 70: well, we don't have an email template yet, this is the bare minimum that needs to be sent. when we create the transactional email we may need to adjust this to send more information

brunocalza:
bugbot run

apps/api/src/controllers/email-verification.controller.ts
dtbuchholz: makes sense. created a UI feature request to match this: https://github.com/recallnet/js-recall/issues/772

apps/api/src/database/repositories/email-verification-repository.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @param token The token data to insert
+ * @returns The inserted token
+ */
+export async function createEmailVerificationToken(token: {
+  id: string;
+  userId?: string;
+  agentId?: string;
+  token: string;
+  expiresAt: Date;
+}): Promise<typeof emailVerificationTokens.$inferSelect> {
```
dtbuchholz: for consistency, we should create a type for `typeof emailVerificationTokens.$inferSelect`, place it on the `core/types.ts` folder, and import it here.

apps/api/src/database/schema/core/defs.ts
```diff
@@ -349,3 +351,47 @@ export const competitionsLeaderboard = pgTable(
...
+  "email_verification_tokens",
+  {
+    id: uuid().primaryKey().notNull(),
+    userId: uuid("user_id"),
+    agentId: uuid("agent_id"),
+    token: varchar("token", { length: 255 }).notNull().unique(),
```
dtbuchholz: the token looks like this `410cf13e-d40a-4568-929e-58f0f4c4a1f7`, which is length 36. is there any reason why `255` is used here?

apps/api/src/services/agent-manager.service.ts
```diff
@@ -70,10 +72,13 @@ export class AgentManager {
   private apiKeyCache: Map<string, ApiAuth>;
   // Cache for inactive agents to avoid repeated database lookups
   private inactiveAgentsCache: Map<string, { reason: string; date: Date }>;
+  // Email service for sending verification emails
+  private emailService: EmailService;
 
   constructor() {
     this.apiKeyCache = new Map();
     this.inactiveAgentsCache = new Map();
+    this.emailService = new EmailService();
```
dtbuchholz: similar point above as `UserManager`. we should be able to take the service as a param and pass the `EmailService` instance in `services/index.ts`.

apps/api/src/services/email-verification.service.ts
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * Email Verification Service
+ * Handles email verification token creation and validation
+ */
+export class EmailVerificationService {
+  private userManager: UserManager;
+  private agentManager: AgentManager;
+
+  constructor() {
```
dtbuchholz: and same here—take these as params and let `services/index.ts` pass them.

apps/api/src/services/email-verification.service.ts
dtbuchholz: i tested it as well w/ a personal account and the comps frontend app running locally...works great

apps/api/src/services/email-verification.service.ts
```diff
@@ -0,0 +1,199 @@
...
+      const loopsApiKey = config.email.apiKey;
+      if (!loopsApiKey) {
+        console.error(
+          "[EmailVerificationService] Missing LOOPS_API_KEY in environment variables",
+        );
+        return;
+      }
+
+      // Prepare the payload for Loops API
+      const loopsPayload = {
```
dtbuchholz, line 166: this is fine for now. we'll follow up and align this with the production Loops configuration https://github.com/recallnet/js-recall/issues/773

apps/api/src/services/email.service.ts
```diff
@@ -0,0 +1,127 @@
...
+const EMAIL_VERIFICATION_PATH = "api/verify-email";
+
+/**
+ * Interface for verification email payload
+ */
+interface Payload {
+  transactionalId: string;
+  email: string;
+  dataVariables: {
+    verification_link: string;
```
dtbuchholz: should we make `verification_link` camel case?

apps/api/src/services/email.service.ts
dtbuchholz: this is fine for now. we'll follow up and align this with the production Loops configuration https://github.com/recallnet/js-recall/issues/773 for both the `dataVariables` and the update contact payload

apps/api/src/services/user-manager.service.ts
```diff
@@ -23,10 +25,13 @@ export class UserManager {
   private userWalletCache: Map<string, string>; // walletAddress -> userId
   // Cache for user profiles by ID
   private userProfileCache: Map<string, SelectUser>; // userId -> user profile
+  // Email service for sending verification emails
+  private emailService: EmailService;
 
   constructor() {
     this.userWalletCache = new Map();
     this.userProfileCache = new Map();
+    this.emailService = new EmailService();
```
dtbuchholz: instead of instantiating the service in the constructor, we should have the constructor take it as a param and then pass it to `UserManager` the root `services/index.ts` instance.

apps/api/src/services/email-verification.service.ts
```diff
@@ -0,0 +1,199 @@
...
+import { findById as findAgentById } from "@/database/repositories/agent-repository.js";
+import {
+  findEmailVerificationTokenByToken,
+  markTokenAsUsed,
+} from "@/database/repositories/email-verification-repository.js";
+import { findById as findUserById } from "@/database/repositories/user-repository.js";
+import { AgentManager } from "@/services/agent-manager.service.js";
+import { UserManager } from "@/services/user-manager.service.js";
+
+type Result<Ok, Err> =
```
dtbuchholz, line 13: in general, our return types and errors in this codebase are messy. some exist in the root `types`; some are within a file; some are in the database defs. 

although this rust-like pattern could be generally useful, it's not used anywhere else.

that said, https://github.com/recallnet/js-recall/issues/632 needs to happen and can unify a consistent pattern across each of our app layers. this might be a useful first step in that direction, so i'm fine if we want to keep this logic.

i know @joewagner has some thoughts on error patterns, types, and zod, so it might help to get his thoughts here.

dtbuchholz:
@brunocalza just a few small tweaks, but overall this works great. we can handle any final deployment and logic tweaks sometime next week or the following, depending on when marketing wants to start using this logic.

apps/api/src/services/email-verification.service.ts
brunocalza: yeah, I couldn't find a pattern in the codebase, also I wasn't sure what's the idiomatic way of doing this in Typescript, so I used this one

apps/api/src/database/schema/core/defs.ts
brunocalza: not really, the migration was done before deciding to use uuid and stayed that way. i'll change it to `char(36)`

dtbuchholz (approved):
🤘

Issue #637, opened 2025-06-17
[FEATURE] Ensure pagination and sorting works on all pages


dtbuchholz:
**Problem Statement**
The frontend implements sorting on tables, but the backend APIs don't always handle this correctly. We need to address _all_ of the API issues.

**Proposed Solution**
1. First, we need to validate APIs work with the [Figma designs](https://www.figma.com/design/WqQqcsg1g66EdMjjM8yfIM/Competitions-App---v.1.0--06-25-?node-id=6-21170&p=f&t=884nUo9fHtWecSlO-0) on _every page_. The easiest way to check this is to look at the designs and check any table headers/sorting fields.
2. Write tests in `apps/api` for _all_ required sorting fields. i.e., if there's a table header with a sort button, it should exist in the tests.
3. Wire up the frontend. All of the tables have _most_ of what we need set up. E.g., look for [`enableSorting`](https://github.com/recallnet/js-recall/blob/main/apps/comps/components/competitions-table.tsx#L66) and the access key.

**Impact**
Needed for app launch as part of the QA process.

**Priority**
Highest.

dtbuchholz:
@stbrody jfyi new priority issue for debugging the pagination/sorting logic and something we want to starting working on tmrw as part of the QA efforts. it's blocked by the in-flight issues that should land today.

dtbuchholz:
(i guess this is sorta "planned" work as part of QA. i'll be firing off a bunch in the coming days, but this one is already a known issue that we can just start working on asap.)

joewagner:
I went through the figma pages again and this all of the tables I found that support sorting.  Putting everything here for clarity, feel free to weigh in. 

## [Specific Competition page](https://www.figma.com/design/WqQqcsg1g66EdMjjM8yfIM/Competitions-App---v.1.0--06-25-?node-id=6-21170&p=f&t=GInHNEW9lAAw3EKf-0)

leaderboard table:
- Rank ( 1st, 2nd, 3rd, etc...)
- Agent (Agent name)
- Portfolio (USD amount in the agents portfolio/account balance)
- P&L (USD value of profit/loss)
- 24h (24 hour percent change)
- Votes (number of votes recieved)

## [Home (comp hub)](https://www.figma.com/design/WqQqcsg1g66EdMjjM8yfIM/Competitions-App---v.1.0--06-25-?node-id=0-1&p=f&t=GInHNEW9lAAw3EKf-0)

Upcoming competitions view:
[...]

PR #725, opened 2025-06-25
fix(comps): The agent profile page competition table shows data and can be sorted.

joewagner:
part of: https://github.com/recallnet/js-recall/issues/637

Connect front end agent profile page competitions table to the parts of the api that are supported currently

### DIFF
--- a/apps/comps/components/agent-profile/comps-table.tsx
+++ b/apps/comps/components/agent-profile/comps-table.tsx
@@ -51,8 +51,7 @@ export function CompetitionTable({
                 Competition
               </SortableTableHeader>
               {
-                //some fields have sorted removed until they are supported by the api
-                // id, name, description, startDate, endDate, createdAt, updatedAt, portfolioValue, pnl, totalTrades, rank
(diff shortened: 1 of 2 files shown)


apps/comps/components/agent-profile/comps-table.tsx
```diff
@@ -51,13 +51,33 @@ export function CompetitionTable({
                 Competition
               </SortableTableHeader>
               {
-                //some fields have sorted removed until they are supported by the api
+                // some fields have sorted removed until they are supported by the api, specifically Trophies and Skills
```
joewagner, line 54: I added the sorting fields that are currently working in the api.

apps/comps/types/competition.ts
```diff
@@ -68,6 +68,14 @@ export interface Competition {
       votedAt?: string;
     };
   };
+  portfolioValue?: number;
```
joewagner, line 71: Since these types being a bit disconnected from the actual api code base, I kept these all optional.

joewagner (commented):
Really small one here, just wiring up the existing api.

Issue #662, opened 2025-06-19
[BUG] Fix comps app bg for `/competitions/:id` error


dtbuchholz:
**Bug Description**
The error color in the comps app should match the app styles.

**Steps to Reproduce**
It's a bit hard to reproduce, but it happens if you sort the table with an invalid sort by param.

**Screenshots**

![Image](https://github.com/user-attachments/assets/2817d7a4-78e2-4893-a150-8a2000e9897e)

PR #726, opened 2025-06-25
fix(comps): add error boundaries for competition and competitions pages

ramirotw:
Fixes https://github.com/recallnet/js-recall/issues/662


dtbuchholz:
@ramirotw hm no, that's not the issue. we have this value set: `NEXT_PUBLIC_API_BASE_URL=https://api.competitions.recall.network/staging/api`

apps/comps/package.json
```diff
@@ -5,7 +5,7 @@
   "private": true,
   "scripts": {
     "dev": "next dev --turbopack --port 3001",
-    "build": "next build",
+    "build": "next build --experimental-build-mode compile",
```
dtbuchholz, line 8: given the point above, we can remove these changes, right?

also, i'm not sure why CI isnt running properly—but once you revert this, i'll keep an eye out to see if the CI keeps hanging

ramirotw: This actually fixes the build issue because it disables SSG, so the CI doesn’t try to make any http calls to the API during build time

dtbuchholz: ah gotcha, kk

dtbuchholz (approved):

PR #730, opened 2025-06-25
Dtb/test bad format

dtbuchholz:


apps/api/src/config/index.ts
```diff
@@ -4,14 +4,18 @@ import path from "path";
 import {
   CROSS_CHAIN_TRADING_TYPE,
   CrossChainTradingType,
+
```
dtbuchholz, line 7: misc changes that should get caught in CI, but they dont. locally, they _do_ cause an error with `pnpm format:check`

PR #734, opened 2025-06-25
Rafael/user prof vote table

Rcontre360:
User votes table. 

TODO:

- [ ] enable sorting: need backend for this
- [ ] vote link: need to know where the link should take you

image

### DIFF
--- a/apps/comps/lib/api-client.ts
+++ b/apps/comps/lib/api-client.ts
@@ -10,6 +10,7 @@ import {
   CreateAgentRequest,
   CreateAgentResponse,
   CreateVoteRequest,
+  EnrichedVotesResponse,
   GetAgentCompetitionsParams,
   GetAgentsParams,
   GetCompetitionAgentsParams,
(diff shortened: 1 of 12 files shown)


ramirotw (approved):
tested locally, lgtm!

BTW, I pushed a small fix for an issue I introduced in the AgentsSummary component when rendering the best placement

apps/comps/lib/api-client.ts
```diff
@@ -445,6 +446,42 @@ export class ApiClient {
...
+    const queryParams = this.formatQueryParams(params);
+    const votes = await this.request<VotesResponse>(
+      `/user/votes${queryParams}`,
+    );
+
+    const enrichedVotes = await Promise.all(
+      votes.votes.map(async (vote) => {
+        const [competition, agent] = await Promise.all([
+          this.getCompetition(vote.competitionId),
+          this.getAgent(vote.agentId),
```
ramirotw, line 466: @dtbuchholz we’ll probably want to include the competition and agent data in the `/user/votes` request to avoid having to handle this separately

dtbuchholz: make sense https://github.com/recallnet/js-recall/issues/765

dtbuchholz: one callout in the comments of https://github.com/recallnet/js-recall/issues/765

the rank from `getAgent` is the global rank. that means the "Your votes" section will show you voted for some agent that may or may not have a rank displayed. this might be confusing. however, with the issue above, we can address the problem.

dtbuchholz (approved):
lgtm

apps/comps/lib/api-client.ts
ramirotw: makes sense, you mean agent.stats will be scoped to the competition

PR #738, opened 2025-06-25
feat: update mini app

mzkrasner:
# Summary

- Comprehensive updates to mini app to adhere to the new API structure
- Inherits majority of api formatting and calls from `apps/comps`
- For verification logic, requires a direct connection to the sandbox database to make updates to `is_verified` (temporary strategy and will not be used once normal comps app goes live)


apps/registration/app/account/layout.tsx
```diff
@@ -7,7 +7,7 @@ export const metadata: Metadata = {
     title: "Account - Recall Developer Hub",
     description:
       "Manage your Recall account, API keys, and agent registrations.",
-    url: `${process.env.NEXTAUTH_URL || "https://recall.network/registration"}/account`,
+    url: `https://recall.network/registration/account`,
```
dtbuchholz: should this be?: `https://register.recall.network/account`

dtbuchholz:
one minor problem—on the edit agent page, the data gets improperly stored as a single string instead of an array value. this causes the backend API's zod parsing to break.
   - example of how it gets stored in the table: `{"skills": "Traditio
nal Investing,Sports Betting,Programming / Coding,Art & Video Creation"}`
   - note: the normal "create agent" flow + skills doesnt have this issue; it's only on the edit page
Screenshot 2025-06-26 at 10 53 11 AM

apps/registration/app/account/layout.tsx
mzkrasner: ah good catch - yes

dtbuchholz (approved):

Issue #732, opened 2025-06-25
Fix `bestPlacement` regression in `/user/agents` API & `getBulkAgentMetrics`


dtbuchholz:
We need the `bestPlacement` response to look like this:
```json
"bestPlacement": {
  "competitionId": "string",
  "rank": 1,
  "score": 1234,
  "totalAgents": 100
}
```

The `/user/agents` endpoint was updated to use `getBulkAgentMetrics`, and this method returns a `number` type. 
```ts
export async function getBulkAgentMetrics(agentIds: string[]): Promise<
  Array<{
    agentId: string;
    name: string;
    description: string | null;
    imageUrl: string | null;
    metadata: unknown;
    globalRank: number | null;
    globalScore: number | null;
    completedCompetitions: number;
    totalVotes: number;
    totalTrades: number;
    bestPlacement: number | null;
  }>
>
```

Note: many tests make a check on `bestPlacement` to see if it's undefined, or not undefined. These are legacy checks (before the `bestPlacement` field had data), so our e2e tests didn't fail. But, if `bestPlacement` exists, we should at least assert the correct fields/shape.

Issue #617, opened 2025-06-14
[FEATURE] Add "total ROI" and "highest PnL" to agent stats API response


dtbuchholz:
**Problem Statement**

The user profile page lists out agents. Two fields are missing from the response:
- Total ROI % (bottom left)
- Highest PnL (bottom right)

<details><summary>Screenshot</summary>

![Image](https://github.com/user-attachments/assets/51d4f6e4-446f-4a62-bc00-9b439ed9cd06)

</details>

**Proposed Solution**
- Alter the agent `stats` field to include two new fields: `totalRoi` and `topPnl`
   - For the feature above, we need it to exist in `GET /user/agent` endpoints (but it'll impact others, too)
   - We just need to alter `attachAgentMetrics` in `agent-manager.service.ts` and include the new fields
   - Then, update the OpenAPI spec route's comments for all of the `/user/agent` etc endpoints that include `stats` in the response.
- Total ROI% (`totalRoi`)
   - The `competitions_leaderboard` stores a per-competition `rank` and `score`. The `score` is the _portfolio value_ at the end of a competition (see comment in [`endCompetition` method](https://github.com/recallnet/js-recall/blob/605f4f3d9ac4920cd43a7baff8651f250d2ceec1/apps/api/src/services/competition-manager.service.ts#L277)).
   - So, we can get _all_ `score`s for an agent from the `competitions_leaderboard` table. This means we have the agent's final portfolio value for every competition.
   - We also need to get the "baseline" to compare each portfolio value against. We can query the `portfolio_snapshots` table and find the first entry for the corresponding `competitionId`. Lastly, compute the avg across all comps.
   - This method has some relevant logic: [here](https://github.com/recallnet/js-recall/blob/605f4f3d9ac4920cd43a7baff8651f250d2ceec1/apps/api/src/services/competition-manager.service.ts#L391); the `pnlPercent` is effectively ROI, and you can see the other logic for PnL calculations.
- Highest PnL (`topPnl`)
  - Since the `totalRoi` calculation above will already provide us with the portfolio value and its baseline, we can easily find the entry with the highest PnL diff.

**Impact**
What is the effect on the end user of doing this? What is the cost or risk of not doing this? Why is this something that needs to be done?

**Priority**
How urgent is this change? When does it need to be delivered by? How does the priority compare to that of other in-flight work?

joewagner:
The "highest pnl" section has been added in https://github.com/recallnet/js-recall/pull/739/
[...]

PR #739, opened 2025-06-25
fix(api): user/agents stats response shape

joewagner:
fixes: https://github.com/recallnet/js-recall/issues/732

Ended up making a biggger change here than might have been initially expected, because some of the perf gains of #703 would have been reverted otherwise.

### DIFF
--- a/apps/api/e2e/tests/user.test.ts
+++ b/apps/api/e2e/tests/user.test.ts
@@ -2,6 +2,7 @@ import axios from "axios";
 import { beforeEach, describe, expect, test } from "vitest";
 
 import { config } from "@/config/index.js";
+import { ApiClient } from "@/e2e/utils/api-client.js";
 import {
   AdminSearchUsersAndAgentsResponse,
   AdminUsersListResponse,
@@ -514,7 +515,7 @@ describe("User API", () => {
       );
       expect(response.success).toBe(true);
       // Small delay to ensure different timestamps
-      await new Promise((resolve) => setTimeout(resolve, 10));
+      await wait(10);
     }
 
     // Test default pagination (should return all agents)
@@ -550,7 +551,7 @@ describe("User API", () => {
         `Description for Agent ${i}`,
       );
       expect(response.success).toBe(true);
-      await new Promise((resolve) => setTimeout(resolve, 10));
+      await wait(10);
     }
 
     // Test limit of 3
@@ -587,7 +588,7 @@ describe("User API", () => {
         `Description for Agent ${i}`,
       );
       expect(response.success).toBe(true);
-      await new Promise((resolve) => setTimeout(resolve, 10));
+      await wait(10);
     }
 
     // Get all agents first to establish baseline
@@ -630,7 +631,7 @@ describe("User API", () => {
         `Description for Agent ${i}`,
       );
       expect(response.success).toBe(true);
-      await new Promise((resolve) => setTimeout(resolve, 10));
+      await wait(10);
     }
 
     // Test pagination: page 1 (offset 0, limit 3)
@@ -691,7 +692,7 @@ describe("User API", () => {
         agent.description,
       );
       expect(response.success).toBe(true);
-      await new Promise((resolve) => setTimeout(resolve, 50)); // Larger delay for timestamp differentiation
+      await wait(50); // Larger delay for timestamp differentiation
     }
 
     // Test name ascending sort
@@ -755,7 +756,7 @@ describe("User API", () => {
         `Description for ${name}`,
       );
       expect(response.success).toBe(true);
-      await new Promise((resolve) => setTimeout(resolve, 10));
+      await wait(10);
     }
 
     // Get first page of 2, sorted by name ascending
@@ -1563,7 +1564,7 @@ describe("User API", () => {
         create1Response as CreateCompetitionResponse
       ).competition.id;
       await userClient.joinCompetition(competitionIdForFirstComp, agent.id);
-      await new Promise((resolve) => setTimeout(resolve, 50));
+      await wait(50);
 
       const create2Response = await adminClient.createCompetition(
         secondComp,
@@ -1573,7 +1574,7 @@ describe("User API", () => {
       const createCompResponse2 = create2Response as CreateCompetitionResponse;
       const competitionIdForSecondComp = createCompResponse2.competition.id;
       await userClient.joinCompetition(competitionIdForSecondComp, agent.id);
-      await new Promise((resolve) => setTimeout(resolve, 50));
+      await wait(50);
 
       const create3Response = await adminClient.createCompetition(
         thirdComp,
@@ -1584,7 +1585,7 @@ describe("User API", () => {
         create3Response as CreateCompetitionResponse
       ).competition.id;
       await userClient.joinCompetition(competitionIdForThirdComp, agent.id);
-      await new Promise((resolve) => setTimeout(resolve, 50));
+      await wait(50);
 
       // Start/end the first competition, start/end the second competition, and keep the third pending
       const startCompResponse = await adminClient.startExistingCompetition(
@@ -1690,7 +1691,7 @@ describe("User API", () => {
         );
 
         // Small delay to ensure different timestamps
-        await new Promise((resolve) => setTimeout(resolve, 50));
+        await wait(50);
       }
 
       // Test correct format: name ascending (no prefix)
@@ -1778,7 +1779,7 @@ describe("User API", () => {
         );
 
         // Controlled delay to ensure different timestamps
-        await new Promise((resolve) => setTimeout(resolve, comp.delay));
+        await wait(comp.delay);
       }
 
       // Test multiple sort fields: name ascending, then createdAt descending
@@ -1849,7 +1850,7 @@ describe("User API", () => {
         await userClient.joinCompetition(competition.id, agents[i]!.id);
 
         // Small delay to ensure different timestamps
-        await new Promise((resolve) => setTimeout(resolve, 50));
+        await wait(50);
       }
 
       // Test agentName ascending sort (should work now!)
@@ -2238,7 +2239,7 @@ describe("User API", () => {
         await userClient.joinCompetition(competition.id, agent.id);
 
         // Controlled delay for predictable timestamps
-        await new Promise((resolve) => setTimeout(resolve, 100));
+        await wait(100);
       }
 
       // Test combined sort: agentName ascending, then createdAt descending
@@ -2298,7 +2299,7 @@ describe("User API", () => {
           .competition;
         await userClient.joinCompetition(competition.id, agents[i]!.id);
 
-        await new Promise((resolve) => setTimeout(resolve, 50));
+        await wait(50);
       }
 
       // Test first page (limit 2, offset 0) with agentName sorting
@@ -2656,7 +2657,7 @@ describe("User API", () => {
         await userClient.joinCompetition(competition.id, agents[i]!.id);
 
         // Controlled delay for predictable timestamps
-        await new Promise((resolve) => setTimeout(resolve, 100));
+        await wait(100);
       }
 
       // Test pagination with mixed sort: agentName (computed) + createdAt descending (database)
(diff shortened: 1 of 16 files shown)


apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -725,6 +727,69 @@ export async function getAgentPortfolioSnapshots(
   }
 }
 
+/**
+ * Get the newest and oldest portfolio snapshots for an agent in a competition
+ * @param competitionId Competition ID
+ * @param agentId Agent ID
+ * @returns Object with newest and oldest snapshots, or null if no snapshots found
+ */
+export async function getBoundedSnapshots(
```
joewagner, line 737: I built this function to help reduce load when we want to calculate the stats for an agent when a competition is ended.
There were places we were using `getAgentPortfolioSnapshots` for this, and that returns all the snapshots for a comp+ agent combo, but we only need the first and last.

apps/api/src/database/repositories/leaderboard-repository.ts
```diff
@@ -197,27 +204,51 @@ export async function getBulkAgentMetrics(agentIds: string[]): Promise<
...
+      .from(competitionsLeaderboard)
+      .where(inArray(competitionsLeaderboard.agentId, agentIds))
+      .orderBy(
+        competitionsLeaderboard.agentId,
+        asc(competitionsLeaderboard.rank),
+      )
+      .limit(1);
+
+    // Query 6: Best Profit/Loss
+    const bestPnlQuery = db
```
joewagner, line 238: I added another query to get the best PnL, which is expected by the app.

apps/api/src/database/schema/core/defs.ts
```diff
@@ -319,6 +320,20 @@ export const competitionsLeaderboard = pgTable(
     agentId: uuid("agent_id").notNull(),
     competitionId: uuid("competition_id").notNull(),
     rank: integer("rank").notNull(),
+    pnl: numeric("pnl", {
```
joewagner: I added these two columns to the `competitions_leaderboard` table so our queries don't have to do quite so much post result processing in memory.  There are still more optimizations to be made similar to this, but it seemed worth doing this much in this PR.

joewagner: Note: I gave these default of `0` to make migrations less complicated. We can potentially remove default later.

apps/api/src/services/agent-manager.service.ts
```diff
@@ -1368,6 +1369,31 @@ export class AgentManager {
     }
   }
 
+  async getAgentPnlForComp(agentId: string, competitionId: string) {
```
joewagner, line 1392: I refactored this into a function so it can be used inside the `endCompetition` function, and the values can be saved instead of recalculating on each request.

apps/api/src/services/agent-manager.service.ts
```diff
@@ -1379,27 +1405,10 @@ export class AgentManager {
     agentId: string,
   ): Promise<EnhancedCompetition> {
     try {
-      // Get portfolio value (latest snapshot for agent in competition)
-      const agentSnapshots = await getAgentPortfolioSnapshots(
```
joewagner, line 1403: Don't get all snapshots, just newest and oldest.

apps/api/src/services/competition-manager.service.ts
```diff
@@ -293,12 +294,31 @@ export class CompetitionManager {
...
-      competitionId,
-      rank: index + 1, // 1-based ranking
-      score: entry.value, // Use the portfolio value as the score
-    }));
+
+    const leaderboardEntries = [];
+    // Note: the leaderboard array could be quite large, avoiding Promise.all
+    //  so that these async calls to get pnl happen in series and don't over
+    //  use system resorces.
+    for (let i = 0; i < leaderboard.length; i++) {
```
joewagner, line 302: Here's the place in `endCompetition` were I'm storing the new columns in the db

apps/comps/components/user-agents/agents-summary.tsx
```diff
@@ -78,13 +78,13 @@ export const AgentsSummary: React.FunctionComponent<{
...
-        <span className="text-2xl font-semibold">
-          {highest ? (
-            <BigNumberDisplay value={highest.toString()} decimals={0} />
-          ) : (
-            "N/A"
-          )}
-        </span>
+        {highest && highest !== Number.MIN_SAFE_INTEGER ? (
+          <span className="text-2xl font-semibold">
+            <BigNumberDisplay value={highest.toString()} decimals={0} /> USD
```
joewagner, line 84: Added units here.  Also, during manual tests noticed that the `decimals` prop doesn't work how I thought it would.  If `highest` is equal to `-0.1` this displays "-0.1".  I would think it would round to nearest int.

apps/comps/components/user-agents/index.tsx
```diff
@@ -41,6 +41,17 @@ export default function UserAgentsSection({ agents }: { agents: Agent[] }) {
     [agents],
   );
 
+  const highest = useMemo(
```
joewagner, line 44: This is available now.

apps/api/src/database/schema/core/defs.ts
joewagner: @mzkrasner These are the columns I mentioned adding in the call today.  These might affect the way you are loading past competition data, but I'm not sure.

apps/api/src/services/competition-manager.service.ts
```diff
@@ -486,14 +508,15 @@ export class CompetitionManager {
...
-      const leaderboard: { agentId: string; value: number }[] = [];
+      const leaderboard: { agentId: string; value: number; pnl: number }[] = [];
 
       for (const agentId of agents) {
         const portfolioValue =
           await this.tradeSimulator.calculatePortfolioValue(agentId);
         leaderboard.push({
           agentId,
           value: portfolioValue,
+          pnl: 0, // TODO: if there's no competitions_leaderboard row we don't have a pnl
```
joewagner, line 519: I didn't want to add any more to this PR, but we might want to handle this differently in the future.

apps/comps/components/user-agents/agents-summary.tsx
joewagner: cc: @ramirotw I made some updates to the comps app to show the available fields, and noticed this.

apps/api/e2e/tests/user.test.ts
```diff
@@ -512,7 +513,7 @@ describe("User API", () => {
       );
       expect(response.success).toBe(true);
       // Small delay to ensure different timestamps
-      await new Promise((resolve) => setTimeout(resolve, 10));
+      await wait(10);
```
joewagner, line 518: Unrelated I found a few places we should be taking advantage of the test helper `wait`

apps/api/e2e/tests/user.test.ts
```diff
@@ -2698,4 +2699,386 @@ describe("User API", () => {
       expect(page2Comps[1]?.name).toBe("Competition E"); // Gamma Agent
     });
   });
+
+  test("user agents have correct bestPlacement stats after competition", async () => {
```
joewagner: These tests are kind of long but seems needed to get the setup we want to test.

apps/api/e2e/tests/user.test.ts
```diff
@@ -2698,4 +2699,386 @@ describe("User API", () => {
...
+        amount: "100",
+        reason: `Bravo Agent good trade ${i + 1}`,
+      });
+      await wait(50);
+    }
+
+    // Charlie Agent: (1 large volitile trade)
+    await agentClients[2]?.executeTrade({
+      fromToken: config.specificChainTokens.eth.usdc,
+      toToken: config.specificChainTokens.eth.vision,
```
joewagner, line 2790: Notice I added this token to the trades.  It seems popular enough to continue to exist, but also volatile enough to create some variance in the rankings.

apps/api/e2e/tests/user.test.ts
```diff
@@ -2698,4 +2699,386 @@ describe("User API", () => {
...
+    expect(endResult.success).toBe(true);
+
+    // Wait a bit for rankings to be calculated
+    await wait(1000);
+
+    // Get all user agents and verify bestPlacement stats
+    const agentsResponse = await siweClient.getUserAgents();
+    expect(agentsResponse.success).toBe(true);
+    const agents = (agentsResponse as GetUserAgentsResponse).agents;
+    expect(agents).toHaveLength(5);
```
joewagner, line 2830: make sure we get all of the user's agents.

apps/api/e2e/tests/user.test.ts
```diff
@@ -2698,4 +2699,386 @@ describe("User API", () => {
...
+    expect(bravoAgent?.stats?.totalTrades).toBe(2);
+    expect(charlieAgent?.stats?.totalTrades).toBe(1);
+    expect(deltaAgent?.stats?.totalTrades).toBe(1);
+    expect(echoAgent?.stats?.totalTrades).toBe(2);
+
+    // Verify all agents have valid rankings
+    expect(alphaAgent?.stats?.bestPlacement?.rank).toBeGreaterThanOrEqual(1);
+    expect(alphaAgent?.stats?.bestPlacement?.rank).toBeLessThanOrEqual(5);
+    expect(bravoAgent?.stats?.bestPlacement?.rank).toBeGreaterThanOrEqual(1);
+    expect(bravoAgent?.stats?.bestPlacement?.rank).toBeLessThanOrEqual(3);
```
joewagner, line 2870: We can't be too specific here, since we don't know exactly how the token values have changed.  We can however say for sure the agents that burned tokens will be at the bottom of the rankings.

apps/api/scripts/competition-status.ts
```diff
@@ -220,10 +220,12 @@ async function showCompetitionStatus() {
         {
           agentId: string;
           value: number;
+          pnl: number; // TODO: casting here probably isn't right
         },
         ...{
           agentId: string;
           value: number;
+          pnl: number; // TODO: casting here probably isn't right
```
joewagner, line 228: Not related, just a note for later.

apps/api/e2e/tests/user.test.ts
```diff
@@ -2698,4 +2699,364 @@ describe("User API", () => {
...
+    expect(uniqueRanks.size).toBe(5); // All ranks should be unique
+
+    // All agents should have the same competition data in bestPlacement
+    for (const agent of agents) {
+      expect(agent.stats?.bestPlacement?.competitionId).toBe(competitionId);
+      expect(agent.stats?.bestPlacement?.totalAgents).toBe(5);
+    }
+  });
+
+  test("two agents in multiple competitions have correct stats", async () => {
```
joewagner, line 2897: This test is about ensuring bestPlacement is loaded with the actual best for an agent when there is more than one comp.

apps/api/e2e/tests/user.test.ts
```diff
@@ -2698,4 +2699,364 @@ describe("User API", () => {
...
+    expect(typeof agentFoxtrot?.stats?.bestPnl).toBe("number");
+    expect(typeof agentHotel?.stats?.bestPnl).toBe("number");
+
+    // Verify that bestPlacement reflects the better performance from either competition
+    // The agent with rank 1 should have the better placement
+    const agentFoxtrotRank = agentFoxtrot?.stats?.bestPlacement?.rank;
+    const agentHotelRank = agentHotel?.stats?.bestPlacement?.rank;
+
+    // Ensure ranks are the same since each agent has won one comp
+    expect(agentFoxtrotRank).toBe(agentHotelRank);
```
joewagner, line 3061: They should both have finished first in one comp, hence they both have a `bestPlacement.rank` of `1`

joewagner:
EDIT:
I'm actually gonna wait until you have chance to review, since the conflicts are just in the drizzle migrations.

apps/api/src/database/schema/core/defs.ts
dtbuchholz: added a follow up: https://github.com/recallnet/js-recall/issues/796

or, we can include it here. it's more or less copy/paste (i tested it locally), and it'll let us manually backfill the `pnl` column for a specific competition ID. it should be pretty straightforward to handle in the normalization process.

apps/api/src/services/competition-manager.service.ts
```diff
@@ -293,12 +294,31 @@ export class CompetitionManager {
...
-      score: entry.value, // Use the portfolio value as the score
-    }));
+
+    const leaderboardEntries = [];
+    // Note: the leaderboard array could be quite large, avoiding Promise.all
+    //  so that these async calls to get pnl happen in series and don't over
+    //  use system resorces.
+    for (let i = 0; i < leaderboard.length; i++) {
+      const entry = leaderboard[i];
+      if (typeof entry === "undefined") continue;
```
dtbuchholz: should this be?: `entry === undefined`

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -725,6 +727,69 @@ export async function getAgentPortfolioSnapshots(
...
+    // Union the queries and execute
+    const results = await unionAll(newestQuery, oldestQuery);
+
+    if (results.length === 0) {
+      return null;
+    }
+
+    // Sort results by timestamp desc to identify newest vs oldest
+    const sortedResults = results.sort(
+      // timestamp has default value of now() + is not nullable, so safe to use `!`
```
dtbuchholz: oooh, actually, the `portfolio_snapshots` table technically _doesn't_ explicitly set the `timestamp` column as non-nullable...it [does this](https://github.com/recallnet/js-recall/blob/main/apps/api/src/database/schema/trading/defs.ts#L180):
```ts
timestamp({ withTimezone: true }).defaultNow()
```

so i think we should just fix the `portfolio_snapshots` table:
1. update the `timestamp` col definition to also include `notNull()`
2. then, we can remove the bang operator here

apps/api/src/database/schema/core/defs.ts
dtbuchholz: @joewagner ahh, now i remember why `competitions_leaderboard` isn't the right place to store `pnl`...this slipped my mind. 

aaron set up the schemas like so:
- `core` => competition-agnostic data, which includes `competitions` or `competitions_leaderboard` etc
- `trading` => trading-specific data....and this includes an "extension" to the `competitions` table called [`trading_competitions`](https://github.com/recallnet/js-recall/blob/main/apps/api/src/database/schema/trading/defs.ts#L36)

i think that's the right pattern here? it makes the queries a bit more painful, though. we'd have to refactor the following:
- keep the `totalAgents` in `competitions_leaderboard`
- create a new `traditing_competitions_leaderboard` table, which just stores the `pnl` value
```ts
// In apps/api/src/database/schema/trading/defs.ts
export const tradingCompetitionsLeaderboard = tradingComps.table(
  "trading_competitions_leaderboard", 
  {
    competitionsLeaderboardId: uuid("competitions_leaderboard_id")
      .primaryKey()
      .references(() => competitionsLeaderboard.id, {
        onDelete: "cascade",
        onUpdate: "cascade",
      }),
    pnl: numeric("pnl", {
      precision: 30,
      scale: 15,
      mode: "number",
    }).notNull().default(0),
  },
  (table) => [
    index("idx_trading_competitions_leaderboard_pnl").on(table.pnl),
  ],
);
```

dtbuchholz:
@joewagner okay, this all looks solid. the only point of contention is which table to store this data in. if you have a strong opinion on keeping the current design, lmk.

apps/api/src/database/schema/core/defs.ts
```diff
@@ -319,6 +320,20 @@ export const competitionsLeaderboard = pgTable(
...
+      precision: 30,
+      scale: 15,
+      mode: "number",
+    })
+      .notNull()
+      .default(0),
+    totalAgents: numeric("total_agents", {
```
dtbuchholz: this can be an `integer` since it's a whole number value

apps/api/src/database/schema/core/defs.ts
joewagner: Ok, interesting.  This works for me.  Maybe we can put roi data there too per #617 (not sure if that makes sense yet.)

apps/api/src/services/competition-manager.service.ts
joewagner: I _think_ both are fine.  This is a hold over from js days.  I'll change it here, but if curious https://wtfjs.com/wtfs/2010-02-15-undefined-is-mutable

apps/api/src/database/schema/core/defs.ts
dtbuchholz: good call, yeah this setup will be perfect for that.

apps/api/src/database/schema/trading/defs.ts
```diff
@@ -53,6 +53,29 @@ export const tradingCompetitions = tradingComps.table(
   ],
 );
 
+/**
+ * Table to hold stats on trading specific competitions
+ */
+export const tradingCompetitionsLeaderboard = tradingComps.table(
```
joewagner, line 59: @dtbuchholz here is the new table that keeps trading specific comp stats separate from general comp stats.

apps/api/src/database/schema/trading/defs.ts
```diff
@@ -177,7 +200,7 @@ export const portfolioSnapshots = tradingComps.table(
     id: serial().primaryKey().notNull(),
     agentId: uuid("agent_id").notNull(),
     competitionId: uuid("competition_id").notNull(),
-    timestamp: timestamp({ withTimezone: true }).defaultNow(),
+    timestamp: timestamp({ withTimezone: true }).defaultNow().notNull(),
```
joewagner, line 203: Added not null, removing the requirement for `!` during sort.

apps/api/src/services/competition-manager.service.ts
joewagner: fixed

apps/api/src/database/repositories/competition-repository.ts
joewagner: fixed

apps/api/src/database/schema/core/defs.ts
joewagner: fixed

joewagner (commented):
The new table has been added and fixed conflicts with main.  Ready for another :eye:

PR #741, opened 2025-06-25
fix: update format check in CI

mzkrasner:
# Summary

- Updates format check logic in code-quality CI workflow file


dtbuchholz:
note: i noticed the same thing in that other PR wrt api-sdk causing it to fail, so im working that script

PR #743, opened 2025-06-26
fix(ci): format scripts and CI checks

dtbuchholz:


.github/workflows/code-quality.yml
```diff
@@ -64,9 +64,6 @@ jobs:
         env:
           SPEAKEASY_API_KEY: "${{ secrets.SPEAKEASY_API_KEY }}"
 
-      - name: Install eslint cli
```
dtbuchholz, line 67: not needed anymore. it's a root dev dep.

apps/comps/components/user-agents/index.tsx
```diff
@@ -113,9 +113,7 @@ export default function UserAgentsSection({ agents }: { agents: Agent[] }) {
           </Button>
         </div>
       </CollapsibleTrigger>
-      <CollapsibleContent className="w-full">
-        {agentList}
-      </CollapsibleContent>
+      <CollapsibleContent className="w-full">{agentList}</CollapsibleContent>
```
dtbuchholz, line 116: there are a few diffs are due to the `format` rules being applied properly, but the GH workflow changes should now prevent this from happening.

packages/api-sdk/package.json
```diff
@@ -27,7 +27,9 @@
     "lint": "eslint --cache --max-warnings=0 src",
     "build:mcp": "bun src/mcp-server/build.mts",
     "build": "npm run build:mcp && tshy",
-    "prepublishOnly": "npm run build"
+    "prepublishOnly": "npm run build",
+    "format:check": "prettier --check . --ignore-path=../../.prettierignore",
```
dtbuchholz, line 31: we had to add a post-processing script for `package.json` to make this possible...we'll might need something similar for the package publishing workflows.

scripts/build-api-sdk.sh
dtbuchholz, line 1: moved the old `api-sdk` script here and made a few tweaks for the `package.json` and format scripts

.github/actions/api-sdk-check/action.yml
```diff
@@ -20,10 +20,7 @@ runs:
     - name: Generate API SDK
       shell: bash
       run: |
-        set -ex
-        (cd apps/api && npm run generate-openapi)
-        (cd packages/api-sdk && speakeasy run --skip-versioning)
-        pnpm format
```
dtbuchholz, line 26: _this_ was our root cause. every time we run the Code Quality check, we were formatting the repo. the `format:check` script happens next, and it always passes because of this step.

build-api-sdk.sh
```diff
@@ -1,13 +0,0 @@
...
-
-cd apps/api
-pnpm run generate
-cd ../..
-
-cd packages/api-sdk
-speakeasy run --skip-versioning
-cd ../..
-
-pnpm format
```
dtbuchholz, line 13: this is the same thing as above. we now use the same `scripts/build-api-sdk.sh` in both CI and locally...just to reduce any differences in behavior.

scripts/build-api-sdk.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -ex
+
+# Generate the OpenAPI spec and the API SDK
+(cd apps/api && pnpm run generate)
+
+# If the openapi spec has changed, we need to regenerate the SDK
+if git status -s | grep -q "apps/api/openapi/openapi.json"; then
```
dtbuchholz, line 9: i debated putting this check in the CI workflow instead of here. but, i figured we should avoid regen-ing it whenever possible since we still use the free plan (e.g., local regen).

dtbuchholz:
example of this working with ill formatted files: https://github.com/recallnet/js-recall/pull/745

turbo.json
```diff
@@ -21,20 +21,22 @@
     "format": {
       "dependsOn": ["^format"],
       "inputs": [
-        "**/*.{js,jsx,ts,tsx,json,md,mdx}",
+        "**/*.{js,jsx,ts,tsx,json,md,mdx,yml,yaml}",
```
dtbuchholz, line 24: note: idk if inputs/outputs actually does anything...might remve

joewagner (approved):
Nice!

PR #747, opened 2025-06-26
fix(register): handle skills as array

dtbuchholz:


dtbuchholz:
this seemed to fix it locally—i can edit/remove the skills on this page
Screenshot 2025-06-26 at 11 06 16 AM

PR #752, opened 2025-06-26
fix(register): remove legacy WalletConnect chain ID from local storage

dtbuchholz:


dtbuchholz:
to test, have the backend API and the registration app running both in dev mode:
1. check out an older commit: `01692a3d39dbbad49790789ae63852db934e678e`
2. delete the local storage value for `@appkit/active_caip_network_id` and refresh the app page
3. check out `main`, and you'll see the WalletConnect modal popup
4. check out this branch, and it'll go away

mzkrasner (approved):

PR #757, opened 2025-06-26
fix: try to assist deep linking for mobile devices

mzkrasner:
# Summary

- Attempt to try and manage behavior between screens on mobile in registration app


mzkrasner:
@dtbuchholz also added in a try to force network switching on mobile

dtbuchholz (approved):

PR #758, opened 2025-06-26
fix(comps): build errors with experimental mode

dtbuchholz:
The staging app is unable to load data from the backend API. This is because the build script's `--experimental-build-mode compile` will prevent the app from reading the `NEXT_PUBLIC_API_BASE_URL` env var, so then all API calls fail. i.e., the API client defaults to the internal `/api` if the env var isn't present (see the logs below).

This PR fixes the errors by reverting the experimental build mode, and it also removes the internal `/api` fallback (legacy for when we had to mock data).

Also, the `AgentsSummary` card was throwing an error on build because the `bestPlacement.competitionId` is nullable, so we have to use the component composition pattern to conditionally render it.

Reference:
Screenshot 2025-06-26 at 6 09 24 PM


dtbuchholz:
(ah, now i see what you were talking about, @ramirotw...hm)

dtbuchholz:
gonna figure this out in a separate PR

PR #766, opened 2025-06-27
fix: attempt to resolve registration signing on mobile with forced reload

mzkrasner:
# Summary

Implemented mobile-specific enhancements to address WalletConnect deep linking issues that prevent mobile wallet apps from properly opening for SIWE (Sign-In with Ethereum) authentication.

## Problem Statement
- Mobile users could connect their wallets successfully
- However, clicking "Sign Message" would not trigger the mobile wallet app to open
- This is a known limitation with WalletConnect v2 on mobile browsers
- Users experienced a "dead-end" where the sign button appeared but did nothing

### 1. Mobile Detection Utilities (`lib/mobile-utils.ts`)
- **Purpose**: Reliable mobile device detection for conditional behavior
- **Features**: 
  - SSR-safe implementation with `typeof window` checks
  - Comprehensive mobile device regex patterns
  - Separate iOS and Android detection functions

### 2. Mobile Wallet Enhancer (`components/mobile-wallet-enhancer.tsx`)
- **Purpose**: Automatically reload page after wallet connection to prime WalletConnect deep links
- **Key Features**:
  - **Precise connection detection**: Uses `useRef` to track `false → true` transitions, preventing flickering issues
  - **Mobile-only behavior**: Only affects mobile devices, desktop unchanged
  - **Session storage safeguards**: Prevents infinite reload loops
  - **Optimized timing**: 1000ms delay for stable connection establishment
  - **Automatic cleanup**: Removes session flags after 3 seconds

### 3. Mobile Sign Helper (`components/mobile-sign-helper.tsx`)
- **Purpose**: Provide user guidance and retry options during mobile authentication
- **Key Features**:
  - **Smart visibility**: Only appears on mobile during authentication attempts
  - **User guidance**: Clear instructions for manual wallet app navigation
  - **Retry functionality**: "Retry Signing" button for failed attempts
  - **Auto-dismiss**: Disappears after 15 seconds or successful authentication
  - **Non-intrusive**: Positioned as overlay notification

### DIFF
--- a/apps/registration/components/mobile-sign-helper.tsx
+++ b/apps/registration/components/mobile-sign-helper.tsx
@@ -1,0 +1,90 @@
+"use client";
+
+import { useEffect, useState } from "react";
+import { useAccount } from "wagmi";
+
+import { useUserSession } from "@/hooks/useAuth";
+import { isMobileDevice } from "@/lib/mobile-utils";
+
+/**
+ * Component that provides mobile-specific signing guidance
+ */
+export function MobileSignHelper() {
+  const { isConnected } = useAccount();
+  const userSession = useUserSession();
+  const [showMobileHelp, setShowMobileHelp] = useState(false);
+
+  useEffect(() => {
+    // Only show on mobile when user is connected but not authenticated
+    const isAuthenticating =
+      isConnected && userSession.isInitialized && !userSession.isAuthenticated;
+
+    if (isMobileDevice() && isAuthenticating) {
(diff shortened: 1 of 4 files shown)


dtbuchholz:
hm this doesnt seem to solve the browser -> scan QR -> mobile signing.

dtbuchholz (approved):
ah my comment above is for the browser + mobile flow. we can give this a try

PR #767, opened 2025-06-27
fix: remove description from update user profile request

mzkrasner:
# Summary

- On agent registration form, the description was incorrectly assigned to the user, causing a query failure
- This has been eliminated and corrected to apply to the agent
- Also removes unused user description on the prior view


dtbuchholz (approved):

dtbuchholz:
note: the logic itself works...approving

PR #770, opened 2025-06-27
feat: alllow for join on admin search for agents

### LINKED ISSUE #742: [FEATURE] Enhance `/api/admin/search` API for user + agent combination
**Problem Statement**
The admin API [`/api/admin/search`](https://github.com/recallnet/js-recall/blob/main/apps/api/src/routes/admin.routes.ts#L1405)needs to let our frontend app:
- Get the API key for a specific agent by searching for: a user's wallet address + agent name.

The API has a `searchType` of `user`, `agent`, or `both`. If `both`, it'll treat the results as independent of one another, but we need to allow for a JOIN-like query. E.g., the frontend wants to query for an agent's name + a user's wallet address. The current logic would give you all of the agents across the whole app with that name or wallet address, and all of the users with that name or wallet address.

Also, scoping the searches like this might make it simpler to use:
```
?user.walletAddress="0x..."&agent.name="bar"
```

See the existing implementation in [`searchUsersAndAgents`](https://github.com/recallnet/js-recall/blob/main/apps/api/src/controllers/admin.controller.ts#L1126).

**Proposed Solution**
Enhance the query logic and params in the existing `/api/admin/search` API to allow for user + agent searches, akin to a JOIN instead of independent combinations:
- Add a new `searchType` called `join` (and make it the default). 
   - Instead of "independent" queries, the default behavior should join across the `users` and `agents`—specifically, the frontend needs to query for: the user's `walletAddress` _and_ for an agent's `name` (i.e., the `user.userId == agent.ownerId`.
- The existing `searchType` named `both` can keep the same "independent" logic, in case an admin _does_ want to get "all users and agents with a specific name", irrespective of any table relationship. We just want to remove it from being the default.
- Optional: Allow for `user.<field>` or `agent.<field>` query parameters, like `?user.walletAddress="0x..."&agent.name="bar"`

**Impact**
Admin API needs it—likely, for both the microsite and comps app

**Priority**
Very high

brunocalza:
Closes #742 

This PR adds one more behavior to `/api/admin/search`. You can know `join` agents with users by providing `join` value to `searchType`. 

Adds a custom query parser to be able to add the scope of the field with the dot notation, e.g. `?user.walletAddress="0x..."&agent.name="bar"`, to the query string.

This change is backwards-compatible. That means the current frontend shouldn't break. 
EDIT: actually it can break because the default is now `join`. And `join` only works correctly if you use the dot notation. Moving the default back to `both`.


apps/api/e2e/tests/admin.test.ts
```diff
@@ -675,6 +692,67 @@ describe("Admin API", () => {
       );
     }
 
+    // TEST CASE 6: Join query - find agents by name that are owned by a user with specific wallet address
```
brunocalza: adds a new test case where a new agent is added to user 1 and we do a join

apps/api/src/controllers/admin.controller.ts
```diff
@@ -1134,23 +1154,35 @@ export function makeAdminController(services: ServiceRegistry) {
           name,
           walletAddress,
           status,
+          user,
+          agent,
           searchType, // 'users', 'agents', or 'both' (default)
         } = req.query;
 
-        const searchTypeFilter = (searchType as string) || "both";
+        const searchTypeFilter = (searchType as string) || "join";
```
brunocalza: default is now join

apps/api/src/controllers/admin.controller.ts
```diff
@@ -1134,23 +1154,35 @@ export function makeAdminController(services: ServiceRegistry) {
...
         const results: AdminSearchResults = {
           users: [],
           agents: [],
         };
 
         // Search users if requested
-        if (searchTypeFilter === "users" || searchTypeFilter === "both") {
```
brunocalza, line 1147: at this point "join" behaves the same a "both"

apps/api/src/controllers/admin.controller.ts
```diff
@@ -1202,6 +1246,21 @@ export function makeAdminController(services: ServiceRegistry) {
           }));
         }
 
+        if (searchTypeFilter === "join") {
```
brunocalza: here we perform a join by filtering only the users we got in `results.users`

apps/api/src/middleware/query-parser.middleware.ts
```diff
@@ -0,0 +1,12 @@
+import { NextFunction, Request, Response } from "express";
+import * as qs from "qs";
+
+export const queryParserMiddleware = () => {
```
brunocalza: we add a custom query parser middleware that parses the **dot**. that allows a more structured way of accessing the fields and values, e.g. `user.name=Bruno&agent.name=Agent` would be accessed as

```javascript
const {
  user: { name: userName } = {},
  agent: { name: agentName } = {},
} = req.query;
```

apps/api/src/controllers/admin.controller.ts
```diff
@@ -70,6 +70,26 @@ interface AdminSearchResults {
   agents: Omit<Agent, "apiKey">[];
 }
 
+interface AdminSearchUsersAndAgentsQuery {
+  // backwards-compatibility
```
brunocalza: keep the old way of querying to not break current frontend

apps/api/src/controllers/admin.controller.ts
brunocalza: the assumption here is that join means: `agent JOIN user`. If that's to limiting, we can expand to `ljoin` (left join) and `rjoin` (right join) so we can also do `user JOIN agent`.

dtbuchholz (approved):
@brunocalza assuming you're good with https://github.com/recallnet/js-recall/pull/788, feel free to merge that branch into this one, and then this into main

PR #771, opened 2025-06-27
fix(api): performance improvements

joewagner:
part of #620 

There are three groups of changes here to help with performance.
- Better database indexing
- Batching of dexscreener api calls 
- lower overhead in sql query result processing

There more improvements to be made, but this seems like a digestable size for a PR and will allow us to do more stress testing and iterate.

### DIFF
--- a/apps/api/drizzle/0021_little_photon.sql
+++ b/apps/api/drizzle/0021_little_photon.sql
@@ -1,0 +1,1 @@
+CREATE INDEX "idx_portfolio_snapshots_competition_agent_timestamp" ON "trading_comps"."portfolio_snapshots" USING btree ("competition_id","agent_id","timestamp");
--- a/apps/api/src/controllers/agent.controller.ts
+++ b/apps/api/src/controllers/agent.controller.ts
@@ -400,12 +400,19 @@ export function makeAgentController(services: ServiceRegistry) {
         let totalValue = 0;
         const tokenValues = [];
 
-        // Calculate values with minimal API calls
+        // Get unique token addresses
+        const uniqueTokens = [
+          ...new Set(balances.map((balance) => balance.tokenAddress)),
+        ];
+
+        // Get token info in bulk, this chunks dexscreener api requests to
+        // reduce load and rate limiting.
+        const tokenInfoMap =
+          await services.priceTracker.getBulkTokenInfo(uniqueTokens);
+
+        // Calculate values efficiently
         for (const balance of balances) {
-          // Get price and token information using the existing service method
-          const tokenInfo = await services.priceTracker.getTokenInfo(
-            balance.tokenAddress,
-          );
+          const tokenInfo = tokenInfoMap.get(balance.tokenAddress);
           const price = tokenInfo?.price || 0;
           const value = price ? balance.amount * price : 0;
           totalValue += value;
--- a/apps/api/src/database/repositories/competition-repository.ts
+++ b/apps/api/src/database/repositories/competition-repository.ts
@@ -10,6 +10,7 @@ import {
   isNotNull,
   lte,
   max,
+  min,
   sql,
 } from "drizzle-orm";
 import { unionAll } from "drizzle-orm/pg-core";
@@ -689,6 +690,8 @@ async function batchCreatePortfolioTokenValuesImpl(
  */
 async function getLatestPortfolioSnapshotsImpl(competitionId: string) {
   try {
+    // TODO: this query seems to be averaging almost 2 full seconds.
+    //  We added indexes on Jul 10, need to find out if indexing fixes the issue.
     const subquery = db
       .select({
         agentId: portfolioSnapshots.agentId,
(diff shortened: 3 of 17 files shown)


apps/api/drizzle/0021_little_photon.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX "idx_portfolio_snapshots_competition_agent_timestamp" ON "trading_comps"."portfolio_snapshots" USING btree ("competition_id","agent_id","timestamp");
```
joewagner, line 1: This is an index recommended by supabase

apps/api/src/controllers/agent.controller.ts
```diff
@@ -400,12 +400,19 @@ export function makeAgentController(services: ServiceRegistry) {
...
 
-        // Calculate values with minimal API calls
+        // Get unique token addresses
+        const uniqueTokens = [
+          ...new Set(balances.map((balance) => balance.tokenAddress)),
+        ];
+
+        // Get token info in bulk, this chunks dexscreener api requests to
+        // reduce load and rate limiting.
+        const tokenInfoMap =
```
joewagner, line 410: See below for implementation. This should reduce dexscreener api calls significantly.

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -680,6 +681,8 @@ async function createPortfolioTokenValueImpl(
  */
 async function getLatestPortfolioSnapshotsImpl(competitionId: string) {
   try {
+    // TODO: this query seems to be averaging almost 2 full seconds.
+    //  We added indexes on Jul 10, need to find out if indexing fixes the issue.
```
joewagner, line 694: note here that the new indexes might make this query acceptable, but we might need to revisit again depending on performance reports.

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -769,6 +772,100 @@ async function getBulkAgentPortfolioSnapshotsImpl(
...
+        earliestSubquery,
+        and(
+          eq(portfolioSnapshots.agentId, earliestSubquery.agentId),
+          eq(portfolioSnapshots.timestamp, earliestSubquery.minTimestamp),
+        ),
+      )
+      .where(eq(portfolioSnapshots.competitionId, competitionId));
+
+    // Get snapshots closest to 24h ago using window functions
+    const snapshots24hAgo = await db
```
joewagner, line 831: This should be decent perf, but like above, we need to be vigilant about looking at the db perfomance monitoring.

apps/api/src/database/repositories/price-repository.ts
```diff
@@ -40,6 +40,41 @@ async function createImpl(priceData: InsertPrice) {
   }
 }
 
+/**
+ * Create multiple price records in a single batch operation
+ * @param pricesData Array of price data to store
+ * @returns Array of created price records
+ */
+export async function createBatch(pricesData: InsertPrice[]) {
```
joewagner, line 48: With the change to getting batches of tokens, we also need to enable the individual storePrice function to be able to store batches of prices.

apps/api/src/database/schema/trading/defs.ts
```diff
@@ -214,6 +214,11 @@ export const portfolioSnapshots = tradingComps.table(
       table.competitionId,
     ),
     index("idx_portfolio_snapshots_timestamp").on(table.timestamp),
+    index("idx_portfolio_snapshots_competition_agent_timestamp").on(
```
joewagner, line 217: new index you'll see in the migration file.

apps/api/src/services/balance-manager.service.ts
```diff
@@ -169,6 +170,59 @@ export class BalanceManager {
     }
   }
 
+  /**
+   * Get all balances for multiple agents in bulk
+   * @param agentIds Array of agent IDs
+   * @returns Array of Balance objects for all agents
+   */
+  async getBulkBalances(agentIds: string[]) {
```
joewagner, line 178: This is used by the new logic in getLeaderboard to avoid n+1 db queries in a loop

apps/api/src/services/competition-manager.service.ts
```diff
@@ -691,75 +691,49 @@ export class CompetitionManager {
...
       );
 
-      // Group snapshots by agent ID with proper typing
-      const snapshotsByAgent = new Map<string, typeof allSnapshots>();
-      for (const snapshot of allSnapshots) {
-        const agentSnapshots = snapshotsByAgent.get(snapshot.agentId) || [];
-        agentSnapshots.push(snapshot);
-        snapshotsByAgent.set(snapshot.agentId, agentSnapshots);
-      }
+      // Create maps for efficient lookup
```
joewagner, line 706: Reducing the size of the response to getting snapshots from the db, and also reducing in-memory work to put results into a `Map`

apps/api/src/services/price-tracker.service.ts
```diff
@@ -14,7 +15,7 @@ import { BlockchainType, PriceSource, SpecificChain } from "@/types/index.js";
 export class PriceTracker {
   providers: PriceSource[];
   // private novesProvider: NovesProvider | null = null;
-  private multiChainProvider: MultiChainProvider | null = null;
+  private multiChainProvider: MultiChainProvider;
```
joewagner, line 18: Cleaned up types here, so we don't have to check for null.

apps/api/src/services/price-tracker.service.ts
```diff
@@ -247,11 +248,148 @@ export class PriceTracker {
   }
 
   /**
-   * Store price data in the database
+   * Get detailed token information for multiple tokens in bulk
+   * @param tokenAddresses Array of token addresses to get info for
+   * @returns Map of token address to token info (null if not available)
+   */
+  async getBulkTokenInfo(
```
joewagner, line 255: This wraps the batched calls to the token data provider

apps/api/src/services/providers/dexscreener.provider.ts
```diff
@@ -515,4 +516,266 @@ export class DexScreenerProvider implements PriceSource {
...
+      // Merge results
+      batchResults.forEach((result, address) => {
+        results.set(address, result);
+      });
+    }
+
+    // After batch processing, retry null values individually. This is needed
+    // in case the dexscreener api returns the wrong quote token and
+    // miscalculates priceUsd
+    const nullTokens: string[] = [];
```
joewagner, line 554: As a fallback, this reverts to the old way of getting prices one-by-one for any tokens that we couldn't get price for in bulk.  This should only be executed if the dexscreener api doesn't give us the priceUsd value that is the actual price in USD.

apps/api/src/services/providers/dexscreener.provider.ts
```diff
@@ -515,4 +516,266 @@ export class DexScreenerProvider implements PriceSource {
...
+    if (
+      uncachedAddresses.length === 0 ||
+      typeof uncachedAddresses[0] === "undefined"
+    ) {
+      return results;
+    }
+
+    const firstAddress = uncachedAddresses[0];
+
+    const dexScreenerChain = this.getDexScreenerChain(
```
joewagner, line 654: This method call is kinda strange.  The first two args aren't used, but I didn't want to change this since it's not part of this PR.  The `firstAddress` variable isn't really needed except to keep tsx happy.

apps/api/src/services/providers/dexscreener.provider.ts
```diff
@@ -515,4 +516,266 @@ export class DexScreenerProvider implements PriceSource {
...
+      firstAddress,
+      chain,
+      specificChain,
+    );
+
+    // Join addresses with comma for batch request
+    const addressesParam = uncachedAddresses
+      .map((addr) => addr.toLowerCase())
+      .join(",");
```
joewagner, line 663: We can send comma delim list of tokens.  This is the new api call to dexscreener.

apps/api/src/services/providers/dexscreener.provider.ts
```diff
@@ -515,4 +516,266 @@ export class DexScreenerProvider implements PriceSource {
...
+    specificChain: SpecificChain,
+  ): { price: number; symbol: string } | null {
+    // Find the pair where our token is the base token
+    const result = data.find(
+      (pair) => pair.baseToken?.address?.toLowerCase() === addr.toLowerCase(),
+    );
+
+    if (
+      result?.quoteToken &&
+      !this.isStablecoin(result.quoteToken.address, specificChain)
```
joewagner, line 764: This is the logic to catch invalid `priceUsd` values.  If this function returns null, later on in processing it will trigger a re-request of the token value using the one-by-one strategy.

apps/api/src/types/index.ts
```diff
@@ -122,6 +122,48 @@ export interface PriceSource {
   ): Promise<boolean>;
 }
 
+// DexScreener API interfaces
```
joewagner, line 125: Added these types to match responses from dexscreener.

joewagner:
Just pushed bugbot fixes.   Should be gtg now.

joewagner:
> lgtm. might want to have @mzkrasner's eyes on this too since he's been closer to the dexscreener related APIs

Good call.  I'll wait for Mark to tal.

mzkrasner (approved):

Issue #775, opened 2025-06-30
[BUG] Profile skeleton mismatch with design


Rcontre360:
Current:

![Image](https://github.com/user-attachments/assets/730e250b-2f99-42b7-b93b-41b4bc56f509)

Design:

![Image](https://github.com/user-attachments/assets/9ae9d711-9f1c-4c41-9031-4080364bc50e)

PR #776, opened 2025-06-30
fix: accurate profile skeleton

Rcontre360:
#775

### DIFF
--- a/apps/comps/components/agent-profile/index.tsx
+++ b/apps/comps/components/agent-profile/index.tsx
@@ -85,12 +85,15 @@ export default function AgentProfile({
       }
     };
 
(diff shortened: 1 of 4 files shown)


ramirotw (changes requested):
lgtm!

the only thing I'd split the PR into 2: one containing the Agent Profile fixes and the other with the Profile skeleton.

There are also other issues with the Competitions table in the Agent Profile:

1. layout breaks when the Claim button appears (it should be hidden as well, at least until we implement it)
image

2. alignment is off for several columns, please compare it with the mockups
image
image

3. Portfolio column needs to be hidden on screens < md

ramirotw (approved):

PR #788, opened 2025-06-30
refactor(api): admin search API

dtbuchholz:


apps/api/src/controllers/request-helpers.ts
```diff
@@ -125,3 +128,35 @@ export function buildPaginationResponse(
...
+  if (url.includes("?")) {
+    const queryString = url.split("?")[1];
+    if (queryString) {
+      result = qs.parse(queryString, { allowDots: true });
+    }
+  }
+  if (!result.user && !result.agent) {
+    throw new ApiError(
+      400,
+      "Invalid request format: missing user or agent query params",
```
brunocalza: nit: missing user and agent query params

brunocalza (approved):
Looks great! Nice work. 

I wonder if `both` is really needed because the "both" logic is applied if `agent` and `user` is present, you don't need to check `searchType`. With that, we would only need a boolean to indicate if we want to join or not. I don't feel strongly about making this change though, just a comment.

apps/api/src/controllers/request-helpers.ts
dtbuchholz: changed it to `must provide user or agent search parameters` so it's clearer.

but, you _can_ pass `agent` or `user`, and you dont need to provide both at the same time. so, i think "or" is semantically correct?

brunocalza: sounds better

dtbuchholz:
> I wonder if both is really needed because the "both" logic is applied if agent and user is present, you don't need to check searchType. With that, we would only need a boolean to indicate if we want to join or not. I don't feel strongly about making this change though, just a comment.

@brunocalza yeah i thought about that too...kk lemme make the change real quick

apps/api/e2e/tests/admin.test.ts
```diff
@@ -755,6 +755,74 @@ describe("Admin API", () => {
     await adminClient.deleteAgent(user3Result.agent!.id);
   });
 
+  test("should use 'join' with different query param formats", async () => {
```
dtbuchholz, line 758: @brunocalza alrighty, i dropped the `searchType` and replaced it with `join`. the different flows are:
1. no `join` => `false` (aka just give all results, like `searchType=both`)
2. `join=false` => `false`
2. `join=true` => `true`
3. `join` => `true` (aka naked `join` is allowed)

Issue #701, opened 2025-06-23
[IMPL] Update Agent registration to create and display both prod and sandbox API keys


stbrody:
**DoD**

- [x] Set up admin sandbox url as env var (i.e., we need both a "production" and "sandbox" API URL) & admin API key
- [x] When a user is created, call the `POST /admin/users` and `POST /admin/agents` endpoints against the Sandbox API (in addition to calling them against the production API as we already do)
- [x] In the UI, after an agent is created, we need to also display their sandbox API key alongside the production API key
- [x] In the UI, if a user requests their agent's info, we need to fetch the API key and display it
- [x] Optional: use an env var to allow an internal dev run the comps app with _only_ one "production" server instead of both a "production" and "sandbox" server

See https://github.com/recallnet/js-recall/issues/609 for more context

stbrody:
Depends on designs from https://github.com/recallnet/js-recall/issues/609

dtbuchholz:
see reference https://github.com/recallnet/js-recall/issues/708

PR #790, opened 2025-06-30
feat(comps): add user/agent sandbox flow

ramirotw:
Fixes https://github.com/recallnet/js-recall/issues/701

To test this locally, you need to run two separate API instances, each connected to a different database. To make this easier, a new `.env.sandbox` file was added, it mainly changes the port and database URL.

1. Start the sandbox API using `NODE_ENV=sandbox pnpm dev`, alongside the regular API. Then grab the Admin API key from the sandbox environment.
2. Update the `/apps/comps/.env` file and set the `NEXT_PUBLIC_SANDBOX_API_URL` and `SANDBOX_ADMIN_API_KEY` environment variables to the correct values.
3. Create a new agent through the comps app. In the final step, the sandbox API key should be displayed. Then go to the Agent Profile page and verify that the Credentials section includes the Sandbox API key.


apps/api/.env.sandbox
```diff
@@ -0,0 +1,75 @@
+# Server Configuration
```
ramirotw, line 1: This was the simplest approach I found to run both APIs at the same time. Just running `pnpm dev` and `NODE_ENV=sandbox pnpm dev` does the trick. Happy to update this if there’s a better way

apps/comps/app/api/sandbox/api-key/route.ts
```diff
@@ -0,0 +1,75 @@
...
+  // Validate that the agent name belongs to the logged-in user
+  const userAgent = userAgentsData.agents.find(
+    (agent) => agent.name === agentName,
+  );
+  if (!userAgent) {
+    throw new Error("Agent not found or does not belong to user");
+  }
+
+  // Find the agent's ID in the sandbox using its name
+  const searchData = await sandboxAdminRequest<AdminSearchResult>(
```
dtbuchholz, line 45: just a heads up—we merged https://github.com/recallnet/js-recall/pull/770 yesterday, which should help with this logic.

you can run a search like:
```
/search?user.walletAddress=0x1234&agent.name=foo&join`
```
this will:
1. get the user with that wallet address
2. get all agents with that name
3. with `join`, it'll effectively do a left join on the users table and give you the user's agent

we decided to get rid of the old/existing top-level search fields like `?name=...`. instead, just define the scope like `user.<field>` or `agent.<field`, and it'll return all results unless you add the `join` param to filter for a user's agents.

ramirotw: Fixed in https://github.com/recallnet/js-recall/pull/790/commits/f8da10896aef3820e143a1d2e49a4283112c89ab

apps/comps/app/api/sandbox/users/route.ts
```diff
@@ -0,0 +1,69 @@
...
+  const profileData = await mainApiRequest<ProfileResponse>(
+    "/user/profile",
+    sessionCookie,
+  );
+  const { user } = profileData;
+  const { walletAddress, name, email, imageUrl, metadata } = user;
+
+  // Check if user already exists in sandbox
+  const searchData = await sandboxAdminRequest<AdminSearchResult>(
+    `/admin/search?walletAddress=${encodeURIComponent(walletAddress)}&searchType=users`,
```
dtbuchholz:
```suggestion
    `/admin/search?user.walletAddress=${encodeURIComponent(walletAddress)}`,
```

dtbuchholz (approved):

apps/comps/app/api/sandbox/users/route.ts
ramirotw: Fixed in https://github.com/recallnet/js-recall/commit/c797d1b23bdcd78707566aacd78a77a7633f8bee

PR #791, opened 2025-06-30
feat(api): updating naming conventions to match designs

joewagner:
fixes: https://github.com/recallnet/js-recall/issues/780

This is a rename, so there's a lot of changes.  The biggest is the changes to table names.  The word "rank" now means 1st, 2nd, 3rd, and so on and the values associated with openskill calcs are called "score"


dtbuchholz:
@joewagner @mzkrasner in terms of db migrations, the agent rank and history tables will effectively be deleted and not migrated, right?

iirc you mentioned creating a script @joewagner to account for this? I just wanna double check what the plan is before we merge this so that we upgrade the server + manually migrate data at the same time.

dtbuchholz:
I suppose @brunocalza might have the original script he used to populate data in that agent rank table

apps/api/drizzle/0019_short_jane_foster.sql
```diff
@@ -0,0 +1,25 @@
...
+ALTER TABLE "agent_rank_history" RENAME TO "agent_score_history";--> statement-breakpoint
+ALTER TABLE "agent_score" DROP CONSTRAINT "unique_agent_rank_agent_id";--> statement-breakpoint
+ALTER TABLE "agent_score" DROP CONSTRAINT "agent_rank_agent_id_fkey";
+--> statement-breakpoint
+ALTER TABLE "agent_score_history" DROP CONSTRAINT "agent_rank_history_agent_id_fkey";
+--> statement-breakpoint
+ALTER TABLE "agent_score_history" DROP CONSTRAINT "agent_rank_history_competition_id_fkey";
+--> statement-breakpoint
+ALTER TABLE "object_index" ALTER COLUMN "data_type" SET DATA TYPE text;--> statement-breakpoint
+DROP TYPE "public"."sync_data_type";--> statement-breakpoint
```
dtbuchholz:
```suggestion
UPDATE "object_index" SET "data_type" = 'agent_score_history' WHERE "data_type" = 'agent_rank_history';--> statement-breakpoint
UPDATE "object_index" SET "data_type" = 'agent_score' WHERE "data_type" = 'agent_rank';--> statement-breakpoint
DROP TYPE "public"."sync_data_type";--> statement-breakpoint
```

ah _now_ i remember the problem you mentioned about scripts and backfilling data. 

try this and see if it works. it seemed to fix the migration problem for me locally:
1. checkout `main` and seed the db with at least 1 "ended" comp. data will exist in `agent_rank_history` & `agent_rank` as well as `object_index`
2. checkout this branch, and run the migration w/o these changes...this will error with `applying migrations...error: invalid input value for enum sync_data_type: "agent_rank_history"`
3. make the change above, and the migration works. data will exist in `agent_score_history` & `agent_score` as well as `object_index`

i guess drizzle + enums might be problematic, but logically, the diff above makes sense. just rename the existing values before dropping the type.

dtbuchholz (approved):

Issue #740, opened 2025-06-25
Add OKX wallet to app login


stbrody:
See https://recalllabs.slack.com/archives/C08TKN2T8LC/p1750863046945509 for more context.

Basically, we think there are good partnership and comarketing opportunities if we can include OKX wallet support in the app, alongside the other wallets we already support.

stbrody:
This is not highest priority - if this turns out to be really complicated or like its going to take more than 1 day of work, we should stop and check in if this is actually worth doing or not

PR #801, opened 2025-07-02
feat: OKX to the top

ukstv:
closes https://github.com/recallnet/js-recall/issues/740

### DIFF
--- a/apps/comps/wagmi-config.ts
+++ b/apps/comps/wagmi-config.ts
@@ -1,4 +1,9 @@
 import { getDefaultConfig } from "@rainbow-me/rainbowkit";
+import {
+  coinbaseWallet,
+  okxWallet,
+  walletConnectWallet,
+} from "@rainbow-me/rainbowkit/wallets";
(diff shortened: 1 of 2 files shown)


dtbuchholz (approved):

apps/comps/wagmi-config.ts
```diff
@@ -1,4 +1,9 @@
 import { getDefaultConfig } from "@rainbow-me/rainbowkit";
+import {
+  coinbaseWallet,
+  okxWallet,
+  walletConnectWallet,
```
dtbuchholz, line 5: i just noticed this behavior: if i select the WalletConnect option, exit w/o connecting, and refresh the page...we see some logged errors
Screenshot 2025-07-02 at 1 27 16 PM

dtbuchholz: it's only for WalletConnect, and i've seen this issue happen elsewhere. it's something we can't control, though, so we can probably keep it...unless you know of a quick fix

PR #802, opened 2025-07-02
fix(api): add multi-origin CORS support for docs.recall.network

derrekcoleman:
## Problem

The Swagger-based API documentation at `https://docs.recall.network/api-reference/endpoints/trade` was failing because the sandbox API server (`api.sandbox.competitions.recall.network`) only allows CORS requests from `register.recall.network`, not from the docs domain.

**Root Cause**: Commit a146e57 added restrictive CORS configuration that only accepts `config.app.url` (set via `FRONTEND_URL`), which currently points to the registration portal.

## Solution

This PR adds support for multiple CORS origins while maintaining backward compatibility:

- **New `CORS_ORIGINS` environment variable**: Accepts comma-separated list of allowed origins
- **Backward compatibility**: Falls back to `FRONTEND_URL` if `CORS_ORIGINS` is not set
- **Updated sandbox config**: Includes both `register.recall.network` and `docs.recall.network`

## Changes

- `apps/api/src/config/index.ts`: Add `corsOrigins` array configuration
- `apps/api/src/index.ts`: Update CORS middleware to accept multiple origins
- `apps/api/.env.sandbox`: Add `CORS_ORIGINS` with both domains
- `apps/api/.env.example`: Document new environment variable

## Alternative Approaches

This is one possible solution among several approaches. Y'all may rather:
- Different environment variable naming/structure
- Server-side configuration instead of environment variables  
- Different CORS handling strategy
- Infrastructure-level CORS configuration

Feel free to modify this approach or close this PR if a different solution is preferred!

## Testing

After deployment, the docs site should be able to successfully make API calls to test endpoints.

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -4,6 +4,7 @@ NODE_ENV=development
 
 # Frontend Configuration
 FRONTEND_URL=http://localhost:3001 # Enable CORS
+CORS_ORIGINS= # Optional: Comma-separated list of allowed CORS origins (overrides FRONTEND_URL if set)
 DOMAIN=.example.com # Cross-subdomain cookies (only set in production mode)
 SESSION_TTL=7200 # Session expiry (defaults to 2 hours)
 COOKIE_NAME=session # Session cookie name (defaults to `session`)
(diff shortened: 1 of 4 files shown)


dtbuchholz:
closing in favor of https://github.com/recallnet/js-recall/pull/803

PR #803, opened 2025-07-02
feat(api): flexible CORS

dtbuchholz:
- Allow multiple CORS origins so that the comps app and docs site can both make API requests
- Add CORS tests from 2 allowed domains, and a 3rd malicious domain

### DIFF
--- a/apps/api/src/config/index.ts
+++ b/apps/api/src/config/index.ts
@@ -162,16 +162,13 @@ export const config = {
   app: {
     // Session management and email verification settings for the frontend competitions app
     url: process.env.FRONTEND_URL || "http://localhost:3001", // TODO: resolve frontend/backend default ports
+    // CORS multi-origin requests and cookie domain
     domain: process.env.DOMAIN,
     cookieName: process.env.COOKIE_NAME || "session",
     sessionPassword:
       process.env.ROOT_ENCRYPTION_KEY ||
       "default_encryption_key_do_not_use_in_production",
     sessionTtl: parseInt(process.env.SESSION_TTL || "7200", 10),
-    // CORS origins for API calls (e.g., frontend app, docs site + OpenAPI UI, etc.)
-    corsOrigins: process.env.CORS_ORIGINS?.split(",") || [
-      process.env.FRONTEND_URL || "http://localhost:3001",
-    ],
   },
   database: {
     ssl: process.env.DB_SSL === "true",
--- a/apps/api/src/config/index.ts
+++ b/apps/api/src/config/index.ts
@@ -160,6 +160,7 @@ export const config = {
   },
   // Frontend app configuration for interfacing with the server
   app: {
+    // Session management and email verification settings for the frontend competitions app
     url: process.env.FRONTEND_URL || "http://localhost:3001", // TODO: resolve frontend/backend default ports
     domain: process.env.DOMAIN,
     cookieName: process.env.COOKIE_NAME || "session",
@@ -167,6 +168,10 @@ export const config = {
       process.env.ROOT_ENCRYPTION_KEY ||
       "default_encryption_key_do_not_use_in_production",
     sessionTtl: parseInt(process.env.SESSION_TTL || "7200", 10),
+    // CORS origins for API calls (e.g., frontend app, docs site + OpenAPI UI, etc.)
+    corsOrigins: process.env.CORS_ORIGINS?.split(",") || [
+      process.env.FRONTEND_URL || "http://localhost:3001",
+    ],
   },
   database: {
     ssl: process.env.DB_SSL === "true",
--- a/apps/api/src/index.ts
+++ b/apps/api/src/index.ts
@@ -87,7 +87,7 @@ app.set("trust proxy", true);
 
 app.use(
   cors({
-    origin: config.app.url,
+    origin: config.app.corsOrigins,
     credentials: true,
     methods: ["GET", "POST", "PUT", "DELETE", "OPTIONS"],
     allowedHeaders: ["Content-Type", "Authorization"],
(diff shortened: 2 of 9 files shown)


apps/api/src/config/index.ts
```diff
@@ -160,13 +160,18 @@ export const config = {
...
+    // Session management and email verification settings for the frontend competitions app
     url: process.env.FRONTEND_URL || "http://localhost:3001", // TODO: resolve frontend/backend default ports
     domain: process.env.DOMAIN,
     cookieName: process.env.COOKIE_NAME || "session",
     sessionPassword:
       process.env.ROOT_ENCRYPTION_KEY ||
       "default_encryption_key_do_not_use_in_production",
     sessionTtl: parseInt(process.env.SESSION_TTL || "7200", 10),
+    // CORS origins for API calls (e.g., frontend app, docs site + OpenAPI UI, etc.)
+    corsOrigins: process.env.CORS_ORIGINS?.split(",") || [
```
dtbuchholz: a new env var that we want to set is a comma separate list of URLs: `CORS_ORIGINS`. i've added a follow up reminder to do this on the actual servers: https://github.com/recallnet/js-recall/issues/804

the docs site makes requests to the production and sandbox servers, so we'll need to configure this value there.

if it's not set, we'll just default to the FRONTEND_URL or localhost defaults.

apps/api/src/index.ts
```diff
@@ -87,7 +87,18 @@ app.set("trust proxy", true);
...
+    origin: function (origin, fn) {
+      if (
+        !origin ||
+        config.app.url === origin ||
+        // Note: the frontend app URL above should already be included in the `corsOrigins` config
+        config.app.corsOrigins.includes(origin)
+      ) {
+        fn(null, true);
+      } else {
+        fn(new ApiError(403, "Forbidden"));
```
dtbuchholz, line 111: some ai-gen'd rationale here, but it seems valid:
1. if a "bad" CORS origin makes a request, we were returning a 404 instead of 403
2. if a "bad" CORS origin makes a preflight request, we leaked a _little_ information for the available methods, headers, and credentials values below.

joewagner: Sorry for the early review, but just figured I'd throw out the docs for the cors package seem like the describe our usecase here: https://www.npmjs.com/package/cors#configuring-cors-asynchronously

```
Configuring CORS Asynchronously
var express = require('express')
var cors = require('cors')
var app = express()
 
var whitelist = ['http://example1.com', 'http://example2.com']
var corsOptionsDelegate = function (req, callback) {
  var corsOptions;
  if (whitelist.indexOf(req.header('Origin')) !== -1) {
    corsOptions = { origin: true } // reflect (enable) the requested origin in the CORS response
  } else {
    corsOptions = { origin: false } // disable CORS for this request
  }
  callback(null, corsOptions) // callback expects two parameters: error and options
}
 
app.get('/products/:id', cors(corsOptionsDelegate), function (req, res, next) {
  res.json({msg: 'This is CORS-enabled for a whitelisted domain.'})
})
 
app.listen(80, function () {
  console.log('CORS-enabled web server listening on port 80')
})
```

In our case, instead of the `whitelist` Array we could just check the domain is either recall.network (or localhost if not in prod)

dtbuchholz: @joewagner nice...yeah that's probably easier than making things so explicit

joewagner: I figure if the docs suggest this, it's probably a reasonable way to do it.  Also, it seems like what we really want is to wildcard all subdomains (which isn't a feature of the CORS spec exactly).

dtbuchholz: Ah so we could probably do that and basically regex the origin...sorta combining both setups https://expressjs.com/en/resources/middleware/cors.html#configuring-cors-w-dynamic-origin

Lemme refactor

derrekcoleman:
@stephhuynh18 @stbrody is it possible to prioritize this fix? The interactive endpoints in our docs are all still broken here in the second week of the hackathon 🙏

apps/api/src/index.ts
```diff
@@ -88,12 +88,24 @@ app.set("trust proxy", true);
...
   cors({
     origin: function (origin, fn) {
-      if (
-        !origin ||
-        config.app.url === origin ||
-        // Note: the frontend app URL above should already be included in the `corsOrigins` config
-        config.app.corsOrigins.includes(origin)
-      ) {
+      // Allow any localhost port in development mode
+      if (config.server.nodeEnv === "development") {
```
dtbuchholz, line 92: 1. we check for `NODE_ENV=development` so that `NODE_ENV=test` can use the `DOMAIN` variable in this CORS logic. (note: `DOMAIN=.example.com`)
2. if dev mode, allow any localhost origin
3. if not dev mode, enforce the origin is the primary or subdomain

the docs link below outlines the regex pattern:
> the pattern `/example\.com$/` will reflect any request that is coming from an origin ending with “example.com”.

dtbuchholz:
🤘

note: i also tested this with `cloudflared tunnel --url http://127.0.0.1:<PORT>` and altered the `.env` files in the `apps/api` and `apps/comps` directories, and things worked as expected:

if you're curious, in `apps/api`:
- `NODE_ENV=foo` (to trigger the flexible HTTPS CORS for non-development envs)
- `DOMAIN=.trycloudflare.com`
- `API_DOMAIN=...` (the value from `cloudflared` and this API server)
- `FRONTEND_URL=...` (the value from the directory blow: running `cloudflared` for the comps frontend)

and in `apps/comps`:
- `NEXT_PUBLIC_API_BASE_URL=...` (the `cloudflared` value `API_DOMAIN` above with `/api` appended)

and then you can run `cloudflared` with the docs or any other site, and it'll just work.

Issue #794, opened 2025-07-01
[FEATURE] Changes on agent profile 01/07


Rcontre360:
Detail fixes on agent profile:

- [x] Verify badge color and thickness
- [x] "developed by" color
- [x] vertical alignment "This agent hasn't earned..."
- [x] Best placement component, polish it and modify it to be able to reuse it
- [x] size of headers, there are some with text-xs and others with nothing. Example "COMPLETED COMPS" vs "AGENT DESCRIPTION"
- [x] agent text color description (primary)
- [x] comps table filters are buggy

Design changes

- [x] Rewards on tables when user is owner
- [x] Agent address position change

ramirotw:
Also the following changes https://github.com/recallnet/js-recall/issues/478#issuecomment-3024980953

PR #805, opened 2025-07-02
Rafael/agent profile

Rcontre360:
Changes on #794

### DIFF
--- a/packages/ui2/src/components/breadcrumb.tsx
+++ b/packages/ui2/src/components/breadcrumb.tsx
@@ -1,14 +1,13 @@
-import {SlashIcon} from "@radix-ui/react-icons";
-import {Slot} from "@radix-ui/react-slot";
+import { Slot } from "@radix-ui/react-slot";
 import * as React from "react";
 
-import {cn} from "@recallnet/ui2/lib/utils";
+import { cn } from "@recallnet/ui2/lib/utils";
 
-function Breadcrumb({...props}: React.ComponentProps<"nav">) {
+function Breadcrumb({ ...props }: React.ComponentProps<"nav">) {
   return <nav aria-label="breadcrumb" data-slot="breadcrumb" {...props} />;
 }
 
-function BreadcrumbList({className, ...props}: React.ComponentProps<"ol">) {
+function BreadcrumbList({ className, ...props }: React.ComponentProps<"ol">) {
   return (
     <ol
       data-slot="breadcrumb-list"
@@ -21,7 +20,7 @@ function BreadcrumbList({className, ...props}: React.ComponentProps<"ol">) {
   );
 }
 
-function BreadcrumbItem({className, ...props}: React.ComponentProps<"li">) {
+function BreadcrumbItem({ className, ...props }: React.ComponentProps<"li">) {
   return (
     <li
       data-slot="breadcrumb-item"
@@ -52,14 +51,14 @@ function BreadcrumbLink({
   );
 }
 
-function BreadcrumbPage({className, ...props}: React.ComponentProps<"span">) {
+function BreadcrumbPage({ className, ...props }: React.ComponentProps<"span">) {
   return (
     <span
       data-slot="breadcrumb-page"
       role="link"
       aria-disabled="true"
       aria-current="page"
-      className={cn("text-gray-600 foreground font-normal", className)}
+      className={cn("foreground font-normal text-gray-600", className)}
       {...props}
     />
   );
@@ -78,7 +77,9 @@ function BreadcrumbSeparator({
       className={cn("[&>svg]:size-3.5", className)}
       {...props}
     >
-      {children ?? <div className="w-[1px] h-5 bg-gray-200 rigin-center rotate-40 mx-2"></div>}
+      {children ?? (
+        <div className="rigin-center rotate-40 mx-2 h-5 w-[1px] bg-gray-200"></div>
+      )}
     </li>
   );
 }


apps/comps/components/agent-profile/agent-credentials.tsx
ramirotw, line 1: @Rcontre360 keep in mind that we already introduced the sandbox and credentials components in https://github.com/recallnet/js-recall/pull/790 (https://github.com/recallnet/js-recall/pull/790/files#diff-1b759c734dc589646ef6e46d26c2723fc262ba446187e14d6d488d42425d8d7b)

apps/comps/lib/api-client.ts
```diff
@@ -301,6 +301,15 @@ export class ApiClient {
     return this.request<AgentApiKeyResponse>(`/user/agents/${agentId}/api-key`);
   }
 
+  /** TODO: once added sandbox api key, change endpoint
```
dtbuchholz: what's the TODO here?

apps/comps/hooks/useApiKey.ts
```diff
@@ -0,0 +1,34 @@
...
+const apiClient = new ApiClient();
+
+/**
+ * Hook to fetch agents with pagination and filtering
+ * @param params Query parameters for agents endpoint
+ * @returns Query result with agents data
+ */
+export const useApiKey = (agentId: string) =>
+  useQuery({
+    queryKey: ["agent", "api-key", agentId],
```
dtbuchholz:
```suggestion
    queryKey: ["agent", "api-key", "live", agentId],
```

apps/comps/hooks/useApiKey.ts
```diff
@@ -0,0 +1,34 @@
...
+  });
+
+/**
+ * Hook to fetch agents with pagination and filtering
+ * @param params Query parameters for agents endpoint
+ * @returns Query result with agents data
+ */
+export const useSandboxApiKey = (agentId: string) =>
+  useQuery({
+    queryKey: ["agent", "api-key", agentId],
```
dtbuchholz:
```suggestion
    queryKey: ["agent", "api-key", "sandbox", agentId],
```

apps/comps/hooks/useApiKey.ts
```diff
@@ -0,0 +1,34 @@
+import { useQuery } from "@tanstack/react-query";
+
+import { ApiClient } from "@/lib/api-client";
+import { AgentApiKeyResponse } from "@/types";
+
+const apiClient = new ApiClient();
+
+/**
+ * Hook to fetch agents with pagination and filtering
```
dtbuchholz: nit: i know this is a legacy comment, but since we should make the fix below for unique keys, we should just update the comment, too

apps/comps/components/clipboard/index.tsx
```diff
@@ -9,10 +13,31 @@ export const Clipboard = ({
   text: string;
   className?: string;
 }) => {
+  const [copied, setCopied] = React.useState(false);
+  const [, copyToClipboard] = useCopyToClipboard();
+
+  const handleCopy = async () => {
+    copyToClipboard(text);
+    setCopied(true);
+    toast.success("Wallet Address copied to clipboard");
```
dtbuchholz, line 22: it doesnt look like this is being applied? i dont see a toast when i copy the agent's allet

apps/comps/components/agent-profile/best-placement.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const AgentBestPlacement = ({
+  rank,
+  places,
+}: {
+  rank?: number;
+  places?: number;
+}) => {
+  const rankInfo = {
+    1: { color: "text-[#FBD362]", name: "1st" },
```
dtbuchholz: we should make these standardized colors in the `ui2` package

apps/comps/components/agent-profile/index.tsx
```diff
@@ -119,28 +124,35 @@ export default function AgentProfile({
...
+              />
+            ) : (
+              <MirrorImage
+                image={agent.imageUrl || "/agent-placeholder.png"}
+                width={170}
+                height={170}
+              />
+            )}
+          </div>
+          {agent.walletAddress && (
```
dtbuchholz, line 148: does this match the latest designs? namely, the agent wallet show up on the card in both public/private agent profiles, it should be copyable, have a border, and be truncated

https://www.figma.com/design/WqQqcsg1g66EdMjjM8yfIM/Competitions-App---v.1.0--06-25-?node-id=6-42386&m=dev
Screenshot 2025-07-08 at 10 42 21 AM


if not, nbd, we can follow up after. i know we've been changing things up a lot.

apps/comps/lib/api-client.ts
Rcontre360: removed. Was used during dev, the right one was useSandboxAgentApiKey from hooks/useSandbox

apps/comps/components/agent-profile/index.tsx
Rcontre360: yes it does, it only shows a wallet if the agent has a wallet btw

image

dtbuchholz: @Rcontre360 ooooh sorry about the confusion here. my local branch somehow wasnt pulling from the remote, so i didnt see some of these latest changs
Screenshot 2025-07-08 at 2 31 32 PM

i'm not sure why, but the spacing with the copy button is a bit off for me. yours has more left padding/margin than what i see

dtbuchholz (approved):
lgtm

PR #807, opened 2025-07-03
Rafael/rewards user profile

Rcontre360:
rebased to 
https://github.com/recallnet/js-recall/pull/805

only added rewards button on user profile comps table


apps/comps/lib/api-client.ts
```diff
@@ -301,6 +301,15 @@ export class ApiClient {
     return this.request<AgentApiKeyResponse>(`/user/agents/${agentId}/api-key`);
   }
 
+  /** TODO: once added sandbox api key, change endpoint
+   * Get agent sandbox api key
+   * @param params - Query parameters
+   * @returns Agents response
+   */
+  async getAgentSandboxApiKey(agentId: string): Promise<AgentApiKeyResponse> {
```
dtbuchholz: this isn't needed, right?

ramiro already implemented [this hook](https://github.com/recallnet/js-recall/blob/d15427a8c637b45a8491de29457245c1bcd33c3e/apps/comps/hooks/useSandbox.ts#L46) and this [sandbox-specific API call](https://github.com/recallnet/js-recall/blob/d15427a8c637b45a8491de29457245c1bcd33c3e/apps/comps/lib/sandbox-client.ts#L68)

Rcontre360: thats, right. Forgot to remove it during conflict resolution

dtbuchholz (approved):

Issue #669, opened 2025-06-19
[FEATURE] Fix competition `ended` status & auto-end


dtbuchholz:
**Problem Statement**
- Every competition has a `status` and a superficial (nullable) `endDate`
- The status is _only_ updated if the the competition is manually ended by the admin; the `endDate` has no affect
- Thus, if a competition `endDate` passes, but the admin doesn't manually end the comp, the `status` will still be `active.
- This is mostly problematic for the UI because we render different views based on `status`.

> It's an edge case because we only run 1 comp at a time, and we're aware of when a comp ~ends

**Proposed Solution**
1. We need "overwrite" the `status` in a competition table query and set it `ended` if the `endDate < current_datetime`. Thus, the `competitions` table sorta acts like a cache. 
2. We need to "auto-end" a competition if the `endDate` hits. This means we need some cron-like service, and we should probably define competition start/end date standards to make this process optimal. E.g., "always end comps at 11:59pm" so that our service can run infrequently—but will most definitely run at this time.
3. We should also revist the admin start/end competition logic, but we also need to make sure we get marketing's input. E.g., should date-less competitions exist?

**Impact**
- From the frontend's perspective, the UX impact of _not_ doing this shouldn't be a huge problem.
- From a comp execution / API perspective, this _does_ mean that an agent can make a trade past the "end date".

**Priority**
Medium/high. Not essential but it's something we need to do.

dtbuchholz:
cc @mzkrasner @joewagner in case this misses anything

joewagner:
The one addition I have is that we should consider using generated columns.  The status of a competition could become a computed/generated value that is the result of checking `is now before, during, or after the competition dates duration?`.

If this makes sense to folks, there are a few other places we might want to consider stored or virtual generated columns.

stbrody:
At minimum, we can reject trades if the time is past the competition end date, and stop the portfolio snapshotter

brunocalza:
Thinking of adding another service to `SchedulerService` to have a cron-like behavior that checks the competition `endDate`? What do you guys think? cc @dtbuchholz @joewagner @mzkrasner

dtbuchholz:
@brunocalza makes sense to me. 

we should also refactor the existing portfolio snapshotting logic to use whatever you come up with. we can handle that in a separate PR, but i'm just calling it out for future compatibility: https://github.com/recallnet/js-recall/issues/797

brunocalza:
[...]

PR #813, opened 2025-07-03
feat: auto end competitions

brunocalza:
## Summary

Closes https://github.com/recallnet/js-recall/issues/669

This PR implements auto-end for competitions. The competition auto ends when end date has passed.

## Implementation

- `SchedulerService` is expanded to support a new scheduler that checks competitions end date. Those with past end date are ended.
- A check to block trades when competition's end date has passed is added
- A check to portifolio snapshotter when competition's end date has passed is added
- `endDate` is added to create/start competition method

## Notes

- I've explored a more robust cron solution with [ `pg-boss`](https://github.com/timgit/pg-boss) but it turned out to be much more complicated. I decided with a simpler (maybe not optimal) approach for now. I didn't want to increase so much complexity right before launch.


apps/api/src/controllers/trade.controller.ts
```diff
@@ -65,6 +65,22 @@ export function makeTradeController(services: ServiceRegistry) {
           `[TradeController] Executing trade with competition ID: ${competitionId}`,
         );
 
+        // Fetch the competition and check if end date has passed
```
brunocalza, line 68: block trades if competitions end date has passed

apps/api/src/services/portfolio-snapshotter.service.ts
```diff
@@ -45,6 +46,25 @@ export class PortfolioSnapshotter {
       `[PortfolioSnapshotter] Taking portfolio snapshot for agent ${agentId} in competition ${competitionId}`,
     );
 
+    // Check if competition exists and if end date has passed
```
brunocalza, line 49: don't take a portifolio snapshot if the competition end date has passed

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -1303,3 +1305,38 @@ export async function getBulkAgentCompetitionRankings(
     return new Map();
   }
 }
+
+/**
+ * Find active competitions that have reached their end date
+ * @returns Array of active competitions that should be ended
+ */
+export async function findActiveCompetitionsPastEndDate() {
```
dtbuchholz: once you merge `main` into this branch, you'll notice there's a new per-repository pattern for metric logging.

we'll have to change this method name to include `Impl` at the end: `findActiveCompetitionsPastEndDateImpl`

and then export a new variable in this file that effectively wraps this method, using the method name you originally defined:
```ts
export const findActiveCompetitionsPastEndDate = createTimedRepositoryFunction(
  findActiveCompetitionsPastEndDateImpl,
  "CompetitionRepository",
  "findActiveCompetitionsPastEndDate",
);
```
and that should be all you need to do.

brunocalza:
@dtbuchholz made those changes, i messed up when rebasing and became hard to add the changes in another commit to make it easier to review, i hope that's okk

PR #821, opened 2025-07-07
fix: optimize portfolio snapshot type for trades and req fallback

mzkrasner:
# Summary

Portfolio snapshots were being triggered for all agents in a competition in several scenarios where only a single agent's portfolio had actually changed. This created unnecessary database load and API calls that scaled poorly with competition size.

## Changes Made
1. Trade Execution Optimization

- Before: Every trade → snapshot all agents in competition
- After: Every trade → snapshot only the trading agent
- File: apps/api/src/services/trade-simulator.service.ts
- Change: takePortfolioSnapshots() → takePortfolioSnapshotForAgent()

2. Portfolio Request Fallback Optimization

- Before: Agent requests portfolio with no snapshots → snapshot all agents
- After: Agent requests portfolio with no snapshots → snapshot only requesting agent
- File: apps/api/src/controllers/agent.controller.ts
- Change: takePortfolioSnapshots() → takePortfolioSnapshotForAgent()

### DIFF
--- a/apps/api/src/controllers/agent.controller.ts
+++ b/apps/api/src/controllers/agent.controller.ts
@@ -379,9 +379,9 @@ export function makeAgentController(services: ServiceRegistry) {
           console.log(
             `[AgentController] No portfolio snapshots found for agent ${agentId} in competition ${activeCompetition.id}`,
           );
-          // Request a snapshot asynchronously (don't await)
+          // Request a snapshot for this agent asynchronously (don't await)
           services.portfolioSnapshotter
-            .takePortfolioSnapshots(activeCompetition.id)
+            .takePortfolioSnapshotForAgent(activeCompetition.id, agentId)
             .catch((error) => {
               console.error(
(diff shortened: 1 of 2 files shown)


apps/api/src/controllers/agent.controller.ts
```diff
@@ -379,9 +379,9 @@ export function makeAgentController(services: ServiceRegistry) {
           console.log(
             `[AgentController] No portfolio snapshots found for agent ${agentId} in competition ${activeCompetition.id}`,
           );
-          // Request a snapshot asynchronously (don't await)
+          // Request a snapshot for this agent asynchronously (don't await)
           services.portfolioSnapshotter
-            .takePortfolioSnapshots(activeCompetition.id)
+            .takePortfolioSnapshotForAgent(activeCompetition.id, agentId)
```
dtbuchholz, line 384: maybe remove this line entirely?

edit: we can handle later; this fix is obvious

Issue #823, opened 2025-07-08
[BUG] App crashes on agent image input


Rcontre360:
**Bug Description**
When creating an agent and putting as input anything but a URL the app crashes

PR #825, opened 2025-07-08
fix: fixed url issue on create agent

Rcontre360:
#823

### DIFF
--- a/apps/comps/components/create-agent/image-preview.tsx
+++ b/apps/comps/components/create-agent/image-preview.tsx
@@ -21,6 +21,16 @@ export function ImagePreview({ imageUrl }: ImagePreviewProps) {
     setImageError(false);
   };
 
+  const isValidUrl = (str: string) => {
+    try {
+      new URL(str);
+      return true;
+      //eslint-disable-next-line
+    } catch (e) {
+      return false;
+    }
+  };
+
   useEffect(() => {
     setImageError(false);
   }, [imageUrl]);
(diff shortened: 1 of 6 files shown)


apps/comps/components/create-agent/image-preview.tsx
```diff
@@ -21,6 +21,16 @@ export function ImagePreview({ imageUrl }: ImagePreviewProps) {
     setImageError(false);
   };
 
+  const isValidUrl = (str: string) => {
+    try {
+      new URL(str);
+      return true;
+      //eslint-disable-next-line
```
dtbuchholz: we shouldnt disable eslint unless it's a "last resort". we dont actually use the `catch (e)`, so we can just catch and return false. otherwise, this lgtm. 

however, i also noticed a similar error happens on the "edit agent" page for the image profile. we should fix this here, too.

steps to reproduce:
- when i create an agent, i use this image URL and can successfully set it as my agent profile pic: https://fastly.picsum.photos/id/1/200/200.jpg?hmac=jZB9EZ0Vtzq-BZSmo7JKBBKJLW46nntxq79VMkCiBG8
- if i try to edit the agent profile image and paste the same URL, it fails/errors
Screenshot 2025-07-08 at 11 07 50 AM
Screenshot 2025-07-08 at 11 07 59 AM
Screenshot 2025-07-08 at 11 09 45 AM

additionally, the user profile doesnt seem to validate things in the same way. perhaps we include the fix here, too. e.g., use some random URL and edit the user profile pic, and it's "valid" but shouldnt be
Screenshot 2025-07-08 at 11 11 43 AM

dtbuchholz (approved):
lgtm!

Issue #781, opened 2025-06-30
[IMPL] Display rewards that competition is offering


stbrody:
DoD:
- Competition page shows the rewards associated with this competition
- Implementing designs from https://github.com/recallnet/js-recall/issues/705
- Database schema changes to store rewards allocated to the competition.  Rewards data has 4 values: total, 1st place, 2nd place, 3rd place (maybe total can be implied by sum of the other 3?)
- Update competition admin APIs to enable setting the rewards.  Should be nullable
- Update competition info API to return the reward fields.
- Update UI for competition page to display reward data

dtbuchholz:
1. for the table design, we want to create a new `rewards` table in the "core" schema with the following columns:
  - `id`: UUID, PK, not null
  - `competition_id`: FK to `competitions` table, not null
  - `rank`: integer, not null
  - `reward`: integer, not null
  - `agent_id`: FK to `agents` ID; is nullable
2. when an admin creates _or_ updates a competition (via `/admin` APIs), we need to write to this new table.
3. when an end ends a competition, we already calculate the "final" data in `competitions_leaderboard`. we should be able to reuse that data to look up the `rewards` table's `rank` (1st, 2nd, etc), match it with the competition ranks, and write the agent ID to the `rewards` table.

PR #829, opened 2025-07-08
feat: implement rewards in api & comps app

ramirotw:
Fix https://github.com/recallnet/js-recall/issues/781


apps/api/openapi/API.md
```diff
@@ -1250,7 +1250,7 @@ Get agents participating in a competition
 
 ##### Description:
 
-Get a list of all agents participating in a specific competition with their scores and positions
+Get a list of all agents participating in a specific competition with their scores and ranks
```
ramirotw, line 1253: unrelated to this PR, was caused by the `generate` npm script

apps/api/src/controllers/admin.controller.ts
```diff
@@ -919,10 +936,31 @@ export function makeAdminController(services: ServiceRegistry) {
             updates,
           );
 
+        let updatedRewards: SelectReward[] = [];
+        if (rewards) {
+          if (updatedCompetition.status === COMPETITION_STATUS.ENDED) {
+            throw new ApiError(
+              400,
+              "Cannot update rewards for an ended competition",
```
ramirotw, line 944: To simplify the request and the logic for modifying the rewards, they are recreated. If the competition has ended, to avoid mistakenly deleting the winning agents, we simply prevent any modification of the rewards

apps/api/src/controllers/competition.controller.ts
```diff
@@ -495,6 +495,10 @@ export function makeCompetitionController(services: ServiceRegistry) {
             pagingParams,
           );
 
+        // Populate the rewards for each competition
+        // const competitionsWithRewards =
+        //   await services.coreRewardService.populateRewards(competitions);
+
```
ramirotw: need to remove this. It was a temp solution to add the rewards to the competitions response, but I replaced this in favor of a join to the rewards table.

ramirotw (commented):
if we are ok with the current implementation, the only thing missing would be the unit tests

Issue #778, opened 2025-06-30
[IMPL] Email verification success/failure flows


stbrody:
implements designs from #777

dtbuchholz:
see comment in https://github.com/recallnet/js-recall/issues/772#issuecomment-3033359585 about altering the backend routes. instead of modals + `/profile` page, we'll point to a dedicated page per the designs.

Rcontre360:
almost finished, waiting for #818

dtbuchholz:
@Rcontre360 this should be good to go—the feature was implemented in https://github.com/recallnet/js-recall/pull/819

PR #833, opened 2025-07-10
feat(comps): email flow

Rcontre360:
closes https://github.com/recallnet/js-recall/issues/778


apps/comps/components/user-info/index.tsx
```diff
@@ -39,6 +42,9 @@ export default function UserInfoSection({
   onSave,
 }: UserInfoSectionProps) {
   const [editField, setEditField] = useState<"email" | "website" | null>(null);
+  const [emailVerifyClicked, setEmailVerifyClicked] = useState(false);
+  const { mutate: verifyEmail } = useVerifyEmail();
+  const verified = false;
```
dtbuchholz: note: we're going to add a new field to the user profile response called `isEmailVerified` per https://github.com/recallnet/js-recall/issues/850

i'll make a note to handle that this separately as a future follow up

dtbuchholz:
other than those small tweaks, lookin solid

apps/comps/components/user-info/index.tsx
dtbuchholz: @Rcontre360 ah, actually....you dont need to wait for that issue to land. the API response includes the data we need already.

so, instead of hardcoding `verified = false`, we can add the `isEmailVerified` field to the `User` type:
```ts
export interface User {
  id: string;
  walletAddress: string;
  status: "active" | "inactive" | "suspended" | "deleted";
  name?: string;
  email?: string;
  isEmailVerified: boolean;
  imageUrl?: string;
  metadata?: {
    website?: string;
  };
  createdAt: string;
  updatedAt: string;
}
```
then, just access the `user.isEmailVerified` to know if the user's email has been verified or not

Rcontre360:
waiting for #855

Issue #837, opened 2025-07-10
[FEATURE] Batch portfolio token values entries


brunocalza:
**Problem Statement**

In supabase I see lots of time being consumed executing the following query
```
insert into
  "trading_comps"."portfolio_token_values" (
    "id",
    "portfolio_snapshot_id",
    "token_address",
    "amount",
    "value_usd",
    "price",
    "specific_chain",
    "symbol"
  )
values
  (default, $1, $2, $3, $4, $5, $6, $7)
returning
  "id",
  "portfolio_snapshot_id",
  "token_address",
  "amount",
  "value_usd",
  "price",
  "specific_chain",
  "symbol"
```
This comes from https://github.com/recallnet/js-recall/blob/main/apps/api/src/services/portfolio-snapshotter.service.ts#L160-L170

**Proposed Solution**
Improve the performance of portfolio snapshotter by batching portfolio token values entries

PR #840, opened 2025-07-10
refactor: batch portfolio token values insert

brunocalza:
Closes https://github.com/recallnet/js-recall/issues/837


mzkrasner (approved):
Looks good! So long as any single agent never holds upwards of about 10k tokens, this will work. We could introduce a fallback strategy but that would be a pretty extreme edge case

PR #846, opened 2025-07-11
fix: correct price logging

mzkrasner:
# Summary

- Multi chain provider previously logged `[object Object]` - for example, `[MultiChainProvider] Successfully found price for Solana token so111...: $[object Object]`
- This PR correctly uses the `price` field on the `price` object when logging from this provider


joewagner (approved):
excellent!

PR #862, opened 2025-07-14
fix(comps): allow empty agent image url

dtbuchholz:
there was a regression that required you to add an agent image url during the agent creation process. else, the "submit" button was disabled. this fixes the issue and either (1) allow empty form values, or (2) uses the image validation logic to check if the image url is valid.


dtbuchholz:
> Looks like ci issues are not related to this PR...

yeahh.....so odd. it seems to be some db error that we saw last week, but it's flakey

PR #869, opened 2025-07-15
feat(api): total roi for agent calc

joewagner:


joewagner:
bugbot run

apps/api/drizzle/0021_jazzy_junta.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "trading_comps"."trading_competitions_leaderboard" ADD COLUMN "starting_value" numeric(30, 15) DEFAULT 0 NOT NULL;
```
joewagner: The `trading_competitions_leaderboard` table is updated to hold a starting value stat.  This allows us to get a percent when combined with the pnl.

apps/api/e2e/tests/total-roi-calculation.test.ts
```diff
@@ -0,0 +1,352 @@
...
+      agentIds: [agent.id],
+      tradingType: CROSS_CHAIN_TRADING_TYPE.DISALLOW_ALL,
+    })) as StartCompetitionResponse;
+    expect(startResponse.success).toBe(true);
+
+    const competitionId = startResponse.competition.id;
+    const startingValue = await getStartingValue(agent.id, competitionId);
+
+    const tradeAmount = 1000;
+    // Burn 1000 USDC (we have a known roi this way)
```
joewagner, line 72: These tests all rely on being able to burn and get a negative roi.  You'll see a lot of places where the roi is checked against `-1 * tradeAmount / startingValue`

apps/api/e2e/utils/test-helpers.ts
```diff
@@ -484,3 +487,25 @@ export async function generateTestCompetitions(adminApiKey: string) {
     user3,
   };
 }
+
+export async function getStartingValue(agentId: string, competitionId: string) {
```
joewagner, line 495: Test helper to allow calculating the expected roi

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -1174,19 +1177,23 @@ async function batchInsertLeaderboardImpl(
...
+    let results: (InsertCompetitionsLeaderboard & {
+      pnl?: number;
+      startingValue?: number;
+    })[] = await db
       .insert(competitionsLeaderboard)
       .values(valuesToInsert)
       .returning();
 
-    const pnlsToInsert = valuesToInsert.filter((e) => e.pnl);
+    const pnlsToInsert = valuesToInsert.filter((e) => e.pnl !== undefined);
```
joewagner, line 1288: this came out of formatting + linting.

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -1174,19 +1177,23 @@ async function batchInsertLeaderboardImpl(
...
     if (pnlsToInsert.length) {
       const pnlResults = await db
         .insert(tradingCompetitionsLeaderboard)
         .values(
           pnlsToInsert.map((entry) => {
             return {
               pnl: entry.pnl,
+              startingValue: entry.startingValue || 0,
```
joewagner, line 1296: Insert the starting value so we can calc percent ROI later on.

apps/api/src/database/repositories/leaderboard-repository.ts
```diff
@@ -332,6 +425,29 @@ async function getBulkAgentMetricsImpl(agentIds: string[]): Promise<
...
+          : null;
+
+        if (
+          typeof totalPnl !== "number" ||
+          typeof totalStartingValue !== "number" ||
+          isNaN(totalPnl) ||
+          isNaN(totalStartingValue) ||
+          totalStartingValue <= 0
+        ) {
+          return [row.agentId, null];
```
joewagner, line 442: If there's no entry in the database we send null

apps/comps/components/user-agents/agent-card.tsx
```diff
@@ -64,7 +64,10 @@ export const AgentCard: React.FunctionComponent<AgentCardProps & CardProps> = ({
       </span>
       <div className="flex justify-center gap-3">
         <div className="text-secondary-foreground text-nowrap rounded border p-2">
-          ROI N/A
+          ROI{" "}
+          {agent.stats?.totalRoi
+            ? `${Math.round(agent.stats?.totalRoi * 100)}%`
```
joewagner, line 71: One small change to the frontend to show the stat.

joewagner (commented):
This is ready for review.

apps/api/src/services/agent-manager.service.ts
```diff
@@ -1412,6 +1417,7 @@ export class AgentManager {
           bestPnl: metrics?.bestPnl ?? undefined,
           rank: metrics?.globalRank ?? undefined,
           score: metrics?.globalScore ?? undefined,
+          totalRoi: metrics?.totalRoi ?? null,
```
dtbuchholz, line 1420: we use `?? undefined` in the others...should we do the same here?

dtbuchholz:
lookin good. maybe @smrz2001 could give it a pass over too?

also, let me double check with @mzkrasner and make sure that if we merge this before launch, it'll be fine w/ his data normalization. (i dont see why it wouldnt)

dtbuchholz:
@joewagner does this change also require we run a backfilling script to populate the new column?

dtbuchholz:
Yeh I think so...just so we can have everything ready in case we want to squeeze it in before launch

apps/comps/types/agent.ts
```diff
@@ -33,6 +33,7 @@ export interface Agent {
     bestPnl: number;
     totalVotes: number;
     totalTrades: number;
+    totalRoi?: number | null;
```
smrz2001: Do we want all three possibilities?
- `totalRoi` is absent.
- `totalRoi` is present, but a number (e.g. `0`).
- `totalRoi` is present, but `null`.

Or do we just want the first two?

dtbuchholz: hm yeah, to keep it consistent, it should use `?` and not explicitly null.

related to [this comment](https://github.com/recallnet/js-recall/pull/869/files/b7be4e7007aa6688fbd84248e003046ed541d206#r2211012397) above

joewagner: Since these are types that come from http responses, the checks here don't have any kind of runtime guarantee.  I left it as potentially being `number | undefined | null` to handle all of the potential types given pre and post migration considerations.  
Specifically:
- pre migration it is undefined because it doesn't exist
- post migration agents who haven't finished a comp would have null, since the value is intentionally absent.
- post migration agents who have finished a comp would have a positive or negative number.

Given we are waiting for the migration to be figured out to merge this, it probably makes sense to make it `number | null`, since those are the possible states returned by the api.  Thoughts?

dtbuchholz (approved):
@joewagner note: i _think_ you might need my script to land before yours wrt the prod db https://github.com/recallnet/js-recall/pull/933. i.e., my script will populate the `competitions_leaderboard` and `trading_competitions_leaderboard` tables (they're currently empty in prod)...maybe it'll make sense to combine ours into one? idk

(if you want, we can do a fast follow on the script itself since that's a non-migration piece and can happen "whenever")

PR #873, opened 2025-07-15
Rafael/qa fixes

### LINKED ISSUE #872: [BUG] List of QA points to fix (Comps pages)
- [x] Footer links dont appear clickable: add hover effect
- [x] incomplete info under "your agents" when logged out
- [x] Participants should send you to comps page, not agent profile
- [x] Comps page main card should ve fully visible on height
- [x] Comps page main card hides Join button when there are too many agents
- [x] Make date component to standardize it

Rcontre360:
closes #872


dtbuchholz:
@Rcontre360 lookin good. couple minor changes.

 `Comps page main card should be fully visible on height` => i can see the "vote" button is cut off at the bottom. i know we've had some back and forth on the card being within the full landing viewport. if that's not possible, let's just go ahead and make sure the buttons don't get cut off...that's the lesser of two evils
Screenshot 2025-07-18 at 3 09 54 PM

apps/comps/utils/format.ts
```diff
@@ -40,3 +42,21 @@ export const formatCompactNumber = (value: number): string => {
...
+    parsedDate = parseISO(date);
+  } else {
+    return "Invalid Date";
+  }
+
+  if (!isValid(parsedDate)) {
+    return "Invalid Date";
+  }
+
+  return format(parsedDate, "MM/dd/yyyy");
```
dtbuchholz: let's change this to `MMMM do` instead of `MM/dd/yyyy`. (the notion docs stated we want to change this to `MMMM do, yyyy`, but i think that'll mess up the spacing)

also to keep things consistent, we should make sure other instances the `format` function via the `import { format } from "date-fns";` use our new `formatDate` function instead. it looks like it's used in:
- components/competition-info.tsx
- components/competitions-table.tsx

dtbuchholz (approved):

PR #876, opened 2025-07-15
feat(api): use env var for email auto-verify

dtbuchholz:
Since the UI will gate access to API keys, a user cannot see their API key until they've verified their email + joined an active sandbox competition. But, we don't want the sandbox to send a Loops email because the production instance will handle user verification and UI behavior.

Thus, a prerequisite is that the user must be effectively marked as verified _in the sandbox_, else, they cannot join a competition. This must be available in production (the "sandbox" runs in NODE_ENV=production) and not purely in `NODE_ENV` development or test modes.

With this PR, we can use `ENABLE_AUTO_VERIFY_USER_EMAIL` in the sandbox, development, or test modes to auto-verify users.

### DIFF
--- a/apps/api/src/config/index.ts
+++ b/apps/api/src/config/index.ts
@@ -155,6 +155,7 @@ export const config = {
     sandboxUrl: "https://api.sandbox.competitions.recall.network",
   },
   email: {
+    autoVerifyUserEmail: process.env.ENABLE_AUTO_VERIFY_USER_EMAIL === "true",
     apiKey: process.env.LOOPS_API_KEY || "",
     transactionalId: process.env.LOOPS_TRANSACTIONAL_ID || "",
   },
--- a/.github/workflows/api-ci.yml
+++ b/.github/workflows/api-ci.yml
@@ -289,6 +289,9 @@ jobs:
           # Optional: API prefix
           API_PREFIX=testing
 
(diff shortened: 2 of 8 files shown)


apps/api/src/config/index.ts
```diff
@@ -155,6 +155,7 @@ export const config = {
     sandboxUrl: "https://api.sandbox.competitions.recall.network",
   },
   email: {
+    autoVerifyUserEmail: process.env.ENABLE_AUTO_VERIFY_USER_EMAIL === "true",
```
joewagner, line 158: Only thing I'd suggest is that we might want to add a check to make sure we never start the production server with this as true.   Thoughts?

joewagner: e.g.  on line 147 do something like:
```
if (process.env.NODE_ENV === "production" && process.env.ENABLE_AUTO_VERIFY_USER_EMAIL === "true") {
  throw new Error("cannot start in production mode with auto email verification");
}
```

dtbuchholz: @joewagner the sandbox server runs with `NODE_ENV=production`, though, and that's where we need this feature too

joewagner: Ah, that makes sense.

Issue #866, opened 2025-07-14
[FEATURE] Add product analytics tracking


brunocalza:
**Problem Statement**
https://www.notion.so/Analytics-Requirements-for-Recall-App-223dfc9427de80b6b9a7c3cb4c14be0c

**Proposed Solution**
We opted for adding Posthog and start tracking the following events 
- LandingPageViewed
- UserLoggedIn
- UserClickedAddAgentButton
- UserSuccessfullyCreatedAgent
- UserVerifiedEmail
- UserUpdatedProfile

For more context see:
- [Analytics Implementation Proposal](https://www.notion.so/Analytics-Implementation-Proposal-229dfc9427de80a990ccc6e741d670b3) 
- [Analytics Implementation Proposal II](https://www.notion.so/Analytics-Implementation-Proposal-II-231dfc9427de807498b7f125ab1f4ff3?d=231dfc9427de80c0a7a8001c991a57c7#231dfc9427de80a58f5bddd77731b1f9)
- https://recalllabs.slack.com/archives/C08TEDYB5E1/p1752699342063019

PR #897, opened 2025-07-17
feat: adds posthog event tracking

brunocalza:
## Summary

Closes https://github.com/recallnet/js-recall/issues/866

In this PR we add integration to `Posthog` and start tracking the following events

- LandingPageViewed
- UserLoggedIn
- UserClickedAddAgentButton
- UserSuccessfullyCreatedAgent
- UserVerifiedEmail
- UserUpdatedProfile

For more context, see:

- [Analytics Implementation Proposal](https://www.notion.so/Analytics-Implementation-Proposal-229dfc9427de80a990ccc6e741d670b3)
- [Analytics Implementation Proposal II](https://www.notion.so/Analytics-Implementation-Proposal-II-231dfc9427de807498b7f125ab1f4ff3?d=231dfc9427de80c0a7a8001c991a57c7#231dfc9427de80a58f5bddd77731b1f9)
- https://recalllabs.slack.com/archives/C08TEDYB5E1/p1752699342063019

Note: this was heavily ai driven, and I don't really have the knowledge to evaluate this frontend code. I have tested it and it seems fine. But lmk what is wrong so I can learn something

## Next steps

- Configure staging. I guess we can use my account for now (1 Pass - Posthog Staging Account). If that starts to be a problem we can switch to another one.
- Set up a prod account
- Set up a reverse proxy to avoid tracking blockers blocking us (https://posthog.com/docs/advanced/proxy)


apps/comps/.env.example
```diff
@@ -14,3 +14,7 @@ NEXT_PUBLIC_HOTJAR_SV=
 
 # Frontend 
 NEXT_PUBLIC_FRONTEND_URL=
+
+# PostHog Analytics
+NEXT_PUBLIC_POSTHOG_KEY=
```
dtbuchholz, line 19: we'll want to add these to the `apps/comps/turbo.json` file

apps/comps/components/user-agents/index.tsx
```diff
@@ -14,10 +14,12 @@ import { cn } from "@recallnet/ui2/lib/utils";
 
 import { AgentCard } from "@/components/user-agents/agent-card";
 import AgentsSummary from "@/components/user-agents/agents-summary";
+import { useAnalytics } from "@/hooks/usePostHog";
 import { Agent } from "@/types";
 
 export default function UserAgentsSection({ agents }: { agents: Agent[] }) {
-  let agentList = <NoAgents />;
+  const { trackEvent } = useAnalytics();
+  let agentList: React.ReactNode;
```
dtbuchholz: this isnt really the "reactjs" what to do things.

we can create a fn and just handle the "no agents" scenario like this:
```ts
export default function UserAgentsSection({ agents }: { agents: Agent[] }) {
  const { trackEvent } = useAnalytics();
  const nAgents = agents.length;

  const bestPlacement = useMemo(
    () =>
      agents.reduce(
        (acc, agent) => {
          if (agent.stats?.bestPlacement?.rank) {
            return agent.stats.bestPlacement.rank < acc.rank
              ? agent.stats.bestPlacement
              : acc;
          }
          return acc;
        },
        {
          competitionId: "",
          rank: Infinity,
          score: 0,
          totalAgents: 0,
        },
      ),
    [agents],
  );

  const highest = useMemo(
    () =>
      agents.reduce((acc, agent) => {
        if (agent.stats?.bestPnl) {
          return agent.stats.bestPnl > acc ? agent.stats.bestPnl : acc;
        }
        return acc;
      }, Number.MIN_SAFE_INTEGER),
    [agents],
  );

  const completedComps = useMemo(() => {
    return agents.reduce((acc, agent) => {
      return acc + (agent.stats?.completedCompetitions ?? 0);
    }, 0);
  }, [agents]);

  const renderAgentList = () => {
    if (nAgents === 0) {
      return <NoAgents trackEvent={trackEvent} />;
    }

    if (nAgents > 0 && nAgents <= 3) {
      return (
        <div
          className={cn(`flex w-full flex-col justify-around gap-4`, {
            "xs:flex-row": nAgents == 1,
            "sm:flex-row": nAgents == 2,
            "lg:flex-row": nAgents == 3,
          })}
        >
          <div
            className={cn("flex gap-6", {
              "xs:overflow-x-visible overflow-x-auto": nAgents < 3,
              "overflow-x-auto md:overflow-x-visible": nAgents == 3,
            })}
          >
            {agents.map((agent, i) => (
              <AgentCard
                key={i}
                agent={agent}
                className="h-87 min-w-64 max-w-80 flex-1"
              />
            ))}
          </div>
          <AgentsSummary
            nAgents={nAgents}
            bestPlacement={bestPlacement}
            completedComps={completedComps}
            highest={highest}
          />
        </div>
      );
    }

    // nAgents >= 4
    return (
      <div className="flex w-full flex-col gap-10">
        <div className="flex justify-around gap-10 overflow-x-auto">
          {agents.map((agent, i) => (
            <AgentCard
              key={i}
              agent={agent}
              className="h-87 min-w-64 max-w-80 flex-1"
            />
          ))}
        </div>
        <AgentsSummary
          nAgents={nAgents}
          bestPlacement={bestPlacement}
          completedComps={completedComps}
          highest={highest}
        />
      </div>
    );
  };

  return (
    <Collapsible defaultOpen className="my-7">
      <CollapsibleTrigger>
        <div className="flex w-full items-center justify-between">
          <div className="ml-2 flex items-center gap-2">
            <span className="text-2xl font-bold">Your Agents</span>
            <span className="text-secondary-foreground">({nAgents})</span>
          </div>
          <Button asChild>
            <Link
              href="/create-agent"
              onClick={() => trackEvent("UserClickedAddAgentButton")}
            >
              {"+ ADD AGENT"}
            </Link>
          </Button>
        </div>
      </CollapsibleTrigger>
      <CollapsibleContent className="w-full">
        {renderAgentList()}
      </CollapsibleContent>
    </Collapsible>
  );
}
```

apps/comps/hooks/useAuth.ts
```diff
@@ -36,15 +38,23 @@ export const useNonce = () => {
...
+  const { trackEvent } = useAnalytics();
+  const posthog = usePostHog();
 
   return useMutation({
     mutationFn: async (data: LoginRequest) => {
       const result = await apiClient.login(data);
       return result;
     },
-    onSuccess: () => {
+    onSuccess: (response, _) => {
```
dtbuchholz: the `, _` is unnecessary.

note: i'm not sure why our `pnpm lint` command doesn't pick this up. i'll look into that separately.

apps/comps/hooks/usePostHog.ts
```diff
@@ -0,0 +1,27 @@
+"use client";
```
dtbuchholz, line 1: if you look at the console logs on the landing page / competitions hub, you'll see a bunch of `LandingPageViewed` events are fired due how the outer `useEffect` hook calls this. `useEffect` makes the UI re-render if any of its dependencies change. if you dont memorize `trackEvent`, then it'll get called if the other `useEffect` dependencies change (e.g., fetch new data).

try this approach with `useCallback`, which will memoize the behavior in the hook:
```ts
"use client";

import { usePostHog } from "posthog-js/react";
import { useCallback } from "react";

/**
 * Custom hook for PostHog event tracking
 * Provides a clean interface for capturing events with proper typing
 */
export function useAnalytics() {
  const posthog = usePostHog();

  const trackEvent = useCallback(
    (eventName: string, properties?: Record<string, unknown>) => {
      if (!posthog) {
        console.warn("PostHog not initialized, event not tracked:", eventName);
        return;
      }

      try {
        posthog.capture(eventName, properties);
        console.log("Event tracked:", eventName, properties);
      } catch (error) {
        console.error("Failed to track event:", eventName, error);
      }
    },
    [posthog],
  );

  return { trackEvent };
}
```

apps/comps/hooks/useProfile.ts
```diff
@@ -41,12 +42,17 @@ export const useProfile = () => {
...
+  const { trackEvent } = useAnalytics();
 
   return useMutation({
     mutationFn: async (data: UpdateProfileRequest) => {
       return apiClient.updateProfile(data);
     },
-    onSuccess: () => {
+    onSuccess: (_, variables) => {
+      trackEvent("UserUpdatedProfile", {
+        updated_fields: Object.keys(variables),
```
dtbuchholz: technically, this will include whatever fields are defined in the original type, even if they're undefined.

we could do something like:
```ts
const updatedFields = Object.keys(variables).filter(
  (key) => variables[key as keyof UpdateProfileRequest] !== undefined,
);
```

dtbuchholz:
> But tbh, not really sure how much of that document was really thought through vs ai generated.

yeah, that original reqs doc was suuuuper loose and not fully flushed out.

> I guess it's something to get us started.

fair point...if we want em, we can add em later

dtbuchholz:
ah nvm, should be fine...that's what their docs state https://posthog.com/docs/libraries/next-js#client-side-setup

PR #907, opened 2025-07-18
Update step component

### LINKED ISSUE #865: Agent created even if user cancels agent creation
Reported by @juampibermani:

> Agent is created even if user backs out of agent creation.  User tries to add agent → presses next to step 2 → presses back to step 1 —> presses back to “cancel” → receives warning that info will be discarded → user chooses to leave → agent is created anyways

juampibermani:
Closes #865 

Added button type so form doesn't get submited when button is clicked

### DIFF
--- a/apps/comps/components/create-agent/basics-step.tsx
+++ b/apps/comps/components/create-agent/basics-step.tsx
@@ -154,7 +154,12 @@ export function BasicsStep({ form, onNext, onBack }: BasicsStepProps) {
         )}
       />
       <div className="flex justify-end gap-2">
-        <Button type="button" variant="outline" className="px-10" onClick={onBack}>
+        <Button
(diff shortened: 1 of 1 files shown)


apps/comps/components/create-agent/basics-step.tsx
```diff
@@ -154,7 +154,7 @@ export function BasicsStep({ form, onNext, onBack }: BasicsStepProps) {
         )}
       />
       <div className="flex justify-end gap-2">
-        <Button variant="outline" className="px-10" onClick={onBack}>
+        <Button type="button" variant="outline" className="px-10" onClick={onBack}>
```
dtbuchholz: lgtm. run `pnpm format`, and then we can merge this

PR #927, opened 2025-07-21
fix(api): omit agent email address in res

dtbuchholz:
closes https://github.com/recallnet/js-recall/issues/926


apps/api/e2e/tests/agent.test.ts
```diff
@@ -953,7 +956,52 @@ describe("Agent API", () => {
...
+      adminApiKey,
+      userName,
+      userEmail,
+      agentName,
+      agentDescription,
+    });
+    const agentId = res.agent.id;
+
+    // Manually update database entry to include email
+    const db = await dbManager.connect();
```
dtbuchholz, line 984: note: iirc we want to eventually remove `agents.email` (and hide it in the current UI), so i figured i'd just directly alter the db entry here vs updating all of our test helper methods

mzkrasner:
@dtbuchholz i'm going to approve it, but just a note that we should probably update the openapi notation given this change:

1. Remove email field from /api/admin/agents GET response schema (around line 970)
2. Remove email field from /api/admin/agents/{agentId} GET response schema (around line 1380)
3. Keep email field in /api/admin/agents POST response schema (it's actually returned)
4. Keep email field in /api/admin/search GET response schema (it's actually returned)

PR #975, opened 2025-07-24
fix: reduce inner loop to limit 1

andrewxhill:
Looking at the timing breakdown:
GroupAggregate: ~29.4ms (82% of total time!)
Nested Loop scanning: ~25.3ms
Final lookup: 0.007ms (negligible)
The problem: We're processing 143,700 rows to find MAX timestamp for just 10 agents. 99.1% IMPROVEMENT FOUND!
LATERAL JOIN Results:
Execution Time: 0.176 ms (vs ~36ms before)
Improvement: 99.1% faster
Buffers: 42 (vs 1,100 before)

### DIFF
--- a/apps/api/src/database/repositories/competition-repository.ts
+++ b/apps/api/src/database/repositories/competition-repository.ts
@@ -10,7 +10,6 @@ import {
   isNotNull,
   lte,
   max,
-  min,
   sql,
 } from "drizzle-orm";
 import { unionAll } from "drizzle-orm/pg-core";
--- a/apps/api/src/database/repositories/competition-repository.ts
+++ b/apps/api/src/database/repositories/competition-repository.ts
@@ -844,9 +844,7 @@ async function get24hSnapshotsImpl(competitionId: string, agentIds: string[]) {
     );
 
     return {
-      earliestSnapshots: earliestSnapshots.map(
-        (row) => row.portfolio_snapshots,
-      ),
+      earliestSnapshots: earliestSnapshots,  // ✅ Direct mapping - already flat!
       snapshots24hAgo: snapshots24hAgo.map((row) => ({
         id: row.id,
         agentId: row.agentId,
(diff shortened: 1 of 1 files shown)


apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -799,33 +798,22 @@ async function get24hSnapshotsImpl(competitionId: string, agentIds: string[]) {
...
-      .from(portfolioSnapshots)
-      .innerJoin(
-        earliestSubquery,
-        and(
-          eq(portfolioSnapshots.agentId, earliestSubquery.agentId),
-          eq(portfolioSnapshots.timestamp, earliestSubquery.minTimestamp),
-        ),
-      )
-      .where(eq(portfolioSnapshots.competitionId, competitionId));
+    // Get earliest snapshots for each agent (optimized)
```
stbrody:
```suggestion
    // Get earliest snapshots for each agent (for PnL calculation)
```

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -799,33 +798,22 @@ async function get24hSnapshotsImpl(competitionId: string, agentIds: string[]) {
...
+    const earliestSnapshots = await Promise.all(
```
stbrody: this is sending N (N = length of agentId map) separate SQL queries in parallel to the DB.  This should be condensed back into a single query

andrewxhill:
all we were trying to do here was:

-- Instead of aggregating 143K rows, get latest directly:
FROM competition_agents ca 
CROSS JOIN LATERAL (
    SELECT * FROM portfolio_snapshots ps 
    WHERE ps.agent_id = ca.agent_id 
    ORDER BY timestamp DESC LIMIT 1  -- Gets exactly 1 latest row!
) ps

but the lateral join + limit 1 part doesn't seem to translate to drizzle

carsonfarmer:
What we _want_ is to translate the final sql statement with the limit into what drizzle needs to craft that sql statement.

stbrody:
Rebase of https://github.com/recallnet/js-recall/pull/979

dtbuchholz:
superseded by https://github.com/recallnet/js-recall/pull/997

PR #977, opened 2025-07-24
fix: Serialize snapshot query

stbrody:


stbrody:
This is a fallback in case there's issues with https://github.com/recallnet/js-recall/pull/979

PR #984, opened 2025-07-24
refactor: remove unneeded getPortfolioTokenValues

brunocalza:


apps/api/e2e/tests/portfolio-snapshots.test.ts
```diff
@@ -47,54 +40,6 @@ describe("Portfolio Snapshots", () => {
     console.log(`Admin API key created: ${adminApiKey.substring(0, 8)}...`);
   });
 
-  // Test that a snapshot is taken when a competition starts
```
brunocalza, line 50: remove tests that use that method, but they are testing an admin api we don´t use

Issue #985, opened 2025-07-24
Add middleware to block trade-related endpoints when comp is not active


dtbuchholz:
## Summary

The `/price` and `/trade` endpoints are "unprotected" in the sense that: if you have an API key, you can call them "whenever". We also don't have robust caching set up. This leads to problems where our APIs are constantly getting hit, DexScreener 429s, etc.—which is potentially causing other cascading issues because 

## Proposed solution

Add middleware to gate these the `/price` and `/trade` routes:
- It should throw an error if there is no active competition
- It should also add better caching in case we, e.g., get identical trade pairs within some short window

PR #986, opened 2025-07-24
feat(api): add middleware to reject certain endpoint requests if there is no active competition

joewagner:
fixes https://github.com/recallnet/js-recall/issues/985

This was an attempt at an inflight fix for the July 23-24 comp.  It wasn't the root of the issue at the time, but seems like a good performance improvement.

### DIFF
--- a/apps/api/src/middleware/active-comp-filter.middleware.ts
+++ b/apps/api/src/middleware/active-comp-filter.middleware.ts
@@ -19,11 +19,14 @@ export const activeCompMiddleware = function () {
     try {
       const activeCompetition = await getActiveComp();
       if (!activeCompetition) {
+        console.log("NO ACTIVE COMP!!");
         middlewareLogger.debug(
           "Active comp middleware: No active competition found",
         );
         throw new ApiError(403, "No active competition");
       }
+
+      req.competitionId = activeCompetition.id;
 
       next();
     } catch (error) {
@@ -37,7 +40,7 @@ export const activeCompMiddleware = function () {
 };
 
 // Cache for active competition check
-let cachedActiveCompetition: { id: string } | null = null;
+let cachedActiveCompetition: { id: string; name: string } | null = null;
 let lastQueryTime = 0;
 const CACHE_ACTIVE_COMP_TTL_MS = config.cache.activeCompetitionTtlMs;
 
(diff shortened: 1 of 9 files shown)


apps/api/src/middleware/active-comp-filter.middleware.ts
```diff
@@ -0,0 +1,73 @@
...
+      middlewareLogger.error(`Error checking active competition:`, error);
+      next(error);
+    }
+  };
+};
+
+// Cache for active competition check
+let cachedActiveCompetition: { id: string } | null = null;
+let lastQueryTime = 0;
+const CACHE_ACTIVE_COMP_TTL_MS = config.cache.activeCompetitionTtlMs;
```
joewagner, line 45: Since this is a very short (default 3 seconds) lived cache and a relatively low overhead query, seems fine to just have a per instance in-memory cache that can't be manually reset.  
If we revisit caching in a different layer of the app, we can potentially re-think this.

apps/api/src/middleware/active-comp-filter.middleware.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+export const activeCompMiddleware = function () {
+  return async function (req: Request, res: Response, next: NextFunction) {
+    try {
+      const activeCompetition = await getActiveComp();
+      if (!activeCompetition) {
+        middlewareLogger.debug(
+          "Active comp middleware: No active competition found",
+        );
+        throw new ApiError(403, "No active competition");
```
joewagner, line 26: Simply throw a 403 if there isn't an active competition.  This middleware can be used by any endpoint that is only valid when a competition is happening.

dtbuchholz (approved):
@joewagner lgmt. but, i noticed some odd behavior that we _could_ choose to fix in this PR. it's sort of related, and maybe we need to write tests for this case. but...it's sort of a separate issue, so we could create a follow up ticket to address the point below.

1. i have an agent trading in an active competition
2. competition ends
3. i try to make a `/trade/execute`
4. instead of this PR's error getting thrown, i get an "invalid api key" error. this is erroneous:
```
{
  "success": false,
  "error": "Invalid API key. This key may have been reset or is no longer associated with an active account. Please ensure you're using your most recent API key."
}
```
e.g., if i make a direct db update and alter the competition's status to `active` and `end_date` to null, i can make the API request in step (1). so, something odd is happening _maybe_ in the other middleware?

i _think_ i've noticed this happen with misc admin API requests, too.

joewagner:
> @joewagner lgmt. but, i noticed some odd behavior that we _could_ choose to fix in this PR. it's sort of related, and maybe we need to write tests for this case. but...it's sort of a separate issue, so we could create a follow up ticket to address the point below.
> 
> 1. i have an agent trading in an active competition
> 2. competition ends
> 3. i try to make a `/trade/execute`
> 4. instead of this PR's error getting thrown, i get an "invalid api key" error. this is erroneous:
> 
> ```
> {
>   "success": false,
>   "error": "Invalid API key. This key may have been reset or is no longer associated with an active account. Please ensure you're using your most recent API key."
> }
> ```
> 
> e.g., if i make a direct db update and alter the competition's status to `active` and `end_date` to null, i can make the API request in step (1). so, something odd is happening _maybe_ in the other middleware?
> 
> i _think_ i've noticed this happen with misc admin API requests, too.

Good catch!  Looks like there was a conditional in the auth middleware that looked specifically at the /api/trade/execute method and threw a 401.  I removed that and added two tests.  I also found a problem with the middleware logic, and had to update a bunch of tests to start a comp before getting prices.

This might be worth another look.

Issue #375, opened 2025-05-21
Fix API server logging via `pino` & env var directives


dtbuchholz:
**Problem Statement**

The `apps/api` package includes extensive logging across controllers, services, repositories (db queries), and middleware. However, it can be "too much"; a user should be able to opt-in to different logging levels. For exmple, if you look at the e2e tests in CI, it takes a minute or two for GitHub to load the window because there are simply way too many logs. Or if you run tests locally, it quickly fills up your terminal history to where you can see the actual test outputs.

**Proposed Solution**

Instead of `console.log` across everything, we can add an env var like `LOG_LEVEL` with different levels and use an external lib to add the proper scopes aligned to this settings.

**Impact**

Server devs

**Priority**

Low.

dtbuchholz:
we were at the point GitHub CI takes a few mins to load the e2e test window because there are too many logs. now, it wont even open and shows an error, asking you to manually download the logs: `This step has been truncated due to its large size. Download the full logs from the  menu once the workflow run has completed.`

<img width="1164" alt="Image" src="https://github.com/user-attachments/assets/2ad2b6e7-b8a0-463c-890f-e5dfcb649926" />

dtbuchholz:
this is becoming a problem for internal DX and needs to be fixed. anytime we do local dev or try to inspect CI failures, it's impossible to debug something because of the `console.log` logging. we need to move to `pino` or something similar.

moving to "ready" + "high"

PR #987, opened 2025-07-25
feat(api): replace console.log with pino

joewagner:
fixes https://github.com/recallnet/js-recall/issues/375

## TL;DR;

This is a huge PR that basically touches everything.  If we want to go with this approach, we should probably merge soonish.
There is still a large output from the tests, but if you look at the github logs, you can actually see the output of each test in the ui.  
Log level defaults to `info`, and you can set it with `LOG_LEVEL` evn var.  We need to make sure to set LOG_LEVEL to debug in production deployment to keep the same level of logs.  
A future PR can take advantage of pino's log rotation etc...

## Details (paraphrasing Claude)

Successfully converted all the `console.log` calls in the API server to use the Pino logging framework. Here's what was done:

### **Created Logger Configuration**
- Created `apps/api/src/lib/logger.ts` with a centralized logger configuration
- Set up environment-aware logging:
  - **Development**: Pretty-printed logs with colors and timestamps
  - **Production**: Structured JSON logs for better parsing
- Created context-specific loggers for different parts of the application:
  - `apiLogger`, `authLogger`, `dbLogger`, `tradeLogger`, `competitionLogger`
  - `adminLogger`, `userLogger`, `agentLogger`, `priceLogger`, `balanceLogger`
  - `repositoryLogger`, `middlewareLogger`, `serviceLogger`, `configLogger`

### **Maintained Log Context and Information**
- Preserved all the contextual information that was in the original console.log calls
- Used appropriate log levels:
  - `info()` for general information
  - `error()` for errors
  - `warn()` for warnings
  - `debug()` for debug information


dtbuchholz:
I can give this a test later today, but looks solid. and then I can update the containers too

dtbuchholz (approved):
@joewagner lgtm. i'll open a separate PR for the test-related cleanups (it's a large diff).

joewagner:
> @joewagner lgtm. i'll open a separate PR for the test-related cleanups (it's a large diff).
> 
> in terms of the prod server, what should we set the `LOG_LEVEL` to?

production should probably be `debug` since that is affectively what it's at now, and I think first step here is just enable log levels.  Next steps can be going through and deciding what level of logging to use and where.

PR #989, opened 2025-07-25
feat: docker compose for local testing

smrz2001:
First pass of Docker Compose config to build both API and UI components and run them locally.

The Dockerfiles could have been more optimized but this version at least somewhat reduces the build time.

### DIFF
--- a/apps/api/trade-simulator-docker/Dockerfile
+++ b/apps/api/trade-simulator-docker/Dockerfile
@@ -2,17 +2,21 @@ FROM node:22 AS builder
 
 WORKDIR /workdir
 
-COPY apps/api apps/api
-COPY packages/api-sdk packages/api-sdk
-COPY packages/eslint-config packages/eslint-config
-COPY packages/typescript-config packages/typescript-config
-COPY .npmrc .prettie* *.js *.json *.yaml ./
+RUN npm install -g pnpm typescript turbo bun
 
-# # Build the application
-RUN npm -g install pnpm typescript turbo
-RUN cd apps/api; pnpm install
-RUN cd packages/api-sdk; pnpm build
-RUN cd apps/api; pnpm build
+COPY package.json pnpm-*.yaml .npmrc ./
+COPY packages/eslint-config/package.json packages/eslint-config/
+COPY packages/typescript-config/package.json packages/typescript-config/
+COPY packages/api-sdk/package.json packages/api-sdk/
+COPY apps/api/package.json apps/api/
+
+RUN pnpm install --frozen-lockfile
+
+COPY . .
+
+# Build the application
+RUN pnpm --filter=@recallnet/api-sdk build && \
+    pnpm --filter=api build
 
 FROM node:22-alpine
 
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,96 @@
...
+    healthcheck:
+      test:
+        [
+          "CMD-SHELL",
+          "wget --no-verbose --tries=1 --spider http://127.0.0.1:${COMPS_PORT:-3001}/competitions || exit 1",
+        ]
+      interval: 600s
+      timeout: 5s
+      retries: 3
+      start_period: 1s
```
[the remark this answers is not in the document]
smrz2001: This was intentional.

I dislike seeing constant healthcheck logs on a test setup. If it's not working, I won't need the healthcheck to tell me, but I also don't have a strong opinion if people are in favor of shorter healthcheck intervals 🤷🏼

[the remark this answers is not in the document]
sam701: Maybe we should drop the healthchecks altogether?

apps/api/trade-simulator-docker/Dockerfile
```diff
@@ -1,18 +1,32 @@
+FROM node:22 AS deps
```
sam701: Just curious - what’s the reason for having a separate `deps` stage? We’re only using it in the `builder`, so would anything change if we combined them?

docker-compose.yml
```diff
@@ -0,0 +1,96 @@
+# This is a test docker-compose file that is not intended for production use!
+name: recall
+
+services:
+  db:
+    image: postgres
```
joewagner: Should we specify version 17 here to match whats running in supabase?

docker-compose.yml
```diff
@@ -0,0 +1,96 @@
...
+    volumes:
+      - "postgres-data:/var/lib/postgresql/data"
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U postgres -d postgres || exit 1"]
+      interval: 600s
+      timeout: 5s
+      retries: 3
+      start_period: 1s
+
+  redis:
```
joewagner, line 22: I've got redis running locally, and when I try to do `docker compose up` (with or without --build flag) i get the error: 
```
Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint recall-db-1 (dd57fe0efc161e3ac2fd515e706ea2a2ea9cabeecb8fbee2d54399a213786d69): failed to bind host port for 0.0.0.0:5432:172.23.0.3:5432/tcp: address already in use
```

joewagner: Can we keep redis limited to only be inside the docker network? i.e. so that the host port 6379 doesn't need to be available?

apps/api/trade-simulator-docker/Dockerfile
smrz2001: Not a strong reason, mainly for cleanliness - dependencies get built/cached in parallel and separately from the code.

I'm fine either way though, what do you think?

sam701: TBH I found it a bit confusing at first because the purpose wasn't immediately clear to me. Usually there is a build image (or a few of them) and then the final one.

smrz2001: Understood, let me collapse them. Will be easier to follow.

PR #990, opened 2025-07-25
fix(api): remove inefficient agent metrics query, leaderboard query, & throw errors 

dtbuchholz:
- Removes `calculateAgentMetrics` from the error fallback in `calculateBulkAgentMetrics`. Note: this method was triggered around 100 times in a 24 hour period due to timeouts in the `calculateBulkAgentMetrics` method, and `calculateAgentMetrics` calls `getAgentPortfolioSnapshots`...which is a problematic query.
- Alters the logic in `getLeaderboard` so that "pending" comps don't make unnecessary calls (i.e., a pending comp has effectively no "results" except the agent IDs)

### DIFF
--- a/apps/api/src/database/repositories/agentscore-repository.ts
+++ b/apps/api/src/database/repositories/agentscore-repository.ts
@@ -1,4 +1,4 @@
-import { desc, eq, sql } from "drizzle-orm";
+import { desc, eq, inArray, sql } from "drizzle-orm";
 import { v4 as uuidv4 } from "uuid";
 
 import { db } from "@/database/db.js";
@@ -38,11 +38,13 @@ export interface AgentRankInfo {
  * the agent information and rank score.
  * @returns An array of objects with agent ID, name, and rank score
  */
-async function getAllAgentRanksImpl(): Promise<AgentRankInfo[]> {
+async function getAllAgentRanksImpl(
+  agentIds?: string[],
+): Promise<AgentRankInfo[]> {
   console.log("[AgentRankRepository] getAllAgentRanks called");
 
   try {
-    const rows = await db
+    const query = db
       .select({
         id: agents.id,
         imageUrl: agents.imageUrl,
@@ -55,6 +57,12 @@ async function getAllAgentRanksImpl(): Promise<AgentRankInfo[]> {
       })
       .from(agentScore)
       .innerJoin(agents, eq(agentScore.agentId, agents.id));
+
+    if (agentIds) {
+      query.where(inArray(agentScore.agentId, agentIds));
+    }
+
+    const rows = await query;
 
     return rows.map((agent) => {
       return {
--- a/apps/api/src/database/repositories/competition-repository.ts
+++ b/apps/api/src/database/repositories/competition-repository.ts
@@ -70,14 +70,31 @@ const competitionOrderByFields: Record<string, AnyColumn> = {
 };
 
 interface Snapshot24hResult {
-  /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-  earliestSnapshots: Array<any>;
-  /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-  snapshots24hAgo: Array<any>;
+  earliestSnapshots: Array<{
+    id: number;
+    agentId: string;
+    competitionId: string;
+    timestamp: Date;
+    totalValue: number;
+  }>;
+  snapshots24hAgo: Array<{
+    id: number;
+    agentId: string;
+    competitionId: string;
+    timestamp: Date;
+    totalValue: number;
+  }>;
 }
 
 const snapshotCache = new Map<string, [number, Snapshot24hResult]>();
 const MAX_CACHE_AGE = 1000 * 60 * 5; // 5 minutes
+
+/**
+ * Clear the snapshot cache - used for testing
+ */
+export function clearSnapshotCache(): void {
+  snapshotCache.clear();
+}
 
 /**
  * Find all competitions
@@ -809,7 +826,7 @@ async function get24hSnapshotsImpl(
     `[CompetitionRepository] get24hSnapshotsImpl called for ${agentIds.length} agents in competition ${competitionId}`,
   );
 
-  const cacheKey = competitionId + "-" + agentIds.join("-");
+  const cacheKey = `${competitionId}-${agentIds.join("-")}`;
   const cachedResult = snapshotCache.get(cacheKey);
   if (cachedResult) {
     const now = Date.now();
@@ -913,13 +930,15 @@ async function get24hSnapshotsImpl(
  * Get portfolio snapshots for an agent in a competition
  * @param competitionId Competition ID
  * @param agentId Agent ID
+ * @param limit Optional limit for the number of snapshots to return
  */
 async function getAgentPortfolioSnapshotsImpl(
   competitionId: string,
   agentId: string,
+  limit?: number,
 ) {
   try {
-    return await db
+    const query = db
       .select()
       .from(portfolioSnapshots)
       .where(
@@ -929,7 +948,13 @@ async function getAgentPortfolioSnapshotsImpl(
         ),
       )
       .orderBy(desc(portfolioSnapshots.timestamp));
-    // TODO: there's no limit here, this is a weakness in perf
+
+    // Apply limit if specified
+    if (limit !== undefined && limit > 0) {
+      return await query.limit(limit);
+    }
+
+    return await query;
   } catch (error) {
     console.error(
       "[CompetitionRepository] Error in getAgentPortfolioSnapshots:",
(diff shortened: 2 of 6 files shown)


apps/api/src/database/repositories/agentscore-repository.ts
```diff
@@ -56,6 +58,12 @@ async function getAllAgentRanksImpl(): Promise<AgentRankInfo[]> {
       .from(agentScore)
       .innerJoin(agents, eq(agentScore.agentId, agents.id));
 
+    if (agentIds) {
```
dtbuchholz, line 61: used below so we can properly sort the "pending" leaderboard and show the ranked agents before unranked ones

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -70,15 +70,32 @@ const competitionOrderByFields: Record<string, AnyColumn> = {
...
+  }>;
 }
 
 const snapshotCache = new Map<string, [number, Snapshot24hResult]>();
 const MAX_CACHE_AGE = 1000 * 60 * 5; // 5 minutes
 
+/**
+ * Clear the snapshot cache - used for testing
+ */
+export function clearSnapshotCache(): void {
```
dtbuchholz, line 95: used in only tests; not sure if actually useful

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -929,7 +948,13 @@ async function getAgentPortfolioSnapshotsImpl(
         ),
       )
       .orderBy(desc(portfolioSnapshots.timestamp));
-    // TODO: there's no limit here, this is a weakness in perf
+
+    // Apply limit if specified
+    if (limit !== undefined && limit > 0) {
```
dtbuchholz, line 953: when an agent requests its portfolio, we just get the most recent snapshot (limit 1)

apps/api/src/services/competition-manager.service.ts
```diff
@@ -532,7 +534,34 @@ export class CompetitionManager {
    */
   async getLeaderboard(competitionId: string) {
     try {
-      // First try to get from the competitions_leaderboard table
+      const competition = await findById(competitionId);
+      const competitionStatus = competition?.status;
+
+      // Default case: `COMPETITION_STATUS.PENDING` will have no leaderboard—so we just return the agents
```
dtbuchholz, line 540: this logic was added because the old logic would try to make some queries and perform portfolio calculations when a competition is "pending". 

a pending comp doesnt have trading data. we just return the agents in order of their global rank, and unranked agents come afterward.

apps/api/src/services/competition-manager.service.ts
```diff
@@ -532,7 +534,34 @@ export class CompetitionManager {
...
+          }))
+          .sort((a, b) => b.globalScore - a.globalScore)
+          .map(({ agentId, value, pnl }) => ({
+            agentId,
+            value,
+            pnl,
+          }));
+      }
+
+      // Case: an `COMPETITION_STATUS.ENDED` should have leaderboard entries in the `competitions_leaderboard` table, unless it was recently ended
```
dtbuchholz, line 564: when we `endCompetition`, there's a small race condition. we can't assume an "ended" competition will have data returned by `findLeaderboardByTradingComp`—but for "old" comps, this method should always return that cached data and never execute the logic below.

apps/api/src/services/competition-manager.service.ts
```diff
@@ -806,111 +834,14 @@ export class CompetitionManager {
...
-          }
-        }
-      }
-
-      return { pnl, pnlPercent, change24h, change24hPercent };
-    } catch (error) {
-      console.error(
-        `[CompetitionManager] Error calculating metrics for agent ${agentId}:`,
-        error,
+      throw new ApiError(
```
dtbuchholz, line 841: new 500 on `calculateBulkAgentMetrics`

dtbuchholz:
note: these three fail locally, for some reason
```
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Tests 3 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL   e2e  e2e/tests/snapshots.test.ts > get24hSnapshots Repository Function > should return consistent data structures for both arrays
AssertionError: expected 0 to be greater than 0
 ❯ e2e/tests/snapshots.test.ts:92:45
     90| 
     91|     // Both arrays should have snapshots for both agents
     92|     expect(result.earliestSnapshots.length).toBeGreaterThan(0);
       |                                             ^
     93|     expect(result.snapshots24hAgo.length).toBeGreaterThan(0);
     94| 

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/3]⎯

 FAIL   e2e  e2e/tests/snapshots.test.ts > get24hSnapshots Repository Function > should handle single agent correctly
AssertionError: expected 0 to be greater than 0
 ❯ e2e/tests/snapshots.test.ts:155:45
    153| 
    154|     // Should have at least one snapshot for the agent
    155|     expect(result.earliestSnapshots.length).toBeGreaterThan(0);
       |                                             ^
    156|     expect(result.snapshots24hAgo.length).toBeGreaterThan(0);
    157| 

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[2/3]⎯

 FAIL   e2e  e2e/tests/snapshots.test.ts > get24hSnapshots Repository Function > should maintain data consistency when snapshots exist at different times
AssertionError: expected 0 to be greater than 0
 ❯ e2e/tests/snapshots.test.ts:310:45
    308| 
    309|     expect(result).toBeDefined();
    310|     expect(result.earliestSnapshots.length).toBeGreaterThan(0);
       |                                             ^
    311|     expect(result.snapshots24hAgo.length).toBeGreaterThan(0);
    312| 

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[3/3]⎯

```

brunocalza (approved):

apps/api/src/database/repositories/agentscore-repository.ts
joewagner: I say we go with this right now, but just a callout, we need to avoid the were in array for agents queries unless we are super sure the max number of ids is < ~30.

joewagner (approved):
:rocket: Awesome!  Looks good on read through.  Also tests pass for me locally.

apps/api/src/database/repositories/agentscore-repository.ts
dtbuchholz: ah kk good callout https://github.com/recallnet/js-recall/issues/993

PR #999, opened 2025-07-28
perf: remove scheduler

brunocalza:


apps/api/src/config/index.ts
```diff
@@ -282,21 +282,14 @@ export const config = {
...
   priceCacheDuration: parseInt(process.env.PRICE_CACHE_MS || "30000", 10), // 30 seconds
   portfolio: {
-    // Default snapshot interval: 2 minutes (120000ms), configurable via env
-    snapshotIntervalMs: parseInt(
-      process.env.PORTFOLIO_SNAPSHOT_INTERVAL_MS || "120000",
-      10,
-    ),
+    // Default snapshot interval: 5 minutes (600000ms)
+    // The snapshot is taken and configured via cron
+    snapshotIntervalMs: parseInt("600000", 10),
```
brunocalza, line 287: I'm keeping this config to make sure I don't break `getRules` API ( https://github.com/recallnet/js-recall/blob/main/apps/api/src/controllers/competition.controller.ts#L369-L371)

dtbuchholz:
i ran this locally, and im noticing an error gets throw in the auto-end competition cron. the comp never auto-ends. this is with `LOG_LEVEL=debug` enabled.
Screenshot 2025-07-29 at 1 33 46 PM

i ran this [`init.sh`](https://github.com/recallnet/js-recall/issues/889) script, and then manually updated the `end_date` for the "active" competition to something a few mins in the future...not sure if that would have an impact here.

dtbuchholz:
> If I comment out that cache condition, the process works even with multiple comps. Not really sure what we should do, keep it like this because we'll only have one active competition, or remove that cache condition.

ooooh yup, this is the root cause. should we just remove `activeCompetitionCache` entirely? it's clearly buggy. plus, that code is over 3 months old (vide coded), and our caching strategy is inconsistent and needs a revamp

---

> Any chance you had two active competitions?

nah, the app logic only lets us start 1 active comp at a time, and i double checked the db data directly (in case i overrode the status).

> If I have two or if in my test I attempt to auto end two (one at each time), it ends up failing, and the reason is this [cache](https://github.com/recallnet/js-recall/blob/aa96f3ea022e35daaff45c9c88fc9e5df25c991b/apps/api/src/services/competition-manager.service.ts#L320-L324):

that sounds like the problem i'm having:
1. start/end comp 1 => manually triggered comp end via admin API
2. start/end comp 2 => manually triggered comp end via admin API
3. start comp 3 with an end date => use auto-end logic => fails to end (and _never_ ends it on subsequent checks)

and interesting, the logs say the "current" active competition is the comp from step (1) and not the previous comp in (2).

dtbuchholz:
@brunocalza if you think we should remove the caching, i created a separate PR https://github.com/recallnet/js-recall/pull/1013

i didnt spend any time trying to rethink the strategy or look into why the cache is incorrect, so that's the other alternative.

dtbuchholz (approved):
lgtm. 

fyi, it looks like we'll land some db migration changes in the next day or two...in case we want to get this cron infra stuff set up prior to that.

Issue #934, opened 2025-07-21
[UI/UX] Introduce Skill Filter on Leaderboard (initially locked to Crypto Trading)


jovirecall:
<img width="906" height="325" alt="Image" src="https://github.com/user-attachments/assets/9f5d4a88-7d46-40b2-8f07-48020b36c4ea" />

<img width="922" height="412" alt="Image" src="https://github.com/user-attachments/assets/ecb76e33-a63a-4e9a-9869-b46f50b4b004" />

<img width="898" height="365" alt="Image" src="https://github.com/user-attachments/assets/6cec2987-8928-4b8d-b34f-b8009f8a0135" />

**Problem Statement**
AgentRank is calculated per skill domain, yet the Leaderboard currently displays a single “Global Leaderboard” without communicating that the numbers shown are specific to Crypto Trading. We need a visible filter component that:
	•	Indicates Crypto Trading is the active context.
	•	Lays the groundwork for additional skills (image creation, text composition, etc.).
	•	Keeps existing leaderboard metrics (Rank, Score, Competitions, Votes) scoped to the selected skill.

**Scope**
	•	Swap the page heading from Global Leaderboard → Skill Leaderboard.
	•	Add a pill-style filter bar below the heading:
	•	Crypto Trading (active / selected state, primary style).
	•	A passive pill: More skills coming soon! (disabled).
	•	All leaderboard stats (header KPIs and table rows) must respect the selected skill (already true for Crypto Trading; this change only surfaces that constraint to users).
	•	URL remains /leaderboard for now—no routing change; 
	•	We will scope the feature of actually having multiple skills in a future task.

**Acceptance Criteria**
AC-1
Given I load the Leaderboard, then I see the heading “Skill Leaderboard” and a pill bar with Crypto Trading selected.
AC-2
Given the Crypto Trading pill is selected, then the table shows only agents ranked for that skill (current behavior) and the KPI boxes (“Active Agents”, “Total Trades”, “Volume Traded”) reflect the skill-scoped totals.
AC-3
Given I hover the disabled “More skills coming soon!” pill, then the cursor shows not-allowed and no action occurs.
AC-4
Given I reload the page, then Crypto Trading remains pre-selected (default).

**Out of Scope**
	•	Backend changes for multi-skill support.
	•	Implementing additional skills—the UI will display the placeholder pill only.

[...]

PR #1000, opened 2025-07-28
feat: leaderboard with skill type filter

Rcontre360:
#934

### DIFF
--- a/apps/comps/components/agent-profile/index.tsx
+++ b/apps/comps/components/agent-profile/index.tsx
@@ -1,38 +1,38 @@
 "use client";
 
-import { ExternalLink } from "lucide-react";
+import {ExternalLink} from "lucide-react";
 import Image from "next/image";
 import Link from "next/link";
 import React from "react";
 
 import Card from "@recallnet/ui2/components/card";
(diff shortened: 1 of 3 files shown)


dtbuchholz:
> Given I hover the disabled “More skills coming soon!” pill, then the cursor shows not-allowed and no action occurs.

@Rcontre360 if you click on "Trading Competitions" and then "More coming soon", you're able to click the button. this shouldn't happen. also, we want the cursor to disable on hover. iirc there might be something have to adjust in the `ui2`'s button component to make this possible.

Screenshot 2025-07-29 at 6 16 13 PM

apps/comps/constants/index.ts
```diff
@@ -1 +1,13 @@
...
+  "Crypto Trading",
+  "Traditional Investing",
+  "Sports Betting",
+  "Prediction Markets",
+  "Social and Chat",
+  "Art & Video Creation",
+  "Programming / Coding",
+  "Deep Research",
+  "Other",
+];
```
ramirotw: nit:

```suggestion
] as const;
```

ramirotw (approved):
lgtm

Issue #881, opened 2025-07-15
[BUG] CI e2e test fail inconsistently


dtbuchholz:
**Bug Description**
- The CI database is having issus. For example, we see errors like `role "root" does not exist` or missing tables. It's not clear why because, eventually, the CI tests will typically pass. It's holding up PRs since it might be hours until it works correclty.
- See example: https://github.com/recallnet/js-recall/actions/runs/16306261105/job/46052907407
- Actual logs (github wont load them in the UI): [logs_41782635385.zip](https://github.com/user-attachments/files/21244942/logs_41782635385.zip)

dtbuchholz:
tip: go to "explain error" in the github UI to get some quick context
<img width="1209" height="676" alt="Image" src="https://github.com/user-attachments/assets/ca21cac3-67a1-4a42-a89e-fb359174b51e" />

dtbuchholz:
related is the error logging https://github.com/recallnet/js-recall/issues/375

smrz2001:
I downloaded and cleaned the job run logs. These are the failures causing the tests to fail:
```
2025-07-15T23:16:50.8219072Z ⎯⎯⎯⎯⎯⎯ Failed Tests 11 ⎯⎯⎯⎯⎯⎯⎯
2025-07-15T23:16:50.8219548Z 
2025-07-15T23:16:50.8222324Z  FAIL   e2e  e2e/tests/competition.test.ts > Competition API > Trophy Logic > should populate trophies correctly via user-specific endpoints (SIWE)
2025-07-15T23:16:50.8239647Z AssertionError: expected false to be true // Object.is equality
2025-07-15T23:16:50.8240220Z 
2025-07-15T23:16:50.8240488Z - Expected
2025-07-15T23:16:50.8240893Z + Received
2025-07-15T23:16:50.8241100Z 
2025-07-15T23:16:50.8241263Z - true
2025-07-15T23:16:50.8241626Z + false
2025-07-15T23:16:50.8241803Z 
2025-07-15T23:16:50.8242338Z  ❯ e2e/tests/competition.test.ts:3350:44
2025-07-15T23:16:50.8257088Z     3348|         agent1.id,
2025-07-15T23:16:50.8257709Z     3349|       );
2025-07-15T23:16:50.8258471Z     3350|       expect(agent1ApiKeyResponse.success).toBe(true);
2025-07-15T23:16:50.8259627Z        |                                            ^
[...]

PR #1005, opened 2025-07-29
fix(api): refactoring the e2e test startup and shutdown

joewagner:
fixes: #881 
This has turned into a large refactor.  There's a few categories of changes.

## duplicated setup and teardown logic

I'm not really sure how our setup and teardown logic was generated, but there was quite a bit of duplicated logic.  There were also a few comments mentioning Jest, which we have never used in this package afaik.  I'm assume there was some LLM hallucinations somewhere in the past that made it onto `main`.  Another thing I found that is a little tricky if you aren't familiar with Vitest is that the config is set so that `e2e/tests/utils/test-setup.ts` file is run for every file, and the `e2e/setup.ts` file is only run globally once.  I think this had resulted in some unnecessary duplication of setup and teardown.

## incorrect handling of process shutdown

The e2e tests are run via Vitest.  During startup for the tests, the `server.ts` file uses the Node.js child_process spawn function to start the api server in a sub-process.  The command to start that server is `npx tsx src/index.ts`.  [`tsx`](https://www.npmjs.com/package/tsx) is very useful in that it allows running typescript as if it was node.js flavored javascript.  The way this works involves starting 4 different processes that all need to be managed by `tsx`, and although shutting down the `tsx` command with `ctrl+c` works when you run the command in your shell, it does not work how you might expect when trying to do `process.kill(pid)` for the sub-process started by the `server.ts` file.  The solution I found is to use the [`tree-kill`](https://www.npmjs.com/package/tree-kill) to handle killing all of the processes.
I also removed the `detached: true` option from the process.spawn call since we want to be able to do `ctrl+c` while the tests are running and actually kill everything. (which was not possible before)

## confusion on in-memory availability of services during test running.

As mentioned above the actual server under test is in a sub-process, so we can't access things like `setInterval` and in-memory caching.  There were a few places in the tests and setup files where we were trying to interact with Services classes that were instantiated in the test runner process, which doesn't do really do anything.


CLAUDE.md
```diff
@@ -15,7 +15,6 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 **Testing:**
 - API: `pnpm --filter api test` - Unit tests
 - API: `pnpm --filter api test:e2e` - E2E tests
-- API: `pnpm --filter api test:e2e:runner` - Run E2E test suite
```
joewagner, line 18: I removed this outdated npm script

apps/api/e2e/setup.ts
```diff
@@ -1,31 +1,27 @@
...
-import { dbManager } from "./utils/db-manager.js";
-import {
-  killExistingServers,
-  startServer,
-  stopServer,
-} from "./utils/server.js";
+import { createLogger } from "@/lib/logger.js";
 
-// Store global server reference
-let server: Server;
```
joewagner, line 20: Here's an example of a place where the Server instance is not that actual server under test, and really doesn't have any purpose.

apps/api/e2e/tests/admin.test.ts
```diff
@@ -36,19 +34,8 @@ describe("Admin API", () => {
 
   // Clean up test state before each test
   beforeEach(async () => {
-    await cleanupTestState();
```
joewagner, line 39: Calling this in individual test files doesn't make sense.  We can just do it once in the test-setup.ts file and make sure it happens for every test.

apps/api/e2e/tests/admin.test.ts
```diff
@@ -36,19 +34,8 @@ describe("Admin API", () => {
...
-      username: ADMIN_USERNAME,
-      password: ADMIN_PASSWORD,
-      email: ADMIN_EMAIL,
-    });
-
     // Store the admin API key for authentication
-    adminApiKey = response.data.admin.apiKey;
-    expect(adminApiKey).toBeDefined();
-    console.log(`Admin API key created: ${adminApiKey.substring(0, 8)}...`);
+    adminApiKey = await getAdminApiKey();
```
joewagner, line 38: added a helper since this code was duplicated a dozen or so times

apps/api/e2e/utils/server.ts
```diff
@@ -1,6 +1,6 @@
 import axios from "axios";
 import { ChildProcess, spawn } from "child_process";
-import { Server } from "http";
+import kill from "tree-kill";
```
joewagner, line 3: This package is a new dependency and is key to actually killing all the processes that get started here.

apps/api/e2e/utils/server.ts
```diff
@@ -18,167 +18,64 @@ const BASE_URL = `http://${HOST}:${PORT}${API_PREFIX ? `/${API_PREFIX}` : ""}`;
  *
  * @returns A promise that resolves to the HTTP server instance
  */
-export async function startServer(): Promise<Server> {
-  // First try to kill any existing servers on the test port
-  await killExistingServers();
+export async function startServer(): Promise<void> {
+  if (serverProcess) {
+    throw new Error("Cannot start test server twice");
```
joewagner, line 23: because we had no check here and the `startServer` function was being called more than once due to confusion on how setup files work, we were sometime trying to start servers repeatedly.  Because the port was in use this may have gone unnoticed.

apps/api/e2e/utils/server.ts
```diff
@@ -18,167 +18,64 @@ const BASE_URL = `http://${HOST}:${PORT}${API_PREFIX ? `/${API_PREFIX}` : ""}`;
...
-          ...process.env,
-          NODE_ENV: "test",
-          PORT: testPort,
-          HOST: testHost,
-          METRICS_PORT: process.env.METRICS_PORT || "3003",
-          METRICS_HOST: "127.0.0.1", // Secure binding for tests
-          TEST_MODE: "true",
-        },
-        stdio: "inherit",
-        detached: true,
```
joewagner, line 45: By removing the detached option, we can now just do `ctrl+c` and actually stop the e2e tests.

apps/api/e2e/utils/server.ts
```diff
@@ -18,167 +18,64 @@ const BASE_URL = `http://${HOST}:${PORT}${API_PREFIX ? `/${API_PREFIX}` : ""}`;
...
+          "code" in error &&
+          error.code === "ESRCH"
+        )
+      ) {
+        // If there is some other error, we can log it and throw.
+        console.error("Error killing server process:", error);
+        throw error;
       }
-
-      const killProcess = spawn(command, args, { stdio: "pipe" });
```
joewagner, line 165: We have been blindly killing processes based only on the port they use.  This is quite dangerous if something important happened to be using port 3001.

apps/api/e2e/utils/server.ts
```diff
@@ -18,167 +18,64 @@ const BASE_URL = `http://${HOST}:${PORT}${API_PREFIX ? `/${API_PREFIX}` : ""}`;
...
+    } catch (error: unknown) {
+      // If there is not a process with the expected PID we don't have to do
+      // anything since it's already been killed. This allows calling this
+      // function repeatedly.
+      if (
+        !(
+          typeof error === "object" &&
+          error !== null &&
+          "code" in error &&
+          error.code === "ESRCH"
```
joewagner, line 71: `ESRCH` means the process doesn't exist.

apps/api/e2e/utils/server.ts
```diff
@@ -198,35 +95,28 @@ async function waitForServerReady(
...
+        console.log(
+          `Metrics server health check: ${metricsHealthResponse.status === 200 ? "OK" : "Failed"}`,
+        );
 
         return;
       }
-      // eslint-disable-next-line @typescript-eslint/no-unused-vars
     } catch (error) {
+      // @ts-expect-error - error is unknown type from catch block
+      if (error?.code !== "ECONNREFUSED") {
```
joewagner, line 117: `ECONNREFUSED` will be the code if the server isn't ready yet.  Otherwise we can throw the error and stop the tests.

apps/api/e2e/utils/test-helpers.ts
```diff
@@ -254,34 +238,6 @@ export async function startExistingTestCompetition(
...
-/**
- * Clean up database state for a given test case
- * This can be used in beforeEach to ensure a clean state
- * Now delegated to the DbManager for consistency
- */
-export async function cleanupTestState(): Promise<void> {
-  await ensureDatabaseInitialized();
-
-  // Also reset rate limiters to ensure clean state between tests
-  resetRateLimiters();
```
joewagner, line 277: Here's another spot were the sub-process rate limiters are the only thing we can about, but since this is interacting with a Service instance that is in the test runner this call to `resetRateLimiters` doesn't do anything.

apps/api/e2e/utils/test-setup.ts
```diff
@@ -1,19 +1,23 @@
...
+ *
+ * NOTE: The actual server under test is started inside a sub-process, hence in
+ *   memory things like caches, timeouts, and intervals cannot be interacted with.
+ *
+ * NOTE: that if you are running --isolate=false, this setup file will be run
+ *   in the same global scope multiple times. Meaning, that you are accessing the
+ *   same global object before each test, so make sure you are not doing the same
+ *   thing more than you need.
  *
- * This is used to set up global Jest configurations and hooks
```
joewagner, line 4: Not sure where these Jest comments came from, but I tried to remove them all.

apps/api/e2e/utils/test-setup.ts
```diff
@@ -1,19 +1,23 @@
...
 
-import { ServiceRegistry } from "@/services/index.js";
-
 import { dbManager } from "./db-manager.js";
 
-const services = new ServiceRegistry();
-
+// TODO: is this log file needed if we use Pino?
 // Path to log file
 const logFile = path.resolve(__dirname, "../e2e-server.log");
```
joewagner, line 22: I don't totally understand why we have this log file, so I left it alone.  This PR is big enough, another can handle figuring this out.

apps/api/e2e/utils/test-setup.ts
```diff
@@ -245,19 +121,22 @@ afterAll(async () => {
...
         (error instanceof Error ? error.message : String(error)),
     );
   }
 });
 
-// Log test lifecycle events are now handled by LogReporter
+// Before every test
+beforeEach(async function () {
+  // Ensure database is initialized
+  await dbManager.initialize();
```
joewagner, line 135: Initialize before every test, and below we reset after every test.

apps/api/package.json
```diff
@@ -17,10 +17,6 @@
     "test": "vitest run --project unit",
     "test:watch": "vitest watch --project unit",
     "test:e2e": "vitest run --project e2e",
-    "test:e2e:watch": "vitest watch --project e2e",
```
joewagner, line 20: I removed these scripts:
- test:e2e:watch, there's no way anyone wants to restart teh e2e tests on every change imo.  If someone is using this, we can add it back.
- test:e2e:runner, not used anymore
- test:multi-chain and test:multi-chain:e2e, these are just shortcuts to running specific files, doesn't seem useful.

apps/api/src/database/db.ts
```diff
@@ -328,3 +328,30 @@ export async function migrateDb() {
 export async function seedDb() {
   await seed(db, schema);
 }
+
+/**
+ * Close all database connections
+ * This should be called during application shutdown to prevent connection leaks
+ */
+export async function closeDb(): Promise<void> {
```
joewagner, line 336: added this so we can gracefully close the database.

apps/api/src/middleware/errorHandler.ts
```diff
@@ -37,6 +39,9 @@ const errorHandler = (
 
   // Handle inactive agent errors
   if (err.message && err.message.includes("inactive")) {
+    // TODO: if an error has the word "inactive" in it's message, it does NOT mean
+    //  it's a 403. Instead of catching individual cases here, the check for
+    //  inactive agents should throw a 403, and this should be removed.
```
joewagner, line 44: Totally unrelated to this PR, but wanted to leave a comment for future work.

joewagner (commented):
rfr

apps/api/src/index.ts
```diff
@@ -284,12 +284,23 @@ const gracefulShutdown = async (signal: string) => {
...
+      try {
+        await closeDb();
+        apiLogger.info("[Shutdown] Database connections closed");
+      } catch (error) {
+        apiLogger.error(
+          "[Shutdown] Error closing database connections:",
+          error,
+        );
+      }
+
```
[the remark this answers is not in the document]
joewagner: This is a good point, but since the setTimeout on line 308 keeps the js runtime going for ten seconds, what we do here is ok.  If something takes longer than 10 seconds, we attempt to force exit.

apps/api/e2e/utils/server.ts
```diff
@@ -18,167 +18,60 @@ const BASE_URL = `http://${HOST}:${PORT}${API_PREFIX ? `/${API_PREFIX}` : ""}`;
...
-        // Unix-like (macOS, Linux) - kill processes on both ports
-        command = "bash";
-        args = [
-          "-c",
-          `(lsof -i:${testPort} -t | xargs -r kill -9); (lsof -i:${metricsPort} -t | xargs -r kill -9)`,
-        ];
+      console.log(`Killing server process with PID: ${serverProcess.pid}`);
+      // Negative PID to kill process group, i.e all child processes
+      kill(serverProcess.pid);
+      serverProcess = null;
```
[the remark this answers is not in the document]
joewagner: For `1`, this is a good catch.  I'll fix now.  
For `2`, the kill function is a synchronous function, we can't call it any other way.

PR #1007, opened 2025-07-29
feat(api): competition rewards

dtbuchholz:
closes https://github.com/recallnet/js-recall/issues/781


apps/api/e2e/tests/admin.test.ts
```diff
@@ -1154,62 +1154,9 @@ describe("Admin API", () => {
         },
       );
       expect(true).toBe(false); // Should not reach here
-    } catch (error) {
```
dtbuchholz, line 1157: note: since this PR introduces new zod schema changes, i noticed we gatekept certain fields like the competition start date or end date. imo, the point of the admin API is to let admin properly create or update a competition they've created, so it doesnt make sense to restrict the behavior like this. \

basically, an admin can't change the UUID, internal db dates (created/updated at), or the comp "status". everything else should be allowed.

apps/api/e2e/tests/competition.test.ts
```diff
@@ -261,7 +260,7 @@ describe("Competition API", () => {
     const buyTradeResponse = await agentClient.executeTrade({
       reason: "testing create comp with trading constraints",
       fromToken: config.specificChainTokens.eth.usdc,
-      toToken: VISION_TOKEN, // low 24h volume
+      toToken: config.specificChainTokens.svm.sol,
```
dtbuchholz, line 251: @joewagner i think we should remove `VISON_TOKEN` as part of #881. these tests started to fail because the token somehow became "tradable", and the test expected a trade failure.

joewagner: Definitely makes sense

apps/api/src/controllers/admin.controller.ts
```diff
@@ -880,10 +897,25 @@ export function makeAdminController(services: ServiceRegistry) {
...
+          throw new ApiError(
+            400,
+            `Invalid request format: ${bodyResult.error}`,
+          );
+        }
+
+        // Extract rewards and tradingConstraints from the validated data
+        const { rewards, tradingConstraints, ...competitionUpdates } =
+          bodyResult.data;
+        const updates = competitionUpdates as UpdateCompetition;
```
joewagner, line 796: Should we add to the [refine function](https://github.com/recallnet/js-recall/pull/1007/files#diff-3bd6fe9cfe145ad1739a596e05f37189aa55332d21c228b1e042a000aeaf671eR88) so that you can't change the endDate of an ended comp, and can't change the endDate of an active or pending comp to a date in the past?
Maybe some other checks I'm not thinking of?

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -54,6 +55,47 @@ import { PartialExcept } from "./types.js";
...
 
+/**
+ * Builds the base competition query with rewards included
+ * @returns A query builder for competitions with rewards
+ */
+function buildCompetitionWithRewardsQuery() {
+  return db
+    .select({
+      crossChainTradingType: tradingCompetitions.crossChainTradingType,
+      rewards: sql<
```
joewagner, line 66: This query selects every competition right?  This might be a problem when we get more competitions.

I'm curious about something else here...

If you look here: https://github.com/recallnet/js-recall/blob/main/apps/api/src/controllers/competition.controller.ts#L467
This is the controller for the `/api/competitions` endpoint.  
It calls `services.competitionManager.getCompetitions`, which calls `findByStatus`, which calls this function `buildCompetitionWithRewardsQuery`.
I'm thinking that this returns the rewards for all the competitions, but if you go back to `competition.controller.ts` you'll see that there is a separate lookup for the rewards on line 510.
So are we double looking up the rewards data?

apps/api/src/database/repositories/competition-rewards-repository.ts
```diff
@@ -0,0 +1,313 @@
...
+/**
+ * Find all rewards for multiple competitions
+ * @param competitionIds Array of competition IDs
+ * @returns Array of reward records grouped by competition ID
+ */
+export async function findRewardsByCompetitionsImpl(
+  competitionIds: string[],
+): Promise<SelectCompetitionReward[]> {
+  try {
+    if (competitionIds.length === 0) {
```
joewagner, line 108: I guess this is ok for competitions right now, but in general we might want to setup a check for maximum array size when queries are going to use `where(inArray(...`.

apps/api/src/database/repositories/competition-rewards-repository.ts
```diff
@@ -0,0 +1,313 @@
...
+    await db.transaction(async (tx) => {
+      const promises = updates.map((update) => {
+        if (!update) return null;
+        return tx
+          .update(competitionRewards)
+          .set({ agentId: update.agentId })
+          .where(eq(competitionRewards.id, update.id));
+      });
+
+      await Promise.all(promises);
```
joewagner: I'm not 100% sure about this, but I don't know if using `Promise.all` with Drizzle's transaction feature is correct.  This makes is seem like we are sending everything in parallel, which won't be the case.  I'm not sure how Drizzle handles this, but I worry this might not work as expected. fwiw I asked Gemini, and it says not to do this.  It suggest it would be using a for loop.

apps/api/src/types/index.ts
```diff
@@ -710,8 +710,14 @@ export const CompetitionAllowedUpdateSchema = z.strictObject({
...
   externalUrl: z.string().optional(),
   imageUrl: z.string().optional(),
+  startDate: z.date().optional(),
+  endDate: z.date().optional(),
   votingStartDate: z.date().optional(),
   votingEndDate: z.date().optional(),
+  joinStartDate: z.date().optional(),
+  joinEndDate: z.date().optional(),
+  rewards: z.record(z.number(), z.number()).optional(),
+  tradingConstraints: z.record(z.string(), z.unknown()).optional(),
```
joewagner: Can we use the `TradingConstraintsSchema` schema defined in admin.schema.ts?

apps/comps/components/rewards.tsx
```diff
@@ -0,0 +1,94 @@
...
+    3: { label: "3rd", color: "text-[#C76E29]" },
+  };
+
+  const sorted = [...rewards].sort((a, b) => a.rank - b.rank);
+
+  return (
+    <div className={cn("flex flex-col gap-2", className)}>
+      <div className="text-primary-foreground font-bold">
+        $
+        {total.toLocaleString(undefined, {
```
joewagner: When the rewards are `0`, this says "$0 in rewards!".  Which is kind of funny, maybe not correct to have an exclamation piont?

apps/api/src/database/repositories/competition-repository.ts
ramirotw: > This query selects every competition right? This might be a problem when we get more competitions.

it's a dynamic query which gets filtered and paginated before executing, this shouldn't be a problem. Let me know if that works for you or if I missed something.

> I'm curious about something else here...
If you look here: https://github.com/recallnet/js-recall/blob/main/apps/api/src/controllers/competition.controller.ts#L467
This is the controller for the /api/competitions endpoint.
It calls services.competitionManager.getCompetitions, which calls findByStatus, which calls this function buildCompetitionWithRewardsQuery.
I'm thinking that this returns the rewards for all the competitions, but if you go back to competition.controller.ts you'll see that there is a separate lookup for the rewards on line 510.
So are we double looking up the rewards data?

you are right, good call. I've removed the 2nd call.

apps/api/src/database/repositories/competition-rewards-repository.ts
ramirotw: just in case I replaced the map and Promise.all calls in favor of a for..of loop with inner await calls. Let me know if that makes sense

apps/comps/components/rewards.tsx
ramirotw: changed to TBA when the competition doesn't have the rewards set

apps/api/src/types/index.ts
ramirotw: https://github.com/recallnet/js-recall/pull/1007/commits/cd7af55f61a03c8853db3f916eb2fe5ff3bdf563 does that work for you?

joewagner: perfect!

ramirotw:
@dtbuchholz @joewagner https://github.com/recallnet/js-recall/actions/runs/16790772309/job/47551808117?pr=1007 

> Error: src/server.ts(27,48): error TS2307: Cannot find module '@recallnet/api-sdk/mcp-server/tools/priceGetApiPriceTokenInfo.js' or its corresponding type declarations.

you know what's this error about?

joewagner:
> @dtbuchholz @joewagner https://github.com/recallnet/js-recall/actions/runs/16790772309/job/47551808117?pr=1007
> 
> > Error: src/server.ts(27,48): error TS2307: Cannot find module '@recallnet/api-sdk/mcp-server/tools/priceGetApiPriceTokenInfo.js' or its corresponding type declarations.
> 
> you know what's this error about?

This is weird.  It all works locally and I don't see any changes that would cause and issue only for this PR.
Claude wants to change the package.json in apps/api-sdk, which of course gets overwritten when it does the speakeasy build.

dtbuchholz:
Hm, we [removed](https://github.com/recallnet/js-recall/commit/c4f0dd826c91ed98917e7f4da0338911b8a5c3a2) that API recently. I wonder have expected a build failure, but I'll make a separate PR to regenerate that and merge it into main, which should fix this.

dtbuchholz:
this should fix it: https://github.com/recallnet/js-recall/pull/1040

PR #1013, opened 2025-07-29
fix(api): remove active comp cache

dtbuchholz:
the "auto end competition" logic fails to trigger due to faulty "active competition" cache. we observed:
1. start/end comp 1 => manually triggered comp end via admin API
2. start/end comp 2 => manually triggered comp end via admin API
3. start comp 3 with an end date => use auto-end logic => fails to end (and never ends it on subsequent checks)

the logs state the "current" active competition is the comp from step (1) and not the previous comp in (2). thus, when you try to auto-end (3), it fails.

---

furthermore, the problem with the current caching is that it auto-loads the current active competition upon service startup and pre-sets the cache with this value. since we're pulling the scheduler _out of the API_ and into its own cron, this adds a risk point for things to get out of sync. hence, why it should be removed (and/or we make this more sophisticated in the future).

### DIFF
--- a/apps/api/src/services/competition-manager.service.ts
+++ b/apps/api/src/services/competition-manager.service.ts
@@ -293,11 +293,6 @@ export class CompetitionManager {
       throw new Error(`Competition is not active: ${competition.status}`);
     }
 
-    const activeCompetition = await findActive();
-    if (!activeCompetition || activeCompetition.id !== competitionId) {
-      throw new Error(`Competition is not the active one`);
-    }
-
     // Take final portfolio snapshots
     await this.portfolioSnapshotter.takePortfolioSnapshots(competitionId);
 


apps/api/src/services/competition-manager.service.ts
```diff
@@ -317,10 +293,9 @@ export class CompetitionManager {
       throw new Error(`Competition is not active: ${competition.status}`);
     }
 
-    if (this.activeCompetitionCache !== competitionId) {
-      throw new Error(
-        `Competition is not the active one: ${this.activeCompetitionCache}`,
-      );
+    const activeCompetition = await findActive();
```
brunocalza: I don't get this logic, we already have a findById and check if it's active above. Looks like duplicated logic

dtbuchholz: ah good catch, fixing

brunocalza (approved):
looks good

PR #1019, opened 2025-07-30
chore: Small cleanups

stbrody:
No functional changes, just code cleanups.

* Commit 1: small cleanup to PriceTracker.getPrice to use an early return
* Commit 2: change TradeSimulator.executeTrade to throw on error instead of returning an error object
* Commit 3: Break up TradeSimulator.executeTrade into many smaller helper functions
* Commit 4: Cursor rule to keep functions small

### DIFF
--- a/apps/api/src/controllers/trade.controller.ts
+++ b/apps/api/src/controllers/trade.controller.ts
@@ -115,7 +115,7 @@ export function makeTradeController(services: ServiceRegistry) {
         }
 
         // Execute the trade with optional chain parameters
-        const result = await services.tradeSimulator.executeTrade(
+        const trade = await services.tradeSimulator.executeTrade(
           agentId,
           competitionId,
           fromToken,
@@ -126,14 +126,10 @@ export function makeTradeController(services: ServiceRegistry) {
           chainOptions,
         );
 
-        if (!result.success) {
-          throw new ApiError(400, result.error || "Trade execution failed");
-        }
-
         // Return successful trade result
         res.status(200).json({
           success: true,
-          transaction: result.trade,
+          transaction: trade,
         });
       } catch (error) {
         next(error);
--- a/apps/api/src/services/price-tracker.service.ts
+++ b/apps/api/src/services/price-tracker.service.ts
@@ -97,56 +97,59 @@ export class PriceTracker {
       `[PriceTracker] ${blockchainType ? "Using provided" : "Detected"} token ${tokenAddress} on chain: ${tokenChain}`,
     );
 
-    // If no cache hit, use MultiChainProvider
-    if (this.multiChainProvider) {
-      try {
+    if (!this.multiChainProvider) {
+      serviceLogger.error(`[PriceTracker] No MultiChainProvider available`);
+
+      return null;
+    }
+
+    try {
+      serviceLogger.debug(
+        `[PriceTracker] Using MultiChainProvider for token ${tokenAddress}`,
+      );
+
+      // Get price from MultiChainProvider (which has its own cache)
+      const priceResult = await this.multiChainProvider.getPrice(
+        tokenAddress,
+        tokenChain,
+        specificChain,
+      );
+
+      if (priceResult !== null) {
+        // Handle both number and PriceReport return types
+        const price = priceResult.price;
+        const chain = priceResult.chain;
+
+        // For number results, ensure we have a valid specificChain
+        const tokenSpecificChain = priceResult.specificChain;
+
+        // Get the symbol
+        const symbol = priceResult.symbol;
+
         serviceLogger.debug(
-          `[PriceTracker] Using MultiChainProvider for token ${tokenAddress}`,
+          `[PriceTracker] Got price $${price} from MultiChainProvider`,
         );
 
-        // Get price from MultiChainProvider (which has its own cache)
-        const priceResult = await this.multiChainProvider.getPrice(
+        // Store price in database for historical record
+        await this.storePrice(
           tokenAddress,
-          tokenChain,
-          specificChain,
+          price,
+          symbol,
+          chain,
+          tokenSpecificChain,
         );
 
-        if (priceResult !== null) {
-          // Handle both number and PriceReport return types
-          const price = priceResult.price;
-          const chain = priceResult.chain;
-
-          // For number results, ensure we have a valid specificChain
-          const tokenSpecificChain = priceResult.specificChain;
-
-          // Get the symbol
-          const symbol = priceResult.symbol;
-
-          serviceLogger.debug(
-            `[PriceTracker] Got price $${price} from MultiChainProvider`,
-          );
-
-          // Store price in database for historical record
-          await this.storePrice(
-            tokenAddress,
-            price,
-            symbol,
-            chain,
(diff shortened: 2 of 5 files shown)


apps/api/src/controllers/trade.controller.ts
```diff
@@ -113,14 +113,10 @@ export function makeTradeController(services: ServiceRegistry) {
           chainOptions,
         );
 
-        if (!result.success) {
```
joewagner, line 129: removing this doesn't seem quite right.  :thinking: 
If `result.success` is equal to `false`, are you sure we want to return the hardcoded `success: true` in the response?

joewagner: Ok, I think I understand now.  Inside execute trade we are throwing if anything goes wrong, so this check isn't needed. :+1:

stbrody:
@joewagner rebased

apps/api/src/services/price-tracker.service.ts
```diff
@@ -97,56 +97,59 @@ export class PriceTracker {
...
+    if (!this.multiChainProvider) {
+      serviceLogger.error(`[PriceTracker] No MultiChainProvider available`);
 
-        // Get price from MultiChainProvider (which has its own cache)
-        const priceResult = await this.multiChainProvider.getPrice(
-          tokenAddress,
-          tokenChain,
-          specificChain,
-        );
+      return null;
```
joewagner, line 103: I much prefer the return based logic to nesting conditionals :heart:

apps/api/src/services/price-tracker.service.ts
```diff
@@ -97,56 +97,59 @@ export class PriceTracker {
...
+        await this.storePrice(
+          tokenAddress,
+          price,
+          symbol,
+          chain,
+          tokenSpecificChain,
+        );
+
+        return priceResult;
+      } else {
```
joewagner, line 143: There is a return right above this `else` block, so we don't really need the else.  Maybe remove the block and just have a the call to logger in the top level of the try block?
Not a big deal, so feel free to keep it if you prefer the style as is.

apps/api/src/services/trade-simulator.service.ts
```diff
@@ -448,21 +178,59 @@ export class TradeSimulator {
...
+   * @throws ApiError if validation fails
+   */
+  private validateTradeInputs(
+    fromToken: string,
+    toToken: string,
+    fromAmount: number,
+    reason: string,
+  ): void {
+    // Validate minimum trade amount
+    if (fromAmount < 0.000001) {
```
joewagner: Maybe hoist the ` 0.000001` value to a `const` at the top of the file, or move to config?

apps/api/src/services/trade-simulator.service.ts
```diff
@@ -717,52 +485,28 @@ export class TradeSimulator {
...
 
     // Check liquidity constraint
-    const liquidityValidationResult = this.validateLiquidConstraint(
-      priceData,
-      constraints,
-    );
-    if (!liquidityValidationResult.success) {
-      return liquidityValidationResult;
-    }
+    this.validateLiquidConstraint(priceData, constraints);
```
joewagner, line 508: Skipping the return and just throwing makes this way more clean to read :heart:

apps/api/src/services/trade-simulator.service.ts
```diff
@@ -840,82 +571,456 @@ export class TradeSimulator {
...
+    }
+  }
+
+  /**
+   * Validates cross-chain trading rules.
+   * @param chainInfo The resolved chain and specific chain information.
+   * @throws ApiError if cross-chain trading is not allowed.
+   */
+  private validateCrossChainTrading(chainInfo: ChainOptions): void {
+    switch (features.CROSS_CHAIN_TRADING_TYPE) {
```
joewagner: This is good to go, but in general doing things like `switch` -> `case` -> `if`, makes it seem like it could be a little more tidy. 
Maybe return base code where we throw if an invalid state is found?
e.g. instead of 
```
switch (features.CROSS_CHAIN_TRADING_TYPE) {
      case "disallowXParent":
        // Check if the tokens are on the same chain
        if (chainInfo.fromChain !== chainInfo.toChain) {
          serviceLogger.debug(
            `[TradeSimulator] Cross-parent chain trading is disabled. Cannot trade between ${chainInfo.fromChain} and ${chainInfo.toChain}`,
          );
          throw new ApiError(
            400,
            "Cross-parent chain trading is disabled. Both tokens must be on the same parent blockchain.",
          );
        }
        break;
```
do:
```
if (chainInfo.fromChain !== chainInfo.toChain && features.CROSS_CHAIN_TRADING_TYPE === "disallowXParent") {
  serviceLogger.debug(
    `[TradeSimulator] Cross-parent chain trading is disabled. Cannot trade between ${chainInfo.fromChain} and ${chainInfo.toChain}`,
  );
  throw new ApiError(
    400,
    "Cross-parent chain trading is disabled. Both tokens must be on the same parent blockchain.",
  );
}
```

apps/api/src/services/trade-simulator.service.ts
```diff
@@ -840,82 +571,456 @@ export class TradeSimulator {
...
+      // Normal trade with slippage
+      const { effectiveFromValueUSD: calculatedEffectiveValue } =
+        calculateSlippage(fromValueUSD);
+      effectiveFromValueUSD = calculatedEffectiveValue;
+      toAmount = effectiveFromValueUSD / toPrice;
+      exchangeRate = toAmount / fromAmount;
+    }
+
+    // Debug logging for price calculations
+    if (toPrice === 0) {
```
joewagner: Looks like a duplicated identical conditional. Would it make sense to move these logs into the if else above?

apps/api/src/services/trade-simulator.service.ts
```diff
@@ -840,82 +571,456 @@ export class TradeSimulator {
...
+      serviceLogger.debug(`[TradeSimulator] Burn trade calculation details:
+              From Token (${fromToken}):
+              - Amount: ${fromAmount}
```
joewagner: Linting passes so this more of a question of my own knowledge, but any reason for the big indent here?

apps/api/src/services/trade-simulator.service.ts
stbrody: made a constant for now

apps/api/src/services/price-tracker.service.ts
stbrody: yeah I don't feel like this matters too much so going to leave it just as is just to avoid merge conflicts

apps/api/src/services/trade-simulator.service.ts
stbrody: done

apps/api/src/services/trade-simulator.service.ts
stbrody: no idea, I just do what prettier tells me 😅

apps/api/src/services/trade-simulator.service.ts
stbrody: done

PR #1020, opened 2025-07-31
feat(api): unique agent handles & custom display name

### LINKED ISSUE #878: [FEATURE] Make agent and user names unique
## Problem Statement

User and agent names are not unique. This is prone to spoofing etc.

## Proposed Solution

There are three possible options:
1. Make agent names unique
2. Make both agent and user names unique, but only scoped to their respective `agents` or `users` tables
3. Make both agent and user names unique, scoped _globally_

Option (1) is what we _need_ to do. But, if users will play a role in the system (e.g., curators), we might as well also make their names unique via (2). And once we do that, that brings (3) into the fold. I'd probably err on the side of (2) because agents and users don't have an overlap in their capabilities. 

However, if this is not true (e.g., agent _can_ curate), then we need to consider (3). Or, the fact that we'd have two different "handles" across different types of actors (agents, users) could be confusing.

If option (2):

- We make the `users.name` and `agents.name` columns unique
- In our migration script, we use the `name` plus some counter so that duplicates are auto-incremented (e.g., `my_name`, `my_name0`, `my_name1`, etc.)
- Frontend should change routes from `/agents/<uuid>` to `/agents/<name>` (see PolyMarket or Pump.fun)

If option (3):

- Create a shared `actors` table that stores the unique name and track the owner (agent or user)

**Impact**
The per-agent rank _does_ uniquely "identify" an agent, to an extent, because some top ranked agent's profile will have different metadata than some low ranking spoof. However, it's too prone to user error and should be addressed.

**Priority**
High. We've seen tangential issues in existing competitions and should get ahead of this.

dtbuchholz:
closes #878


apps/api/src/controllers/admin.controller.ts
```diff
@@ -53,6 +54,7 @@ interface Agent {
   ownerId: string;
   walletAddress: string | null;
   name: string;
+  handle: string | null; // TODO(dtb): remove `null` once agent handle migration is complete
```
dtbuchholz: note: in order to run the backfill agent handles script, we have to first allow this to be null. once this PR is approved, i'll create a follow up to make this NOT NULL.

so, we will:
1. tear down the API
2. run the db first migration
3. run the backfill script
4. run the second db migration
5. restart the API

apps/api/src/controllers/admin.controller.ts
```diff
@@ -295,6 +298,7 @@ export function makeAdminController(services: ServiceRegistry) {
               agent = await services.agentManager.createAgent({
                 ownerId: user.id,
                 name: agentName,
+                handle: agentHandle ?? generateHandleFromName(agentName), // Auto-generate from name
```
dtbuchholz, line 301: for the admin APIs, we use this helper. but in non-admin APIs, the `handle` is a required field for agent creation.

apps/api/src/lib/handle-utils.ts
```diff
@@ -0,0 +1,64 @@
...
+ * Converts a name to a valid handle format
+ * @param name The display name to convert
+ * @returns A valid handle string
+ * @example
+ * generateHandleFromName("John Doe") // returns "john_doe"
+ * generateHandleFromName("Agent@123!") // returns "agent_123_"
+ * generateHandleFromName("My Cool Agent") // returns "my_cool_agent"
+ * generateHandleFromName("_underscore") // returns "_underscore"
+ */
+export function generateHandleFromName(name: string): string {
```
dtbuchholz, line 21: twitter's and Pump.fun rules:
- max 15 chars
- min 3 chars
- alphanumeric with underscores

note: only lowercase handles. else, we'd have to store the "raw" handle with uppercase values...which is a bit messier.

apps/api/src/services/agent-manager.service.ts
```diff
@@ -149,7 +154,35 @@ export class AgentManager {
       };
 
       // Store in database
-      const savedAgent = await create(agent);
+      let savedAgent;
+      try {
+        savedAgent = await create(agent);
+      } catch (error) {
+        if (error instanceof DatabaseError) {
```
dtbuchholz, line 161: added these to make it easier on the frontend to show which 409 is happening.

apps/api/src/types/index.ts
```diff
@@ -765,6 +776,23 @@ export const UpdateUserProfileSchema = z
   })
   .strict();
 
+/**
+ * Agent handle validation schema
+ */
+export const AgentHandleSchema = z
```
dtbuchholz, line 782: twitter's and Pump.fun rules:
- max 15 chars
- min 3 chars
- alphanumeric with underscores

note: only lowercase handles. else, we'd have to store the "raw" handle with uppercase values and return those purely for cosmetic info...which is a bit messier.

apps/api/e2e/utils/test-helpers.ts
```diff
@@ -81,20 +138,22 @@ export async function registerUserAndAgentAndGetClient({
...
   userImageUrl?: string;
   agentName?: string;
+  agentHandle?: string;
   agentDescription?: string;
   agentImageUrl?: string;
   agentMetadata?: Record<string, unknown>;
   agentWalletAddress?: string;
 }) {
-  const sdk = getApiSdk(adminApiKey);
+  const sdk = new ApiClient(adminApiKey);
```
dtbuchholz, line 147: note: i removed the Speakeasy SDK usage here to avoid regenerating it...idk what we want to do there, anymor

apps/api/scripts/populate-agent-handles.ts
dtbuchholz, line 1: i tested this script locally with a prod dump. it'll give priority to the `agent_score` entries so that, e.g., the "real" `moonsage_alpha` gets the handle. and dupes thereafter get a value appended to their name, like `moonsage_alpha1` etc

dtbuchholz:
note: i'll open two more PRs after this gets merged:
1. one with another db migration for NOT NULL on the handle
2. frontend changes to account for the agent's handle

PR #1023, opened 2025-07-31
feat: Make all db updates for finalizing a trade atomic

stbrody:
builds on https://github.com/recallnet/js-recall/pull/1019

closes https://linear.app/recall-labs/issue/REC-237/atomic-trade-execution

### DIFF
--- a/apps/api/src/database/repositories/balance-repository.ts
+++ b/apps/api/src/database/repositories/balance-repository.ts
@@ -1,4 +1,4 @@
-import { and, count as drizzleCount, eq, inArray } from "drizzle-orm";
+import { and, count as drizzleCount, eq, inArray, sql } from "drizzle-orm";
 
 import { config } from "@/config/index.js";
 import { db } from "@/database/db.js";
@@ -6,6 +6,8 @@ import { balances } from "@/database/schema/trading/defs.js";
 import { repositoryLogger } from "@/lib/logger.js";
 import { createTimedRepositoryFunction } from "@/lib/repository-timing.js";
 import { SpecificChain } from "@/types/index.js";
+
+type DatabaseTransaction = Parameters<Parameters<typeof db.transaction>[0]>[0];
 
 /**
  * Balance Repository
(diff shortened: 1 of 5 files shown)


brunocalza:
Just posting here for reference in case anyone does not know about this. The OOP way of solving this is with the [**Unit of Work**](https://martinfowler.com/eaaCatalog/unitOfWork.html) pattern. You do pretty much what is done in this PR, but you use a specialized class called Unit of Work that manage multiple repo calls within a tx. Here's a [ChatGPT code example](https://chatgpt.com/c/688d283f-8f28-8006-aa9d-5ead1b6c7c04). Not saying we should do this, but can be useful to follow a pattern if there are many places that need this.

apps/api/src/database/repositories/trade-repository.ts
```diff
@@ -1,23 +1,72 @@
...
+  };
+}> {
+  return await db.transaction(async (tx) => {
+    // Decrement the "from" token balance using the helper function
+    const fromTokenBalance = await decrementBalanceInTransaction(
+      tx,
+      trade.agentId,
+      trade.fromToken,
+      trade.fromAmount,
+      trade.fromSpecificChain as SpecificChain,
```
joewagner: We should avoid casting and instead use Zod to check the type here.

apps/api/src/database/repositories/balance-repository.ts
```diff
@@ -246,6 +198,136 @@ async function resetAgentBalancesImpl(
...
+
+      return newResult.amount;
+    } else {
+      // For decrements, the balance should already exist
+      throw new Error(
+        `Balance not found for agent ${agentId}, token ${tokenAddress}`,
+      );
+    }
+  } else if (amountDelta < 0 && result.amount < 0) {
+    throw new Error(
```
joewagner, line 263: This works well afaict, but curious if there's any advantage to using `tx.rollback()` instead of throwing?

stbrody: I guess I assumed that drizzle would roll back the transaction automatically if an Error is thrown during transaction execution

stbrody: Calling `rollback()` is literally just implemented as throwing an Error: https://github.com/drizzle-team/drizzle-orm/blob/8b8d78ec3d209ce3224b609d633b7ce090092889/drizzle-orm/src/pg-core/session.ts#L257

joewagner: Seems like both work.  I was mostly curious, I asked the llms and they say `tx.rollback` lets you return a value, throwing let's you specify a message.
This looks good!

apps/api/src/database/repositories/trade-repository.ts
stbrody: ooh good call, I've never seen Zod before but that's pretty spiffy.  Done

PR #1024, opened 2025-07-31
fix: block trades of unregistered agents

brunocalza:
Closes https://linear.app/recall-labs/issue/REC-49/bug-more-agents-in-the-snapshot-than-in-the-competition

### DIFF
--- a/apps/api/e2e/tests/trading.test.ts
+++ b/apps/api/e2e/tests/trading.test.ts
@@ -2362,4 +2362,89 @@ describe("Trading API", () => {
 
     console.log("✅ All token burn tests passed!");
   });
+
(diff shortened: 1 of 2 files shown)


apps/api/e2e/tests/trading.test.ts
```diff
@@ -2362,4 +2362,89 @@ describe("Trading API", () => {
 
     console.log("✅ All token burn tests passed!");
   });
+
+  test("agent cannot execute trade on competition they are not registered for", async () => {
```
brunocalza, line 2366: adds a test that confirms the bug

dtbuchholz (approved):

PR #1026, opened 2025-08-05
chore: Remove this.getPriceForSpecificEVMChain

stbrody:


apps/api/e2e/tests/price/multi-chain-provider.test.ts
```diff
@@ -276,13 +276,11 @@ describe("Multi-Chain Provider Tests", () => {
...
-          // Get detailed token info to see which chain it was found on
-          const tokenInfo = await multiChainProvider.getTokenInfo(
-            unknownToken,
-            BlockchainType.EVM,
-          );
-          if (tokenInfo && tokenInfo.specificChain) {
-            console.log(`Token was found on chain: ${tokenInfo.specificChain}`);
+          // The price report already contains chain information
+          if (priceReport && priceReport.specificChain) {
+            console.log(
```
dtbuchholz, line 281: lets remove `console.log` from tests

apps/api/e2e/tests/price/multi-chain-provider.test.ts
```diff
@@ -411,19 +394,17 @@ describe("Multi-Chain Provider Tests", () => {
...
-      expect(trackerTokenInfo).not.toBeNull();
-      if (trackerTokenInfo) {
-        expect(trackerTokenInfo.chain).toBe(BlockchainType.EVM);
-        if (trackerTokenInfo.price !== null) {
-          expect(trackerTokenInfo.price).toBeGreaterThan(0);
-        }
+      // The price report should have info regardless of the price
+      expect(trackerPriceReport).not.toBeNull();
+      expect(trackerPriceReport!.chain).toBe(BlockchainType.EVM);
+      if (trackerPriceReport!.price !== null) {
```
dtbuchholz, line 406: we should use assertions instead of `if` statements (it's a problem in many of the tests)

apps/api/e2e/tests/price/multi-chain-provider.test.ts
```diff
@@ -468,57 +449,47 @@ describe("Multi-Chain Provider Tests", () => {
...
-        // Verify we got a result
-        expect(tokenInfo).not.toBeNull();
+        // Chain type should always be correctly identified
+        expect(priceReport!.chain).toBe(BlockchainType.EVM);
 
-        if (tokenInfo) {
-          // Chain type should always be correctly identified
-          expect(tokenInfo.chain).toBe(BlockchainType.EVM);
+        // If we got a price, verify it looks reasonable
+        if (priceReport!.price !== null) {
```
dtbuchholz, line 459: (i wont comment elsewhere, but same theme here: lets get rid of `if`s and `console.logs`.)

apps/api/src/services/providers/__tests__/multi-chain.provider.test.ts
```diff
@@ -111,44 +111,45 @@ describe("MultiChainProvider", () => {
       const duration = Date.now() - start;
 
       // Second call should be very quick (under 5ms) as it's using the cache
+      // TODO(stbrody): This is a bad way to check if the cache is being used. Make this check more targeted and less flaky.
```
dtbuchholz, line 114: note: there are a few TODO comments. do we want to address these in this PR, or follow up?

apps/api/src/services/providers/multi-chain.provider.ts
```diff
@@ -126,88 +91,30 @@ export class MultiChainProvider implements PriceSource {
...
+            `[MultiChainProvider] Getting price for token ${normalizedAddress} on specific chain ${specificChain} with type ${detectedChainType}`,
           );
 
           // Use DexScreenerProvider to get price for a specific chain
-          const price = await this.getPriceForSpecificEVMChain(
+          const price = await this.dexScreenerProvider.getPrice(
             normalizedAddress,
-            specificChain,
+            detectedChainType,
+            specificChain!,
```
dtbuchholz: we should avoid `!` operators

stbrody: oops, not sure how that slipped in, it's not even needed here.  Removed.

apps/api/src/services/providers/__tests__/multi-chain.provider.test.ts
stbrody: in follow-ups, these are reminders for later

stbrody (commented):
agreed about removing the if-statements from tests, but given how much else is happening in this PR, I'd like to separate that type of change so if the tests do start failing we can be clear on where the issue stems from.  I'll put together a follow-up PR to at least clean up the tests that are changed in this PR

PR #1022, opened 2025-07-31
feat(api): add portfolio tracking endpoint

brunocalza:
# Summary
Closes #950 

Adds the  `/competitions/:id/performance` endpoint for https://github.com/recallnet/js-recall/issues/490

# Details
Response example:
```json
{
  "success": true,
  "competitionId": "af7fab90-189f-4473-b334-0445b4c15c3e",
  "performance": [
    {
      "agentId": "427fdde1-8e44-4429-8d78-3a7b02c79477",
      "agentName": "Agent 0",
      "timeline": [
        {
          "date": "2025-07-31",
          "totalValue": 14951.7377867605
        }
      ]
    },
    {
      "agentId": "c1f0dfe9-9821-4574-a5b3-2a57834791d4",
      "agentName": "Agent 1",
      "timeline": [
        {
          "date": "2025-07-31",
          "totalValue": 14951.8036959766
        }
      ]
    },
    {
      "agentId": "e6929720-19eb-4db3-965f-13c63638fb6f",
      "agentName": "Agent 2",
      "timeline": [
        {
          "date": "2025-07-31",
          "totalValue": 14951.8036959766
        }
      ]
    },
    {
      "agentId": "05664ac6-9a14-4af0-9692-4df098680313",
      "agentName": "Agent 3",
      "timeline": [
        {
          "date": "2025-07-31",
          "totalValue": 14951.8036959766
        }
      ]
    },
    {
      "agentId": "f660aa6a-561b-436d-97d7-72f765c04edb",
      "agentName": "Agent 4",
      "timeline": [
        {
          "date": "2025-07-31",
          "totalValue": 14951.8036959766
        }
      ]
    },
    {
      "agentId": "c42df7e1-2d0f-4e4a-8562-cec36bcb114d",
      "agentName": "Agent 5",
      "timeline": [
        {
          "date": "2025-07-31",
          "totalValue": 14951.7384251849
        }
      ]
    },
    {
      "agentId": "60627532-db1b-4f23-bbd7-eebebfeed138",
      "agentName": "Agent 6",
      "timeline": [
        {
          "date": "2025-07-31",
          "totalValue": 14951.8036959766
        }
      ]
    },
    {
      "agentId": "fcd1dbbf-b681-4761-98dd-e907c8b8bdb2",
      "agentName": "Agent 7",
      "timeline": [
        {
          "date": "2025-07-31",
          "totalValue": 14951.8036959766
        }
      ]
    },
    {
[...]

PR #1028, opened 2025-08-05
Rafael/portfolio performance

Rcontre360:


dtbuchholz:
note: i was testing https://github.com/recallnet/js-recall/pull/1022 against this and ended up making a handful of tweaks / fixes. the designs were a bit old and didn't describe some UX related behavior, and because of that, it's easier if i just take this PR over. i'll open a separate PR that'll supersede this one.

PR #1029, opened 2025-08-05
test(api): fix flaky token tests

dtbuchholz:


apps/api/e2e/tests/chain-specific.test.ts
```diff
@@ -277,14 +277,14 @@ describe("Specific Chains", () => {
...
-      looseConstraints,
+      noTradingConstraints,
     );
 
     // Get agent's current balances
     const balanceResponse = (await client.getBalance()) as BalancesResponse;
     expect(balanceResponse.success).toBe(true);
 
-    // Target token we want to purchase
+    // Target token we want to purchase (USDC on optimism)
```
dtbuchholz: usdc on OP apparently has terribly low volume, so we simply remove the trading constraints for this test

apps/api/e2e/tests/multi-agent-competition.test.ts
```diff
@@ -339,9 +324,6 @@ describe("Multi-Agent Competition", () => {
       if (tradeResponse.transaction.toAmount) {
         const tokenAmount = tradeResponse.transaction.toAmount;
         tokenQuantities[tokenToTrade!] = tokenAmount;
-        console.log(
```
dtbuchholz, line 342: note: we should make sure new tests dont have logging...and/or just give someone a half day to blast through removing logging in all existing tests so that LLMs stop recreating this behavior.

apps/api/e2e/tests/user.test.ts
```diff
@@ -2804,33 +2803,33 @@ describe("User API", () => {
...
     }
 
-    // Alpha Agent: (3 big stable trades)
+    // Alpha Agent: (3 small bad trades)
     for (let i = 0; i < 3; i++) {
       await agentClients[0]?.executeTrade({
         fromToken: config.specificChainTokens.eth.usdc,
-        toToken: config.specificChainTokens.eth.eth,
-        amount: "1000",
+        toToken: "0x000000000000000000000000000000000000dead",
```
dtbuchholz, line 2810: this test just wants deterministic ordering, so we burn tokens in different amounts instead of relying on potentially flaky token behavior

apps/api/e2e/tests/multi-agent-competition.test.ts
dtbuchholz: https://linear.app/recall-labs/issue/REC-290/remove-consolelog-consolewarn-etc-from-all-appsapi-tests

stbrody: this feels like a change that could be entirely AI driven.  @andrewxhill @stephhuynh18, do we have a linear integration like the one we had on github to mark a ticket "good for AI" and get Cursor to just do it and make a PR?

apps/api/e2e/tests/chain-specific.test.ts
```diff
@@ -277,14 +277,14 @@ describe("Specific Chains", () => {
       undefined,
       undefined,
       undefined,
-      looseConstraints,
+      noTradingConstraints,
```
stbrody: can we add a comment why its ok to loosen the constraints here?

apps/api/e2e/tests/chain-specific.test.ts
stbrody: Should we just pick a different token that will pass the constraints?

apps/api/e2e/tests/multi-agent-competition.test.ts
```diff
@@ -670,18 +573,7 @@ describe("Multi-Agent Competition", () => {
...
-      const agent = agentClients[i];
-      assert(agent, "Agent is undefined");
-      const changes = portfolioChanges[agent.agent.id];
-      console.log(
-        `Agent ${i + 1} (${agent.agent.name}): $${changes?.initial.toFixed(2)} → $${changes?.final.toFixed(2)}, Change: ${changes?.percentChange.toFixed(4)}%`,
-      );
-    }
-
-    // Step 9: Verify that not all agents have exactly the same portfolio change
+    // Step 8: Verify that not all agents have exactly the same portfolio change
```
stbrody: everything from here down in this test case is just logging, there are no actual test checks.  Is there something we should actually be testing here?

apps/api/e2e/tests/multi-agent-competition.test.ts
dtbuchholz: Note: We have the tag on GH tickets to trigger AI PRa

PR #1032, opened 2025-08-05
  feat(comps): agent handle

dtbuchholz:
- Adds an agent's handle to agent creation flows, along with error handling for the new global uniqueness.
- Displays the agent's handle in various places in the UI, alongside their name: `Name (@handle)`. Note: we can clean this up later, but this'll get the basics into the app.

### DIFF
--- a/apps/comps/components/agent-profile/edit-field.tsx
+++ b/apps/comps/components/agent-profile/edit-field.tsx
@@ -18,7 +18,7 @@ interface EditAgentFieldProps {
   title: string;
   value: string;
   useTextarea?: boolean;
-  onSave: (newValue: string) => void;
+  onSave: (newValue: string) => Promise<void>;
   placeholder?: string;
   children: React.ReactNode;
 }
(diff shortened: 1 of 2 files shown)


apps/comps/components/agent-profile/edit-field.tsx
```diff
@@ -35,10 +35,21 @@ export const EditAgentField: React.FC<EditAgentFieldProps> = ({
...
+  // Update localValue when the parent's value prop changes
+  useEffect(() => {
+    setLocalValue(value);
+  }, [value]);
+
   const handleSave = async () => {
-    onSave(inputValue);
-    setLocalValue(inputValue);
-    setDialogOpen(false);
+    try {
```
dtbuchholz, line 44: i noticed that when you edit an exiting agent's handle, if the value you input was invalid, you had to refresh the page to properly try with a new value.

apps/comps/components/agent-profile/index.tsx
```diff
@@ -84,43 +85,66 @@ export default function AgentProfile({
   }, [sortState]);
 
   const handleSaveChange =
-    (field: "imageUrl" | "description" | "name" | "skills") =>
+    (field: "imageUrl" | "description" | "name" | "handle" | "skills") =>
     async (value: unknown) => {
       if (!agent) return;
 
+      if (field === "handle") {
```
dtbuchholz, line 92: parse agent handle before attempting API request

apps/comps/components/agent-profile/index.tsx
```diff
@@ -84,43 +85,66 @@ export default function AgentProfile({
...
-            toast.error("Failed to update agent name in sandbox environment", {
-              description: "The name was not updated. Please try again.",
-            });
-            return;
+            throw sandboxError;
           }
         }
 
-        // Update in main environment
-        await updateAgent.mutateAsync({
```
dtbuchholz, line 114: this was incorrectly _after_ the sandbox API call; it should come before

PR #1038, opened 2025-08-06
feat(comps): agent handle tweaks

dtbuchholz:
Remove the `@handle` from most places that _aren't_ the actual agent card, and only display the name


joewagner (approved):
lgtm!

PR #1039, opened 2025-08-06
fix: Actually assert test conditions instead of logging in multi-chain-provider.test.ts

stbrody:


apps/api/e2e/tests/price/multi-chain-provider.test.ts
```diff
@@ -600,114 +386,38 @@ describe("Multi-Chain Provider Tests", () => {
...
-      } catch (error) {
-        // Error is expected
-        console.log(
-          `Expected error for invalid token: ${error instanceof Error ? error.message : "Unknown error"}`,
-        );
-        expect(error).toBeDefined();
-      }
-    });
-
-    it("should fetch prices faster when providing the exact chain parameter", async () => {
```
stbrody, line 655: this is just a bad test that is fundamentally flaky

PR #1041, opened 2025-08-07
Rafael/api cache

### LINKED ISSUE #983: [FEATURE] Cache comps app frontend API calls
**Problem Statement**
The frontend makes too many calls to the backend APIs, e.g., in hooks. We should cache calls on the frontend and reuse them across components.

**Proposed Solution**
- We already use a caching package called `tanstack` for caching _some_ data and state management. 
- We should implement this everywhere else to improve

**Impact**
Performance improvement.

**Priority**
Medium

Rcontre360:
closes #983


Rcontre360:
also checking on other endpoints. Working first on those called the most

PR #1042, opened 2025-08-07
feat(api): enable updating trading constraints + some cleanup and better storage

### LINKED ISSUE #969: [BACKEND] Follow ups for `trading_constraints` table & liquidity rules
- [X] Add the `tradingConstraints` to the `/competitions/rules` (or the `/competitions/:id`) endpoint response so we can see what the constraints are
- [X] *Always* store the data in `trading_constraints`; we shouldn't rely on runtime defaults.
- [X] We'll need a script to update the previous comps and default to zero for all `trading_constraints`—except for the one 1 day hack comp that we're running July 23rd. This one should have the defaults stored in the table.
- [X] Ensure the admin create, update, and start endpoints account for `tradingConstraints` as a body param. E.g., we need to update the "always on" sandbox comp to use these new rules, but we have no way to do so.

joewagner:
fixes #969

### DIFF
--- a/.cursor/rules/api-specific-config.mdc
+++ b/.cursor/rules/api-specific-config.mdc
@@ -1,5 +1,5 @@
 ---
-description: 
+description:
 globs: apps/api/**/*
 alwaysApply: false
 ---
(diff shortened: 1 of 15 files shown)


joewagner:
updating contraints already made it to main in a different PR.  Going to reopen with different branch.

PR #1043, opened 2025-08-07
Modify release workflow to trigger deploys to staging

sam701:
- Trigger update-image-workflow on every push to `main`
- This change affects only staging!

### DIFF
--- a/.github/actions/trigger-deployment/action.yml
+++ b/.github/actions/trigger-deployment/action.yml
@@ -13,7 +13,7 @@ inputs:
   app-id:
     description: 'GitHub App ID'
     required: true
-  private-key:
+  app-private-key:
     description: 'GitHub App private key'
     required: true
 
@@ -25,7 +25,7 @@ runs:
       uses: actions/create-github-app-token@v1
       with:
         app-id: ${{ inputs.app-id }}
-        private-key: ${{ inputs.private-key }}
+        private-key: ${{ inputs.app-private-key }}
         owner: ${{ github.repository_owner }}
         repositories: competition-app-deployments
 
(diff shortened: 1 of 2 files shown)


dtbuchholz (approved):

dtbuchholz:
note: i'll need to make some changes to staging tomorrow that'll require a db migration. i can test this out as part of that process.

so, i'll have to:
1. do the typical proces where i ssh into the VM, update `js-recall`, and manually run the migration
2. instead of updating the docker image etc, i should just run this GH action?

sam701:
@dtbuchholz the GH workflow is configured to run on every push to `main`. You will probably need the final image to run the DB migration. That means you need to merge your PR with the DB migration code before triggering this workflow. I can wait with merging of this PR until the DB migration is done. Then after merge we should see that the new image is deployed.

dtbuchholz:
Totally, it'd be great to decouple the process! Agreed re: waiting to merge this. 

I suppose the only thing left that we "need" is: when code is merged to main, this action just needs to be able to run the [`db:migrate`](https://github.com/recallnet/js-recall/blob/b95aa7ce820503579fd3f93740589dedf8a2b376/apps/api/package.json#L24) command. I'm guessing there's a docker-friendly way to do set this up, but it'd require downtime, I think.

asutula:
Ok, so I just need some more context here. Is this PR about deploying the trading api app whenever we merge to main? I guess this relies on merging #1063 first?

sam701:
> Ok, so I just need some more context here. Is this PR about deploying the trading api app whenever we merge to main? I guess this relies on merging #1063 first?

@asutula exactly, merging to main would trigger a workflow to deploy to staging.

asutula (approved):

PR #1044, opened 2025-08-07
feat(api): contraints are returns via endpoints and always stored in db 

### LINKED ISSUE #969: [BACKEND] Follow ups for `trading_constraints` table & liquidity rules

joewagner:
fixes #969 

Work also described here: https://linear.app/recall-labs/issue/REC-44/backend-follow-ups-for-trading-constraints-table-and-liquidity-rules

### DIFF
--- a/apps/api/e2e/tests/trading.test.ts
+++ b/apps/api/e2e/tests/trading.test.ts
@@ -624,7 +624,7 @@ describe("Trading API", () => {
       undefined,
       undefined,
       undefined,
-      looseTradingConstraints,
+      { ...looseTradingConstraints, minimumLiquidityUsd: 500 },
     );
 
     // Wait for balances to be properly initialized
--- a/apps/api/e2e/utils/api-types.ts
+++ b/apps/api/e2e/utils/api-types.ts
@@ -364,12 +364,7 @@ export interface Competition {
   // Join date constraint fields
   joinStartDate: string | null;
   joinEndDate: string | null;
-  tradingConstraints?: {
-    minimumPairAgeHours?: number;
-    minimum24hVolumeUsd?: number;
-    minimumLiquidityUsd?: number;
-    minimumFdvUsd?: number;
-  };
+  tradingConstraints?: TradingConstraints;
   rewards?: {
     rank: number;
     reward: number;
@@ -802,9 +797,9 @@ export interface GlobalLeaderboardResponse extends ApiResponse {
   };
 }
 
-// ===========================
+// ---------------------------
 // Vote-related types
-// ===========================
+// ---------------------------
 
 /**
  * Response type for casting a vote
@@ -938,5 +933,3 @@ export interface AdminAddAgentToCompetitionResponse extends ApiResponse {
     status: string;
   };
 }
-
-// ===========================
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -44,6 +44,7 @@
     "end:competition": "tsx scripts/end-competition.ts",
     "comp:status": "tsx scripts/competition-status.ts",
     "scripts:backfill-agent-score": "tsx scripts/backfill-agent-score.ts",
+    "scripts:backfill-constraints": "tsx scripts/backfill-comp-constraints.ts",
     "scripts:backfill-stats": "tsx scripts/backfill-stats.ts",
     "scripts:delete-competitions": "tsx scripts/delete-competitions.ts",
     "scripts:populate-agent-handles": "tsx scripts/populate-agent-handles.ts",
--- a/apps/api/src/controllers/competition.controller.ts
+++ b/apps/api/src/controllers/competition.controller.ts
@@ -344,6 +344,12 @@ export function makeCompetitionController(services: ServiceRegistry) {
           }
         }
 
+        // Get trading constraints for the active competition
+        const tradingConstraints =
+          await services.tradingConstraintsService.getConstraintsWithDefaults(
+            activeCompetition.id,
+          );
+
         // Define base rules
         const tradingRules = [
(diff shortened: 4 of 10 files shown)


apps/api/e2e/tests/trading.test.ts
```diff
@@ -624,7 +624,7 @@ describe("Trading API", () => {
       undefined,
       undefined,
       undefined,
-      looseTradingConstraints,
+      { ...looseTradingConstraints, minimumLiquidityUsd: 500 },
```
joewagner, line 627: had to add this to get tests passing.

apps/api/e2e/utils/api-types.ts
```diff
@@ -364,12 +364,7 @@ export interface Competition {
   // Join date constraint fields
   joinStartDate: string | null;
   joinEndDate: string | null;
-  tradingConstraints?: {
-    minimumPairAgeHours?: number;
-    minimum24hVolumeUsd?: number;
-    minimumLiquidityUsd?: number;
-    minimumFdvUsd?: number;
-  };
+  tradingConstraints?: TradingConstraints;
```
joewagner, line 367: de-duplicating the type here

apps/api/e2e/utils/api-types.ts
```diff
@@ -802,9 +797,9 @@ export interface GlobalLeaderboardResponse extends ApiResponse {
   };
 }
 
-// ===========================
+// ---------------------------
```
joewagner, line 800: The `======` character set is indicative of a rebase or merge conflict.  It's easier for me to handle them if comments don't use it as a divider.  Not a big deal, so I can undo if anyone has strong preference.

apps/api/package.json
```diff
@@ -44,6 +44,7 @@
     "end:competition": "tsx scripts/end-competition.ts",
     "comp:status": "tsx scripts/competition-status.ts",
     "scripts:backfill-agent-score": "tsx scripts/backfill-agent-score.ts",
+    "scripts:backfill-constraints": "tsx scripts/backfill-comp-constraints.ts",
```
joewagner, line 47: This is the backfill script.  I tested it on data I got from production on Aug. 6th, and it works as expected.

apps/api/src/controllers/competition.controller.ts
```diff
@@ -624,17 +635,6 @@ export function makeCompetitionController(services: ServiceRegistry) {
           ]).size,
         };
 
-        // Get trading constraints and rewards for this competition
-        const tradingConstraintsRaw =
-          await services.tradingConstraintsService.getConstraints(
-            competitionId,
-          );
-        const tradingConstraints = {
```
joewagner, line 632: Using the service method below instead of this to dry out the code.

apps/api/src/services/competition-manager.service.ts
```diff
@@ -165,16 +165,6 @@ export class CompetitionManager {
 
     await createCompetition(competition);
 
-    // Create trading constraints if provided
-    if (tradingConstraints) {
```
joewagner, line 169: We are going to always create the constraints now, instead of relying on defaults from the config during get requests.

apps/api/e2e/tests/trading.test.ts
dtbuchholz: note: i recently changed this because the previous token pair was also failing. i chose [WETH](https://dexscreener.com/solana/7vfCXTUXx5WJV5JADk17DUJ4ksgau7utNKj4b963voxs) because i figured that'd be safe enough....but i guess not

Image

apps/api/src/services/competition-manager.service.ts
```diff
@@ -188,6 +178,15 @@ export class CompetitionManager {
       );
     }
 
+    // Always create trading constraints (with defaults if not provided)
+    await this.tradingConstraintsService.createConstraints({
+      competitionId: id,
+      ...tradingConstraints,
```
dtbuchholz, line 184: this doesnt create with the defaults, though, right? it uses the param passed by the method, which would come from the admin API. 

i.e., either in this method or the [admin method](https://github.com/recallnet/js-recall/blob/9722600dcd55a6391f8a428a5e11e141f084d3bd/apps/api/src/controllers/admin.controller.ts#L598), we should check if `tradingConstraints` are provided. and if not, we would get defaults from the config and then use those as the values that are written to the table.

dtbuchholz: edit: after doing some local testing, i think my comment above is incorrect.

but, i think we should include the `tradingConstraints` in the response for `startCompetition` and `createCompetition`. that'll help confirm it's actually being set.

joewagner: > edit: after doing some local testing, i think my comment above is incorrect.

That's correct.  If you look here, you'll see this takes whats passed in and fills any missing constraints with the values in config
https://github.com/recallnet/js-recall/blob/main/apps/api/src/services/trading-constraints.service.ts#L32

> but, i think we should include the tradingConstraints in the response for startCompetition and createCompetition. that'll help confirm it's actually being set.

Good call, I'll update that now

PR #1045, opened 2025-08-07
chore(api-sdk): regen speakeasy sdk

dtbuchholz:


.github/actions/api-sdk-check/action.yml
```diff
@@ -20,7 +20,7 @@ runs:
     - name: Generate API SDK
       shell: bash
       run: |
-        sh scripts/build-api-sdk.sh
+        bash scripts/build-api-sdk.sh
```
dtbuchholz: to fix this https://github.com/recallnet/js-recall/actions/runs/16812830870/job/47622181916#step:9:83

(or, we can also just remove this from CI...)

dtbuchholz: edit: i removed this. it's more of a nuisance than anything else, and if we want to keep it, we can just add a manual "publish" action...or just make manual updates.

joewagner (approved):
gtg

PR #1046, opened 2025-08-07
feat(comps): frontend cleanups

dtbuchholz:
- Fixed the comp rewards card not having a space between a word.
- Fixed colors to make sure trophies/ranks are consistent, and green / red / yellow use the figma colors.
- Made all of the agent rows in the global leaderboard and per-comp results clickable, vs a dedicated column for the agent profile.
- Global leaderboard tweaks to account for better spacing and responsiveness (i.e., hide the final 2 cols on smaller viewports).
- Fixed "jumpy" nav that shifted text upward upon clicking.
- Footer section updates.
- Fixed "add agent" button to prompt for wallet connection or agent creation, depending on user state.
- Fixed buttons being unable to have the "disable" cursor upon hover.


joewagner (approved):
Looks good

Issue #1049, opened 2025-08-08
[IMPROVEMENT] getCompetitions database impact is too high.


joewagner:
I noticed the `getCompetitions` method (currently here: https://github.com/recallnet/js-recall/blob/main/apps/api/src/controllers/competition.controller.ts#L452) will potentially do 22 db queries per request. Once we have 100+ competitions, it will max out doing 502 db queries per request.  Might be a while until we get to that point, but it seems good to capture work to fix this.

PR #1050, opened 2025-08-08
chore(api): reduce 4N query to 3

joewagner:
fixes: #1049

### DIFF
--- a/apps/api/src/controllers/competition.controller.ts
+++ b/apps/api/src/controllers/competition.controller.ts
@@ -1,6 +1,10 @@
 import { NextFunction, Response } from "express";
 
 import { config } from "@/config/index.js";
+import {
+  getBatchVoteCounts,
+  getEnrichedCompetitions,
+} from "@/database/repositories/competition-repository.js";
 import { SelectCompetitionReward } from "@/database/schema/core/types.js";
 import { competitionLogger } from "@/lib/logger.js";
 import { ApiError } from "@/middleware/errorHandler.js";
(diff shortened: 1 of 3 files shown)


apps/api/src/controllers/competition.controller.ts
```diff
@@ -479,69 +483,66 @@ export function makeCompetitionController(services: ServiceRegistry) {
...
             status,
+            // Default limit 10, max 100. It's important we don't call this without a limit
             pagingParams,
           );
 
         // If user is authenticated, enrich competitions with voting information
         let enrichedCompetitions = competitions;
         if (userId) {
-          enrichedCompetitions = await Promise.all(
-            competitions.map(async (competition) => {
```
joewagner, line 490: This `await Promise.all( ... array.map(...` with database calls inside it is one of the obvious patterns we can search for to find places the db will get overloaded when data gets bigger.  This PR removes this and simply does two queries that involve joins.  reducing the max query count for this controller from 402 down to always being 3.

apps/api/src/controllers/competition.controller.ts
```diff
@@ -479,69 +483,66 @@ export function makeCompetitionController(services: ServiceRegistry) {
...
-                  votingEnabled: false,
-                  totalVotes: 0,
-                };
-              }
-            }),
+          const competitionIds = competitions.map((c) => c.id);
+
+          // Fetch all data in parallel with batch queries (2 queries total instead of 4N)
+          const [enrichmentData, voteCountsMap] = await Promise.all([
+            getEnrichedCompetitions(userId, competitionIds),
```
joewagner, line 498: `Promise.all` is just two methods, and they both only do one query no matter what.

apps/api/src/controllers/competition.controller.ts
```diff
@@ -479,69 +483,66 @@ export function makeCompetitionController(services: ServiceRegistry) {
...
+
+          enrichedCompetitions = competitions.map((competition) => {
+            const enrichment = enrichmentMap.get(competition.id);
+            if (!enrichment) {
+              throw new ApiError(500, "invalid competition state");
+            }
+
+            const hasVoted = !!enrichment.userVoteAgentId;
+            const compVotingStatus =
+              services.voteManager.checkCompetitionVotingEligibility(
```
joewagner, line 515: I added this method to reduce controller logic, and allow getting voting status in one call.

apps/api/src/services/vote-manager.service.ts
```diff
@@ -369,4 +369,33 @@ export class VoteManager {
 
     return { canVote: true, reason: "voting is open" };
   }
+
+  /**
+   * Check overall competition voting eligibility combining status and date checks
+   * @param competition The competition record
+   * @returns Object with canVote boolean and reason string
+   */
+  checkCompetitionVotingEligibility(competition: SelectCompetition): {
```
joewagner, line 378: Here's the helper method mentioned above.

dtbuchholz (approved):

Issue #1053, opened 2025-08-08
[PARENT] Need to fix all places where an increasing size of a query response results in an increasing number of additional db queries


joewagner:
Here is a list of the places I've found (so far) where we allow the size of a dataset to increase the load on the database:

1. [x] **agent.controller.ts** (line \~240):
   * `Promise.all(balances.map(async (balance) => { ... await getLatestPrice(...) ... }))`
   * Queries price for each balance
2. [x] **agentscore-repository.ts** (line \~177):
   * `Promise.all(rankDataArray.map(async (rankData) => { ... await tx.insert(...) ... }))`
   * Performs database insert/update for each rank data
3. [x] **agent-manager.service.ts** (line \~977):
   * `Promise.all(validCompetitions.map(async (competition) => await this.attachCompetitionMetrics(competition, agentId)))`
   * Attaches metrics for each competition
4. [ ] **agent-manager.service.ts** (line \~1520):
   * `Promise.all(sanitizedAgents.map((agent) => this.attachAgentMetrics(agent)))`
   * Fallback that attaches metrics for each agent when bulk operation fails
5. [x] **portfolio-snapshotter.service.ts** (line \~269):
   * `snapshots.map(async (snapshot) => { ... await getPortfolioTokenValues(snapshot.id) ... })`
   * Gets token values for each snapshot
6. [x] **competition-manager.service.ts** (line \~381):
   * Sequential for loop calling `getAgentPerformanceForComp` for each agent in the leaderboard
   * Not using Promise.all but still an N+1 pattern

These are some instances where the number of database queries increases linearly with the size of the result array from an earlier query, which can lead to performance issues as the data grows.

Note on how I found these: 
I tried asking Claude and gpt to find this kind of thing, but they just find comments left from previous PRs that say the issue is fixed and claim nothing like this exists.  After that I spent a little time looking manually and found one instance, which I then showed to Claude.  After doing this it found these other places.   
[#1050](https://github.com/recallnet/js-recall/issues/1050) fixes the initial instance I found.  It took a little bit of convincing to get the llm to fix it correctly, I have a suspicion that since this pattern exists in the codebase it's being duplicated by Claude/GPT etc...  I think it would also make sense to remove and avoid comments that say things like "This replaces the N+1 query problem from the old calculateGlobalMetrics method", since it's not really relevant to what the code is doing and it seems to confuse the llms when they scan the code.

PR #1054, opened 2025-08-08
chore(api): reduce 3N queries to constant

joewagner:
part of #1053

### DIFF
--- a/apps/api/src/controllers/agent.controller.ts
+++ b/apps/api/src/controllers/agent.controller.ts
@@ -5,7 +5,6 @@ import {
   getAgentPortfolioSnapshots,
   getPortfolioTokenValues,
 } from "@/database/repositories/competition-repository.js";
-import { getLatestPrice } from "@/database/repositories/price-repository.js";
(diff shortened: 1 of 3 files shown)


dtbuchholz (approved):
@joewagner lgtm.

random observations: it looks like the `getCompetitions` method ultimately calls [`findByStatusImpl`](https://github.com/recallnet/js-recall/blob/15223ffc3d60e40d8199014fed36d8c03639154f/apps/api/src/database/repositories/competition-repository.ts#L1183), and our zod defaults don't apply any default sorting—so i think PG will return whatever it wants. i.e., i think we want to apply default sorting on the `/api/competitions`, `/api/user/competitions`, and `/api/agent/competitions` query so that we return the newest comps first. also, the comps app doesn't account for paginated competitions, so this will also need to be addressed.

both of these aren't needed now since we have fewer than 10 comps, but i've created a Linear issue to address this sometime later: https://linear.app/recall-labs/issue/REC-363/apply-default-sorting-to-competitions-endpoints

PR #1055, opened 2025-08-08
chore(api): refactor mutli query tx into single query

joewagner:
part of #1053

### DIFF
--- a/apps/api/src/controllers/agent.controller.ts
+++ b/apps/api/src/controllers/agent.controller.ts
@@ -231,7 +231,7 @@ export function makeAgentController(services: ServiceRegistry) {
       try {
         const agentId = req.agentId as string;
 
-        // Get the balances, this could be hundereds
+        // Get the balances, this could be hundreds
         const balances = await services.balanceManager.getAllBalances(agentId);
 
         // Extract all unique token addresses
--- a/apps/api/src/database/repositories/agentscore-repository.ts
+++ b/apps/api/src/database/repositories/agentscore-repository.ts
@@ -10,10 +10,13 @@ import {
 import {
   InsertAgentScore,
   InsertAgentScoreHistory,
+  SelectAgentScore,
 } from "@/database/schema/ranking/types.js";
 import { repositoryLogger } from "@/lib/logger.js";
 import { createTimedRepositoryFunction } from "@/lib/repository-timing.js";
 import { AgentMetadata } from "@/types/index.js";
+
+type DatabaseTransaction = Parameters<Parameters<typeof db.transaction>[0]>[0];
 
 /**
  * Agent Rank Repository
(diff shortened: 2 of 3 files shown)


apps/api/src/controllers/agent.controller.ts
```diff
@@ -233,74 +231,39 @@ export function makeAgentController(services: ServiceRegistry) {
       try {
         const agentId = req.agentId as string;
 
-        // Get the balances
+        // Get the balances, this could be hundereds
         const balances = await services.balanceManager.getAllBalances(agentId);
 
-        // Enhance balances with chain information
-        const enhancedBalances = await Promise.all(
-          balances.map(async (balance) => {
```
joewagner: This is the pattern we need to remove from the codebase.  This currenlty amounts to ~40 queries for a normal comp, but as things get more complex and traders use more tokens, this could blowup.

apps/api/src/database/repositories/agentscore-repository.ts
```diff
@@ -175,32 +178,8 @@ async function batchUpdateAgentRanksImpl(
         }),
       );
 
-      // Batch update agent ranks
-      const results = await Promise.all(
-        rankDataArray.map(async (rankData) => {
```
joewagner, line 180: The main reason for the PR is removing the `Promise.all(array.map(async` call here.

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -1976,7 +1976,12 @@ async function getBatchVoteCountsImpl(
           totalVotes: 0,
         });
       }
-      const competition = competitionVoteMap.get(competitionId)!;
+      const competition = competitionVoteMap.get(competitionId);
+      if (!competition) {
```
joewagner, line 1980: The changes in this file are formatting, linting, and removing the use of the typescript `!` operator.

apps/api/src/database/repositories/agentscore-repository.ts
```diff
@@ -244,6 +223,57 @@ async function getAllAgentRankHistoryImpl(competitionId?: string) {
...
+  rankDataArray: InsertAgentScore[],
+): Promise<SelectAgentScore[]> {
+  if (rankDataArray.length === 0) {
+    return [];
+  }
+
+  // Build the VALUES part dynamically
+  const sqlChunks: ReturnType<typeof sql>[] = [];
+
+  sqlChunks.push(sql`
```
stbrody, line 240: I'm a little bit worried about how often we've been reaching for raw SQL lately instead of actually using the ORM.  We really need to be careful about potential for SQL injection

joewagner: This is a really good point.  For what it is worth I did think (a little bit ) about this here.  The `sql` template literal should be escaping everything.  Here are the docs: https://orm.drizzle.team/docs/sql

PR #1062, opened 2025-08-12
Comp detail actions change

### LINKED ISSUE #932: [UI/UX] Make “Vote Now” the primary CTA; add secondary “Add Agent” + conditional “PnL Graph” buttons
Image

Image

**Problem Statement**
User testing shows that the Join Competition CTA competes with the intended main action—voting. We must switch hierarchy so that users are first guided to vote on existing agents, while still enabling agent submission (join flow) and PnL exploration when available.

**Proposed Solution**
	1.	Primary CTA
	•	Button copy: “Vote”
	•	Style: primary button (brand highlight)
	•	Action: smooth-scroll to the Local Leaderboard table section of the same page.
	2.	Secondary CTAs
	•	“Compete” — launches current Join-Flow modal.
	•	“PnL” — links to external microsite only if competition.hasPnlGraph === true; otherwise hide entirely.
	3.	State Handling
	•	If user.hasVotedInCompetition === true, render Vote Now in a disabled state (40 % opacity, cursor: not-allowed, accessible aria-disabled="true").
	•	Tooltip on disabled state: “You’ve already voted in this competition.”
	4.	Layout
	•	Desktop: buttons aligned right, ordered Vote Now → Add Agent → PnL Graph.
	•	Tablet / Mobile: full-width stacked (primary first, then secondaries).

**Acceptance Criteria**
AC-1
Given a competition with a PnL graph, when the detail page loads, then I see three buttons in order: Vote (primary style), Compete, PnL.
AC-2
Given the competition lacks a PnL graph (hasPnlGraph=false), then I only see Vote and Compete.
AC-3
When I click Vote, then the page scrolls to the Agents table (no full reload).
AC-4
Given I have already cast a vote, then Vote appears disabled with tooltip copy “You’ve already voted in this competition.”
AC-5
When I click Compete, then the existing Join-Flow modal opens with no regression.
AC-6
When I click PnL, then I’m taken to /pnl/{competitionSlug} in a new tab.
AC-7
All buttons retain keyboard focusability & accessible labels.

**Out of Scope**
	•	Backend voting logic (unchanged).
	•	Modifying the Join-Flow itself.

[Figma Link](https://www.figma.com/design/Gg5j6btgKAbWJwSBcIujVI/Competitions-App---v.1.1--07-25-?node-id=3323-36650)

Rcontre360:
closes #932


ramirotw (changes requested):
1. Vote button is always disabled
2. Vote button is hidden if the user already voted
3. Buttons should be re-arrange to: Vote, Compete and PnL/Performance/Chart
4. PnL/Performance/Chart button should scroll down to the chart on the current page

Rcontre360:
> 1. Vote button is always disabled
> 2. Vote button is hidden if the user already voted
> 3. Buttons should be re-arranged to: Vote, Compete and PnL/Performance/Chart
> 4. PnL/Performance/Chart button should scroll down to the chart on the current page

Related to the point 4. There's a new page in route /competitions/:id/chart. Shouldn't the PnL button take you there @dtbuchholz ?

ramirotw (changes requested):
5. The PNL button looks different compared to the COMPETE button, style is off. Also the icon should be changed to a chevron-right or something similar as it's not an external link.

PR #1063, opened 2025-08-12
feat: automate sql schema migration

sam701:
- Run DB migration on every app start.
- Use postgres advisory lock to prevent running DB migration concurrently by different app instances.
- Log SQL statements during DB migration

Tested with 10 concurrent apps.


apps/api/src/database/db.ts
```diff
@@ -318,11 +319,55 @@ export async function dropAll() {
...
+ */
 export async function migrateDb() {
-  // Run normal Drizzle migrations
-  await migrate(db, {
-    migrationsFolder: path.join(__dirname, "../../drizzle"),
-  });
+  const MIGRATION_LOCK_ID = 77; // Arbitrary but consistent lock ID
+  const MAX_WAIT_TIME = "5min";
+
+  try {
```
dtbuchholz, line 330: i'm not an expert here, so i had gpt-5 provide an analysis. take it with a grain of salt. i think point (1) needs to be addressed. for the others, i figured i'd include them but idk what the best practices are here. maybe @joewagner or @stbrody have an opinion.

1. lock_timeout is set on the wrong connection and never reset
> You call await `db.execute("SET lock_timeout = '5min'")` on the shared db and then run migrations on migrationDb (a different connection). That means the timeout does not protect the DDL in your migration session, but it does persist on the pooled connection used for the SET, potentially leaking a 5‑minute lock wait into normal app traffic after startup (session GUCs stick to that session). 
> 
> Recommended: set timeouts with `SET LOCAL` inside a transaction on the same connection used for migrations, or `RESET lock_timeout` afterward.
> 
> ```ts
> await migrationDb.execute(sql.raw("BEGIN"));
> await migrationDb.execute(sql.raw("SET LOCAL lock_timeout = '5min'"));
> await migrationDb.execute(sql.raw("SET LOCAL statement_timeout = '10min'")); // optional guard
> await migrate(migrationDb, { migrationsFolder: ... });
> await migrationDb.execute(sql.raw("COMMIT"));
> 
> // …and do not set lock_timeout on db at all.
> ```

2. Advisory lock usage blocks indefinitely under some failures
> `pg_advisory_lock()` blocks until it can acquire the lock. If one runner dies while holding a DB session open (e.g., stuck connection), other runners can hang for a long time. You tried to mitigate with lock_timeout, but that governs waiting for locks on objects/statements; advisory locks are special and best handled either with `pg_try_advisory_lock` + retry/timeout, or a lock_timeout/statement_timeout on the same session actually acquiring the advisory lock. Many teams prefer optimistic `pg_try_advisory_lock(...)` and bail/backoff if unavailable.  ￼ ￼
>
> ```ts
> const start = Date.now();
> const maxMs = 5*60*1000;
> while (true) {
>   const { rows: [{ pg_try_advisory_lock: got }] } =
>     await db.execute(sql.raw(`SELECT pg_try_advisory_lock(${MIGRATION_LOCK_ID})`));
>   if (got) break;
>   if (Date.now() - start > maxMs) throw new Error("migration lock timeout");
>   await new Promise(r => setTimeout(r, 1000));
> }
> ```

3. `MIGRATION_LOCK_ID = 77` is a collision risk
> Advisory lock keys are just integers. If multiple apps share the same database, a low constant like 77 can collide. Consider deriving a namespaced key (e.g., 64‑bit hash of app name) or use the two‑int form `pg_advisory_lock(key1,key2)` with an app‑specific key1. 
> ```sql
> pg_try_advisory_lock( hashtext('js-recall:migration')::int, 0 )
> ```

4. Running migrations unconditionally in production is risky
> Long/locking DDL (e.g., `ALTER TABLE ...`) can block writers/readers and cause downtime if it grabs strong locks; at scale, many teams gate migrations behind a flag or run them in a CI/CD step or an initContainer/one‑shot Job, not on every app instance. Even with a global lock, the app process is the wrong place to decide when to perform breaking DDL. Consider: `RUN_MIGRATIONS_ON_STARTUP=true` opt‑in and keep default off for prod. Guidance on using lock/statement timeouts to avoid lock pile‑ups during DDL.

5. Loss of DB readiness/backfill logic in Docker
> Removing `startup.sh` drops `pg_isready` waiting and the `RUN_BACKFILL` path (baseline + migrations). Now the container will try to start immediately and fail fast if the DB isn’t ready; in compose or Kubernetes this may cause crash loops. If you keep “migrate on start”, consider re‑introducing a small DB readiness wait or rely on orchestrator health checks/initContainers.

6. Startup failure mode = `process exit`
> In `index.ts`, a migration error logs then process.exit(1). That’s defensible, but combined with (4)/(5) can create repeated restarts during a long migration window. If you stick with this model, add clear logs & metrics and consider surfacing a /healthz that reports “starting/migrating” vs “ready”.

7. Migration query logging might be too chatty
> The new logger logs the first 200 chars of every migration query. That’s generally fine, but watch for sensitive literals in data migrations (rare, but possible). You might want to log DDL only or redact literals.

8) `knownOperations` includes "SET" (OK) but check other SQL emitted by Drizzle
> Adding "SET" to the allowlist fixes the lock‑timeout log line classification. Just ensure other control statements Drizzle might emit during migrations (e.g., BEGIN, COMMIT) aren’t accidentally filtered—review your logger’s behavior around transaction control.

---

examples:

<details>
<summary>apps/api/src/database/migrate.ts</summary>

```ts
// apps/api/src/database/migrate.ts
import { Client } from "pg";
import { drizzle } from "drizzle-orm/node-postgres";
import { migrate } from "drizzle-orm/node-postgres/migrator";

// Tune these as you like:
const LOCK_KEY_1 = 0x6a73;      // namespace for "js"
const LOCK_KEY_2 = 1063;        // this PR id (any app-specific small int)
const TRY_LOCK_MAX_MS = 5 * 60 * 1000; // 5 minutes
const TRY_LOCK_SLEEP_MS = 1000;        // 1s backoff
const LOCKED_LOG_EVERY_MS = 15 * 1000; // log while waiting

export type MigrateOptions = {
  connectionString: string;
  migrationsFolder: string;       // e.g. "drizzle"
  lockTimeout?: string;           // e.g. "5min"
  statementTimeout?: string;      // e.g. "10min"
  logger?: (msg: string) => void; // optional logging hook
};

export async function migrateDb(opts: MigrateOptions) {
  const {
    connectionString,
    migrationsFolder,
    lockTimeout = "5min",
    statementTimeout = "10min",
    logger = console.log,
  } = opts;

  const client = new Client({ connectionString });
  await client.connect();

  const log = (m: string) => logger?.(`[migrate] ${m}`);

  // 1) acquire advisory lock (non-blocking w/ backoff)
  const start = Date.now();
  let lastLogged = 0;

  try {
    for (;;) {
      const { rows: [{ pg_try_advisory_lock: got }] } = await client.query<{
        pg_try_advisory_lock: boolean;
      }>("select pg_try_advisory_lock($1::int, $2::int)", [LOCK_KEY_1, LOCK_KEY_2]);

      if (got) {
        log(`acquired advisory lock (${LOCK_KEY_1}, ${LOCK_KEY_2})`);
        break;
      }

      const waited = Date.now() - start;
      if (waited - lastLogged >= LOCKED_LOG_EVERY_MS) {
        log(`waiting for advisory lock... (${Math.round(waited / 1000)}s)`);
        lastLogged = waited;
      }

      if (waited >= TRY_LOCK_MAX_MS) {
        throw new Error("migration lock timeout");
      }
      await new Promise((r) => setTimeout(r, TRY_LOCK_SLEEP_MS));
    }

    // 2) run migrations within an explicit transaction + SET LOCAL timeouts
    // (SET LOCAL applies only for this tx and won’t leak to other sessions)
    await client.query("BEGIN");
    await client.query(`SET LOCAL lock_timeout = $1`, [lockTimeout]);
    await client.query(`SET LOCAL statement_timeout = $1`, [statementTimeout]);

    log(`running drizzle migrations from "${migrationsFolder}"...`);
    // Use a dedicated client for the migrator.
    await migrate(drizzle(client), { migrationsFolder });
    await client.query("COMMIT");
    log("migrations complete ✅");
  } catch (err) {
    // If something blew up during migration, try to roll back the tx.
    try { await client.query("ROLLBACK"); } catch {}
    throw err;
  } finally {
    // 3) release the advisory lock if we still hold it, then disconnect
    try {
      await client.query("select pg_advisory_unlock($1::int, $2::int)", [LOCK_KEY_1, LOCK_KEY_2]);
      log(`released advisory lock (${LOCK_KEY_1}, ${LOCK_KEY_2})`);
    } catch {}
    await client.end();
  }
}
```

</details>

<details>
<summary>apps/api/src/index.ts</summary>

```ts
import { migrateDb } from "./database/migrate";
// ... your existing imports

async function bootstrap() {
  // Optionally run migrations on startup (default off)
  if (process.env.MIGRATE_ON_STARTUP === "true") {
    await migrateDb({
      connectionString: process.env.DATABASE_URL!,
      migrationsFolder: "drizzle",
      lockTimeout: process.env.MIGRATION_LOCK_TIMEOUT ?? "5min",
      statementTimeout: process.env.MIGRATION_STATEMENT_TIMEOUT ?? "10min",
      logger: (m) => console.info(m),
    });
  }

  // then start Express, etc.
  // app.listen(...)
}

bootstrap().catch((e) => {
  console.error("fatal during bootstrap:", e);
  process.exit(1);
});
```

</details>

sam701: @dtbuchholz thanks for running this through GPT-5! Some quick thoughts on its feedback:

1.  `SET` is session-level and we're using the same db connection for both the `lock_timeout` change and `pg_advisory_lock`, so we're good. `SET LOCAL` does not have any effect (only a warning) if called outside a transaction.
2. Not correct - advisory locks auto-release when the connection drops.
3. We have a single app using the DB. All locks can be viewed with `select * from pg_locks`. I find a "simple" number more clear. No strong opinion though.
4. True, migrations are inherently risky, especially zero-downtime ones. But that's exactly what we're trying to achieve here. I'd say we just need to stay sharp and make sure we understand what each migration does before running it.
5. If the DB is down, the app won't start and we'll get alerted - seems fine to me.
6. Could be worth an issue to explore, but doesn't seem urgent.
7. Yes, the idea is to have all migrations logged.
8. Not clear what GPT-5 meant here.

joewagner:
I think Alexei is probably the most knowledgeable among us, but I've got a question on how this affects the deployment process.
If I understand right, advisory locks don't actually lock access to the database unless the application is setup to honor the lock. (which our api does not do, right?).  

To be clear, I'm not sure this is true, but consider if we were to have a running app during a competition and then start a second one. The running app will be trying to make changes  to the database, the startup of the second app would try to run migrations.
Is this the case, or is there a safegaurd against that?  Or maybe we just make sure never to do that?  We should never stop the app during a competition, but I can see wanting to start a second instance of the api.

sam701:
>  If I understand right, advisory locks don't actually lock access to the database unless the application is setup to honor the lock. (which our api does not do, right?).

That's correct. Advisory locks are just named mutexes. It just causes subsequent calls to `pg_advisory_lock(123)` wait until the connection holding the lock releases it. All other SQL statements are unaffected.

> The running app will be trying to make changes to the database, the startup of the second app would try to run migrations. Is this the case, or is there a safegaurd against that? 

Correct, the second app will run migrations while the first is still serving traffic. This means we need to ensure the current app version remains compatible with the new database schema.

joewagner:
> FWIW I don't think this is a safe assumption at all. I think we need to be able to restart the app, and even deploy new versions of the code, mid-competition

When I said "We should never stop the app during a competition, but I can see wanting to start a second instance of the api.", I was trying to say deploying would ideally involve starting a second instance, then stopping the first instance, as a means to deploy a new version.  Which means the "app", in the sense of what is available on the website, would never stop working from the user perspective.  

Which makes auto migrating kind of complex, but I think what we have in this PR will work as long as we don't generate migrations that require downtime during a competition.  Or if the migrations do require downtime, we have to correctly shutdown the api before the migration runs.

Does that make sense?

stbrody:
I don't think we should expect zero-downtime deploys until we have a proper HA setup with multiple app servers up all the time.  Right now the app is not well set up for multiple app servers to coexist, and doing so, even for a brief moment during an upgrade, could cause data corruption.  See https://linear.app/recall-labs/project/ha-app-deployment-53449701f869/overview for more info

sam701:
@dtbuchholz @mzkrasner, I would love to get your approval on this PR too when you get a chance! 

This automated migration stuff is a bit of a shift in how we handle database changes, so wanted to make sure you're on board. The main thing to keep in mind is that going forward, we'll need to make sure our PRs are backward and forward compatible. Not a huge deal right now since we're running a single instance and expect some downtime during deployments (like @stbrody mentioned), but it'll be pretty important once we move to HA mode.

Just want to make sure we're all good with this direction and there aren't any red flags I'm missing. Would appreciate approval from you if it looks good!

asutula:
This is looking pretty good, but question: Did you try concurrent app migrations without the postgres lock logic? Drizzle has a table in our database called `migrations` where it tracks applied migrations. I haven't verified this myself, but I'd imagine it applies the migrations and updates that table within a single transaction, and therefore concurrent attempted migrations would be handled by postgres' normal table/txn based locking mechanisms. i.e. the first app that attempts the migration would succeed, and other apps would be a no-op because there are no migrations to apply.

sam701:
@asutula Good question! I'd actually looked into this before and came across several sources saying you shouldn't run migrations concurrently, even with drizzle's transaction wrapping. The idea is that multiple apps can still start their own transactions at the same time and run into conflicts.

But since you mentioned it, I thought I'd test it out to see what actually happens. Disabled the advisory lock and spun up 30 apps concurrently - and yes, things broke depending on timing. Here's what I saw:
```
app-1   | [08:35:32 UTC] ERROR: [Database] Error during migration lock process:
app-1   | [08:35:32 UTC] ERROR: [API] Database initialization error: error: duplicate key value violates unique constraint "pg_namespace_nspname_index"
app-1   | [08:35:32 UTC] ERROR: [API] Exiting...
app-5   | [08:35:32 UTC] ERROR: [Database] Error during migration lock process:
app-5   | [08:35:32 UTC] ERROR: [API] Database initialization error: error: duplicate key value violates unique constraint "pg_namespace_nspname_index"
app-5   | [08:35:32 UTC] ERROR: [API] Exiting...
app-4   | [08:35:32 UTC] ERROR: [Database] Error during migration lock process:
app-4   | [08:35:32 UTC] ERROR: [API] Database initialization error: error: duplicate key value violates unique constraint "pg_namespace_nspname_index"
app-4   | [08:35:32 UTC] ERROR: [API] Exiting...
app-21  | [08:35:32 UTC] ERROR: [Database] Error during migration lock process:
app-21  | [08:35:32 UTC] ERROR: [API] Database initialization error: error: duplicate key value violates unique constraint "pg_namespace_nspname_index"
app-21  | [08:35:32 UTC] ERROR: [API] Exiting...
```
I think the issue with the single transaction approach is that there can be still multiple transactions in flight. Even if the DB stays in the correct state, multiple transactions will be reverted and we see some errors in logs.

asutula:
Thanks for testing that out @sam701. Your solution here in this PR is great then.

asutula (approved):

PR #1070, opened 2025-08-13
feat: Add competition participant limits (REC-426)

mzkrasner:
## Summary

Implements optional maximum participant limits for competitions, allowing admins to cap registration at a specified number. Includes comprehensive API updates, bug fixes, and full test coverage.

### Database & Schema
- Added `max_participants` INTEGER column to `competitions` table (nullable for backward compatibility)
- Updated database schema types to include `maxParticipants: number | null`
- Added `getParticipantCount()` repository function for querying active participants

### API Endpoints Updated
Updated 8 competition endpoints to include `maxParticipants` field:
- `POST /api/admin/competition/create` (request parameter)
- `POST /api/admin/competition/start` (response field)
- `POST /api/admin/competition/end` (response field)
- `GET /api/competitions/{competitionId}` (response field)
- `GET /api/competitions/{competitionId}/agents` (response field)
- `GET /api/competitions/upcoming` (response field)
- `GET /api/user/competitions` (response field)
- `GET /api/agents/{agentId}/competitions` (response field)

### Service Layer
- Added `getParticipantCount()` method to CompetitionManager service following MVC patterns
- Enhanced participant limit validation in competition registration flow
- Updated `createCompetition()` method to accept `maxParticipants` parameter

### Bug Fix
Fixed critical bug in `getCompetitionById` where `stats.totalAgents` incorrectly counted only agents with trades instead of all registered participants. This was affecting participant counts for pending competitions and competitions without active trading.

### Testing
- Added comprehensive E2E test suite covering participant limits functionality
- Tests include: competition creation with limits, registration enforcement, API response validation, edge cases
- Verified cross-client consistency (admin, agent, user) for pending competitions
- Eliminated all `any` types and added proper TypeScript imports

### API Behavior
- `maxParticipants: null` = unlimited participants (backward compatible)
- `maxParticipants: number` = enforced limit with 403 error when full
- `stats.totalAgents` now correctly counts all registered participants
- Proper HTTP status codes and error messages for limit scenarios

### Validation
- Added Zod schema validation for `maxParticipants` (minimum 1 when specified)
- Updated OpenAPI documentation across all affected endpoints
- Maintained backward compatibility - existing competitions remain unlimited

## Test Coverage
- Multi-client testing across admin, agent, and user access patterns
- Registration limit enforcement during agent registration
- API response validation for participant counts
- Edge cases including unlimited competitions and validation errors

## Ready for Frontend Integration
All backend APIs now return consistent `maxParticipants` data required for frontend participant count displays and registration flow logic.


apps/api/src/services/competition-manager.service.ts
```diff
@@ -155,6 +159,7 @@ export class CompetitionManager {
       votingEndDate: votingEndDate || null,
       joinStartDate: joinStartDate || null,
       joinEndDate: joinEndDate || null,
+      maxParticipants: maxParticipants || null,
```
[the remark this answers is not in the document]
mzkrasner: fixed - changed `maxParticipants: maxParticipants || null` to
  `maxParticipants: maxParticipants ?? null` to ensure `0` values are preserved.

  This allows setting competitions with 0 participants (effectively closed competitions) while
   still converting `undefined` to `null` for database consistency. The fix is included in the
   latest commit.

apps/api/src/controllers/admin.controller.ts
```diff
@@ -695,6 +697,7 @@ export function makeAdminController(services: ServiceRegistry) {
             votingEndDate ? new Date(votingEndDate) : undefined,
             joinStartDate ? new Date(joinStartDate) : undefined,
             joinEndDate ? new Date(joinEndDate) : undefined,
+            undefined, // maxParticipants (not available in start competition schema)
```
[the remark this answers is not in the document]
mzkrasner: You're correct that we should preserve the existing `maxParticipants`
  value. However, this only applies when starting an **existing** competition. The current
  code correctly handles both cases:

  - When starting an existing competition, the `startCompetition` service method fetches the
  competition from the database and preserves all existing properties including
  `maxParticipants`
  - When creating a new competition during the start process, `undefined` is appropriate since
   no participant limit was specified in the start request

  The service layer architecture ensures participant limits are preserved correctly without
  needing controller-level intervention.

dtbuchholz (approved):
lgtm

dtbuchholz:
@mzkrasner  ah actually, we're missing the "current particpants" data from the response. e.g., when we get all `/competitions`, the frontend needs to be able to display `N/maxParticipants registered`, but we only have `maxParticipants`:
```
  {
    "crossChainTradingType": "disallowAll",
    "rewards": null,
    "id": "6f7db5b5-89dc-404a-9920-14ad02cd1307",
    "name": "Automated Competition 3 Pending 1755113473",
    "description": "Third competition created by script, should remain pending.",
    "type": "trading",
    "externalUrl": "https://recall.network",
    "imageUrl": "https://fastly.picsum.photos/id/237/200/300.jpg?hmac=TmmQSbShHz9CdQm0NkEjx1Dyh_Y984R9LpNrpvH2D_U",
    "startDate": null,
    "endDate": null,
    "votingStartDate": "2025-08-13T20:34:34.036Z",
    "votingEndDate": "2025-08-14T19:34:34.036Z",
    "joinStartDate": "2025-08-13T19:42:48.286Z",
    "joinEndDate": "2025-08-14T19:42:48.286Z",
    "maxParticipants": 1,
    "status": "pending",
    "sandboxMode": false,
    "createdAt": "2025-08-13T19:31:13.419Z",
    "updatedAt": "2025-08-13T19:31:13.419Z"
  }
```

apps/api/e2e/tests/competition.test.ts
```diff
@@ -4442,4 +4449,651 @@ describe("Competition API", () => {
...
+        undefined, // endDate
+        undefined, // votingStartDate
+        undefined, // votingEndDate
+        undefined, // joinStartDate
+        undefined, // joinEndDate
+        maxParticipants, // maxParticipants
+      );
+
+      expect(createResponse.success).toBe(true);
+      if (!createResponse.success)
```
stbrody: this check is redundant and unnecessary.  AI seems to really like having if-statements checking conditions that were already asserted on.  Its a pattern I've been seeing throughout our tests and one I'd very much like to break, so lets make sure not to add more examples of it for the AI to copy.

I'm not going to comment on every line that does this in this test file, but there are a bunch, let's please remove them all.

Basically I'm finding that any time you see an if-statement in a test file, you should probably be suspicious of it since it should probably be an assert instead

apps/api/src/services/competition-manager.service.ts
```diff
@@ -973,20 +1006,23 @@ export class CompetitionManager {
       throw error;
     }
 
-    // 5. Check agent status is eligible
+    // 5. Check participant limit
+    await this.validateParticipantLimit(competition);
```
stbrody: this introduces a race condition.  If two agents try to register at the same time, they could both get past this check and be added, even if there's only room left for one of them.  This will get even worse when we have multiple app servers processing requests and updating the database in parallel.

This check should be done within the same database transaction that adds the user to the competition, so that the transaction succeeds or fails atomically including the agent limit check

stbrody (dismissed):

dtbuchholz:
@stbrody i've addressed your comments in a separate PR, which i'll merge into this one: https://github.com/recallnet/js-recall/pull/1072

that PR also adds a `registeredAgents` field to all competition-related responses. there are a few places in the frontend that need to know if a user/agent can join the limited comp based on this information.

dtbuchholz:
before we merge this—what do we think about the naming convention for `registeredParticipants` & `maxParticipants`? 

should this, instead, be `registeredAgents` & `maxAgents`? or even nested in some field like `registrationLimit: { current: 0, max: 10, reserved: 5, isFull: false }` (where `reserved` is a future-state feature)? (the latter would be a bit of a pain due to `maxParticipants` being in the raw db response, but the API shape is a bit cleaner.)

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -269,28 +295,57 @@ async function updateOneImpl(
...
  * @param competitionId Competition ID
  * @param agentId Agent ID to add
+ * @param maxParticipants Maximum number of participants allowed (null means no limit)
+ * @throws Error if participant limit would be exceeded
  */
 async function addAgentToCompetitionImpl(
   competitionId: string,
   agentId: string,
-) {
+  maxParticipants?: number | null,
```
stbrody: can the maxParticipants for a competition change after the competition is already live?  If so, we'd need to read this value from the db within the transaction rather than getting it as an arg that could be out of date.

But if we're confident that we're never going to change the size of an already existing competition, then how its done here is fine

apps/api/src/database/repositories/agent-repository.ts
```diff
@@ -154,7 +155,21 @@ async function findAgentCompetitionsImpl(
     }
 
     let query = db
-      .select()
+      .select({
```
stbrody: hmm, this change seems more complicated than I expected it to be...

I wonder if it would be cleaner to have an integer field in the competition table tracking the number of registered agents, that we increment/decrement within the transaction to add/remove agents from a comp.  Then we don't need to do a SQL count every time we need this, it's just already there in the competition table.

WDYT @asutula?

apps/api/src/database/repositories/competition-repository.ts
dtbuchholz: nope, it's an admin-set value that we'll set in the "pending" / registration phase. we could potentially change it pre-comp going live, but we wouldnt ever change this once the comp starts because agents can join once it begins.

apps/api/src/database/repositories/agent-repository.ts
dtbuchholz: fwiw that'd also make the drizzle-inferred types cleaner because we could infer the `registeredAgents` field in the native type vs needing to extend the type.

asutula: There are better ways to do this query to get the count that would read cleaner and provide the correct TS type (use the Drizzle `count` function), but yes, it's probably better from a performance perspective to maintain a count column. Would be more simple as well.

apps/api/src/controllers/admin.controller.ts
```diff
@@ -1828,7 +1831,11 @@ export function makeAdminController(services: ServiceRegistry) {
         }
 
         // Add agent to competition using repository method
-        await addAgentToCompetition(competitionId, agentId);
+        await addAgentToCompetition(
+          competitionId,
+          agentId,
+          competition.maxParticipants,
+        );
```
asutula: Just wondering how this argument was missing before, but I don't see an update to the `addAgentToCompetition` function to add the new parameter. Maybe I'm missing something? Also, seems strange to have to specify that as a parameter.

dtbuchholz:
fair points. btw i took over this PR because i thought this would be a quick change. i'm still focused on the privy integration feature...in case anyone wants to take this PR back over and fix the two things noted above:
- in `addAgentToCompetitionImpl`, use a db query to check `maxParticipants` instead of passing as a param
- revert the complex count logic in `fedb0de` (but keeps the tests). add a new `registeredParticipants` column to `competitions`, and then update this field whenever we write to the `competition_agents` table. iirc this only happens when:
   - an agent joins a competition
   - whenever an admin removes an agent from a competition

mzkrasner:
> fair points. btw i took over this PR because i thought this would be a quick change. i'm still focused on the privy integration feature...in case anyone wants to take this PR back over and fix the two things noted above:
> 
> * in `addAgentToCompetitionImpl`, use a db query to check `maxParticipants` instead of passing as a param
> * revert the complex count logic in `fedb0de` (but keeps the tests). add a new `registeredParticipants` column to `competitions`, and then update this field whenever we write to the `competition_agents` table. iirc this only happens when:
>   
>   * an agent joins a competition
>   * whenever an admin removes an agent from a competition

I can wrap these up

mzkrasner:
> > fair points. btw i took over this PR because i thought this would be a quick change. i'm still focused on the privy integration feature...in case anyone wants to take this PR back over and fix the two things noted above:
> 
> > 
> 
> > * in `addAgentToCompetitionImpl`, use a db query to check `maxParticipants` instead of passing as a param
> 
> > * revert the complex count logic in `fedb0de` (but keeps the tests). add a new `registeredParticipants` column to `competitions`, and then update this field whenever we write to the `competition_agents` table. iirc this only happens when:
> 
> >   
> 
> >   * an agent joins a competition
> 
> >   * whenever an admin removes an agent from a competition
> 
> 
> 
> I can wrap these up

Latest changes should be in line with your comment and ready to review

apps/api/src/database/repositories/agent-repository.ts
dtbuchholz: @mzkrasner we want to revert my overly complex logic here, and since `registeredParticipants` is a table column, we can use the original query logic

apps/api/src/database/repositories/agent-repository.ts
```diff
@@ -769,7 +786,22 @@ async function findUserAgentCompetitionsImpl(
     }
 
     let fullResultsQuery = db
-      .select()
+      .select({
```
dtbuchholz: same here

apps/api/src/services/agent-manager.service.ts
```diff
@@ -970,10 +970,10 @@ export class AgentManager {
       // Get competitions from repository
       const results = await findAgentCompetitions(agentId, params);
 
-      // Filter out null competitions upfront to optimize processing
+      // Filter out null competitions and ensure required fields are present
       const validCompetitions = results.competitions.filter(
-        (comp): comp is SelectCompetition => comp !== null,
-      );
+        (comp) => comp !== null,
+      ) as (SelectCompetition & { registeredParticipants: number })[];
```
dtbuchholz: same here. the `SelectCompetition` type has the `registeredParticipants`

apps/api/src/services/agent-manager.service.ts
```diff
@@ -1557,7 +1557,7 @@ export class AgentManager {
    * @returns Enhanced competition with agent metrics
    */
   async attachCompetitionMetrics(
-    competition: SelectCompetition,
+    competition: SelectCompetition & { registeredParticipants: number },
```
dtbuchholz: remove ` & { registeredParticipants: number }`

apps/api/src/controllers/competition.controller.ts
```diff
@@ -622,10 +622,14 @@ export function makeCompetitionController(services: ServiceRegistry) {
           0,
         );
 
+        // Get registered participant count (includes all registered agents, not just those with trades)
+        const participantCount =
```
dtbuchholz: unneeded. we can use `competition.registeredParticipants`

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -1098,6 +1287,33 @@ async function countAgentCompetitionsImpl(agentId: string): Promise<number> {
   }
 }
 
+/**
+ * Count the number of active participants in a competition
+ * @param competitionId The ID of the competition
+ * @returns The number of active participants in the competition
+ */
+async function getParticipantCountImpl(competitionId: string): Promise<number> {
```
dtbuchholz: i _think_ we can drop this method

apps/api/src/services/competition-manager.service.ts
```diff
@@ -353,11 +376,18 @@ export class CompetitionManager {
       `[CompetitionManager] Competition ended. ${competitionAgents.length} agents remain 'active' in completed competition`,
     );
 
-    // Update competition status
+    // Update competition status (re-fetch to get the current registeredParticipants)
```
dtbuchholz: why is this needed?

PR #1079, opened 2025-08-18
feat(api): add endpoints to retrieve trades for competitions and agents

### LINKED ISSUE #908: [FEATURE] Agent trade history & logs (backed APIs)
**Problem Statement**
- Our mini apps show per-agent and per-comp trading history. But, this data is accessed through direct db queries.
- Community members look at this view quite often, and we need to add it to the app.

**Proposed Solution**
1. Create a new API for `/competitions/:id/trades` to let users fetch pagination "most recent" trades (chronologically across all agents). It should include the trade log (price etc), agent, and agent reason.
2. Create a new API for `/competitions/:id/agents/:id/trades` so that you can look up an agents trades in a comp.

For (2), we could also consider adding these...we should do whatever is more RESTful:
- `/agents/:id/competitions/:id/trades` => get info about an agent in comp, including trades
- `/agents/:id/trades/:competitionId` => ^same

**Impact**
users already "have" this feature, so we need it

**Priority**
high

ramirotw:
Fixes #908 

- Implemented `getCompetitionTrades` and `getAgentTradesInCompetition` methods in the competition controller.
- Added corresponding routes in `competitions.routes.ts` for fetching trades.
- Updated the API client with methods to call these new endpoints.
- Updated the OpenAPI documentation and added tests.

### DIFF
--- a/apps/api/src/routes/competitions.routes.ts
+++ b/apps/api/src/routes/competitions.routes.ts
@@ -98,6 +98,15 @@ export function configureCompetitionsRoutes(
    *                         type: string
    *                         format: date-time
    *                         description: When the competition was last updated
+   *                       registeredParticipants:
+   *                         type: integer
+   *                         description: Number of participants registered for this competition
+   *                         example: 10
+   *                       maxParticipants:
+   *                         type: integer
+   *                         nullable: true
+   *                         description: Maximum number of participants allowed to register for this competition. null means no limit.
+   *                         example: 50
    *                       rewards:
    *                         nullable: true
    *                         type: array
@@ -679,6 +688,15 @@ export function configureCompetitionsRoutes(
    *                       type: string
    *                       format: date-time
    *                       description: When the competition was last updated
+   *                     registeredParticipants:
+   *                       type: integer
+   *                       description: Number of participants registered for this competition
+   *                       example: 10
+   *                     maxParticipants:
+   *                       type: integer
+   *                       nullable: true
+   *                       description: Maximum number of participants allowed to register for this competition. null means no limit.
+   *                       example: 50
    *                     rewards:
    *                       type: array
    *                       nullable: true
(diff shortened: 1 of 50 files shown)


apps/api/src/routes/competitions.routes.ts
```diff
@@ -1093,5 +1093,141 @@ export function configureCompetitionsRoutes(
...
+   *       404:
+   *         description: Competition not found
+   *       401:
+   *         description: Unauthorized - Missing or invalid authentication
+   *       500:
+   *         description: Server error
+   */
+  router.get(
+    "/:competitionId/trades",
+    ...authMiddlewares,
```
dtbuchholz, line 1184: i'm wondering if we want to use `optionalAuthMiddleware` and down below. if we use `authMiddlewares`, it'll gate access to in-app users who've authenticated—but prevent non-auth'd users from seeing the trade data.

ramirotw: who would be the users of these endpoints?

dtbuchholz: the idea is that our app will display this info. since we want parity with the current "mini apps" like https://tradingcomp.recall.network/trades, and since these are "open", i suppose we'd want to keep that behavior.

dtbuchholz: actually....it's better to put it behind auth. and when we build the frontend component, we can display some CTA to "sign up / log in to view the full history"

ramirotw: btw, do we have the UI implementation issue for this? I can take it

dtbuchholz (approved):

apps/api/src/routes/competitions.routes.ts
dtbuchholz: yup—let me see if we can add you to Linear...for some reason, tickets arent syncing between Github<>Linear.

i just created this: https://github.com/recallnet/js-recall/issues/1094

PR #1083, opened 2025-08-19
Make sorting arrows in tables intutive

timothyylim:
Fixes [REC-65](https://linear.app/recall-labs/issue/REC-65/uiux-optimizations-for-sorting-arrows-in-every-table):

- Arrows cycle through unselected, ascending and descending
- Keeps styling consistent


timothyylim:
@dtbuchholz , this is cleaner and easier to review now

dtbuchholz:
ah, @timothyylim you'll need to set up [GPG signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/checking-for-existing-gpg-keys). the easiest path: "uncommit" the 3 commits from this branch, and just re-commit them as a single commit. or you could do something like [this](https://stackoverflow.com/questions/41882919/is-there-a-way-to-gpg-sign-all-previous-commits)

timothyylim:
hey @dtbuchholz , I think the deployment might have failed because I don't have a seat on Vercel

Issue #1069, opened 2025-08-13
Stress Testing API Layer


stephhuynh18:
## Problem

We want to build "confidence" in our system to handle a large 7 day paper trading competition for a partner.

Should be able to handle on-going voting of 200k votes

## Context

* Supabase has flagged our slow queries  [https://supabase.com/dashboard/project/fefmaihjhhkpzsnetesg/advisors/query-performance](https://supabase.com/dashboard/project/fefmaihjhhkpzsnetesg/advisors/query-performance)

## Proposed Solution

In staging

* create a test comp with *N* agents
* run it for a few days
* have agents make some *M* avg trades per day
* have bots hit api endpoints used for display of the next.js app.  The large amount of traffic viewing the site was the main root of all the issues with the july 23rd comp (Or put another way, there is only going to be 50-100 agents using the trading endpoints, but there will be 500k+ trying to load the website)
* ping pertinent APIs and capture varying response times

## Need to do

- [ ] Trades submitted by a lot of agents in parallel across different tokens
- [ ] Per competition page which loads results per active competition 
  - [ ] tests portfolio snapshots
- [ ] Voting (gated by user auth) needs to use playwright 
- [ ] able to have different levels of load 
- [ ] Make sure ending competition is successful while being hammered

## Nice to Have

* SIWE Flow (but may not need it) 

## DOD

* Figure out what the agent and user limit is 
* repeatable manual stress test
* DOES NOT INCLUDE FIXES FOR ANY IDENTIFIED PERF PROBLEMS

PR #1085, opened 2025-08-19
feat(load-test): add leaderboard artillery test

ramirotw:
Fixes [#1069](https://github.com/recallnet/js-recall/issues/1069)

## WIP

This first implementation aims to test the loading of the Leaderboards page and navigation to the Agent Profile using Playwright as the engine, running for 5 minutes with a load of 2 new users per second. The number of concurrent users is capped at 100 in local mode to avoid resource saturation.

More complex scenarios can be designed to better simulate real user situations. It’s also possible to test the API directly without going through the web app (and using Playwright), this would reduce resource consumption and allow for a significant increase in the number of virtual users.

### DIFF
--- a/packages/load-test/.env.example
+++ b/packages/load-test/.env.example
@@ -1,0 +1,1 @@
+HOST=https://staging.recall.network
--- a/packages/load-test/README.md
+++ b/packages/load-test/README.md
@@ -1,0 +1,15 @@
+# Load Test Package
+
+This package contains load tests for the `apps/api` and `apps/comps` applications using Artillery.
+
+## Usage
+
+To run the load tests, use the following command:
+
+```
+pnpm --filter @recallnet/load-test test:leaderboard
+```
+
+### Scenarios
+
+- `test:leaderboard`: Tests the leaderboard page with Playwright programmatically.
--- a/packages/load-test/eslint.config.js
+++ b/packages/load-test/eslint.config.js
@@ -1,0 +1,4 @@
+import { config } from "@recallnet/eslint-config/base";
+
+/** @type {import("eslint").Linter.Config} */
+export default config;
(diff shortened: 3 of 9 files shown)


apps/api/src/index.ts
```diff
@@ -138,7 +138,7 @@ app.use(
 
 // Apply rate limiting middleware AFTER authentication
 // This ensures we can properly rate limit by agent/user ID
-app.use(rateLimiterMiddleware);
+//app.use(rateLimiterMiddleware);
```
ramirotw: @dtbuchholz as I mentioned the other day, we need a proper way to disable and/or adjust the rate limiter so it doesn't block the local load testing

dtbuchholz: Can we set up some new env var to conditionally disable it?

ramirotw: [68880dd](https://github.com/recallnet/js-recall/pull/1085/commits/68880dd9bbb2f63b65f28f447a94afb245ec8f5f)

the e2e test that fails is unrelated to these changes

image

carsonfarmer:
> running the e2e load test with Playwright and thousands of virtual users will require some AWS Fargate or Lambda support. I tried running it locally on a MacBook Pro M4 with 64GB of RAM, and it struggled with 100-300 virtual users. Using plain HTTP to hit the API directly needs less hardware and might work locally, but it’s not as realistic as a full end-to-end test.

Interesting. Could you walk me through this at some point this week? I'd like to get a sense of costs and how often we'd want to run this at "full capacity". In the mean time, should we start scoping a follow on project to get this scaled up?

asutula:
@smrz2001 can you be sure to take a look at this? Maybe you can take it over as part of https://linear.app/recall-labs/project/performance-testing-and-baseline-establishment-183ef6e31a5a/overview?

PR #1093, opened 2025-08-19
Fix API Key Validation N+1 Query Pattern

smrz2001:
## Problem
API key validation was performing O(n) full table scans on every authentication request, causing performance bottlenecks that could lead to production failures at scale.

## Database Changes
- New column: `agents.api_key_hash` (varchar(64), nullable)
- New index: `idx_agents_api_key_hash` (btree)
- Migration script: `pnpm --filter api scripts:migrate-api-key-hashes`

## Performance Impact
- **Before**: `SELECT * FROM agents` (O(n) full table scan)
- **After**: `SELECT * FROM agents WHERE api_key_hash = $1` (O(1) indexed lookup)
- **Result**: Eliminates N+1 query pattern, enables production scaling


apps/api/src/services/agent-manager.service.ts
```diff
@@ -314,49 +319,73 @@ export class AgentManager {
...
+      if (!agent) {
+        // Try to find and migrate legacy agent without hash
+        // This provides backward compatibility during migration period
+        const agents = await findAll();
+
+        for (const legacyAgent of agents) {
+          // Skip agents that already have a hash
+          if (legacyAgent.apiKeyHash) {
+            continue;
+          }
```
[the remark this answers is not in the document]
smrz2001: The N+1 pattern during migration is unavoidable. However, this is:

1. Temporary - only during migration period until all agents have hashes
2. Infrequent - only happens for agents without hashes making requests
3. Self-improving - each successful migration reduces future lookups

apps/api/scripts/migrate-api-key-hashes.ts
```diff
@@ -0,0 +1,297 @@
...
+    .where(isNull(agents.apiKeyHash));
+  const missingHashes = missingHashesResult[0]?.count || 0;
+
+  const existingHashesResult = await db
+    .select({ count: sql<number>`COUNT(*)::int` })
+    .from(agents)
+    .where(sql`${agents.apiKeyHash} IS NOT NULL`);
+  const existingHashes = existingHashesResult[0]?.count || 0;
+
+  return {
```
stbrody, line 73: note that because these queries aren't in a transaction, if new agents register in between the steps of the different queries, then the `existingHashes` + `missingHashes` could wind up being greater than `totalAgents`.

Not a problem given this is only used for diagnostic logging purposes, but probably worth a comment

apps/api/src/services/agent-manager.service.ts
```diff
@@ -314,49 +320,84 @@ export class AgentManager {
...
+      let agent = await findByApiKeyHash(apiKeyHash);
 
-            // Add to cache (now includes inactive agents for per-competition checking)
-            this.apiKeyCache.set(apiKey, {
-              agentId: agent.id,
-              key: apiKey,
-            });
+      if (!agent) {
+        // Try to find and migrate legacy agent without hash
+        // This provides backward compatibility during migration period
```
stbrody: Let's file a ticket to remove this fallback logic once all agents have been migrated, and add a `TODO(REC-XXX)` todo comment referencing the ticket to go back and delete this fallback

apps/api/src/services/agent-manager.service.ts
```diff
@@ -314,49 +320,84 @@ export class AgentManager {
         return cachedAuth.agentId;
       }
```
stbrody, line 422: optional nitpick: this whole function is pretty large and could benefit from being broken up with some helper functions

PR #1095, opened 2025-08-20
feat: rm price cache in the database

stbrody:


apps/api/package.json
```diff
@@ -48,7 +48,6 @@
     "scripts:backfill-constraints": "tsx scripts/backfill-comp-constraints.ts",
     "scripts:backfill-stats": "tsx scripts/backfill-stats.ts",
     "scripts:delete-competitions": "tsx scripts/delete-competitions.ts",
-    "scripts:fix-alphawave-volumes": "tsx scripts/fix-alphawave-trade-volumes.ts",
```
stbrody, line 51: removed this script since it heavily used the price table.  Since it has the word "alphawave" in its name, I'm assuming it was a one-time script for that comp and that we don't need it anymore going forward.  @dtbuchholz @mzkrasner please let me know if that isn't the case and we actually need to update the script to work without the price table

dtbuchholz (approved):

apps/api/package.json
dtbuchholz: it was a one-time script that we used to fix bad alphawave data. not needed

PR #1096, opened 2025-08-20
feat: Make price cache fixed size and use LRU eviction

stbrody:


stbrody:
> can we add the new env vars to the .env.example, remove the old ones from .env.test, and remove whatever we dont need from the CI env setup [here](https://github.com/recallnet/js-recall/blob/lruCache-rec-556/.github/workflows/api-ci.yml?rgh-link-date=2025-08-20T22%3A27%3A09Z#L256)

I don't think the old env vars show up in `.env.test`, and the CI setup file is updated already between this PR and the follow-up PR (https://github.com/recallnet/js-recall/pull/1097).

So I think the only thing to do here is add an example with the new env vars to `.env.example`.  I'll go ahead and do that now

PR #1097, opened 2025-08-20
feat: Remove unused env vars and remove snapshot interval from rules response

stbrody:
builds on https://github.com/recallnet/js-recall/pull/1096


apps/api/src/controllers/competition.controller.ts
```diff
@@ -384,17 +384,13 @@ export function makeCompetitionController(services: ServiceRegistry) {
         };
         const slippageFormula =
           "baseSlippage = (tradeAmountUSD / 10000) * 0.05%, actualSlippage = baseSlippage * (0.9 + (Math.random() * 0.2))";
-        const portfolioSnapshots = {
-          interval: `${config.portfolio.snapshotIntervalMs / 60000} minutes`,
```
stbrody, line 395: Removing this because the config used in the app is totally unrelated from the cron interval running the script that is now responsible for taking snapshots, so the data in this response could very easily be wrong.

dtbuchholz (approved):

PR #1098, opened 2025-08-20
fix: Mark competition as ended before taking final snapshot to prevent new trades from coming in

stbrody:


stbrody:
superceded by https://github.com/recallnet/js-recall/pull/1099

PR #1099, opened 2025-08-20
fix: Clean up endCompetition

stbrody:
Mostly code cleanup, but there is one functional change: moving marking the competition as ended to be before we take the final snapshot.  This ensures that no last trades can sneak in after the final snapshot.

PR is composed of 3 independent commits, so you can review the functional change separately from the cleanups if you want

### DIFF
--- a/apps/api/src/services/competition-manager.service.ts
+++ b/apps/api/src/services/competition-manager.service.ts
@@ -364,6 +364,53 @@ export class CompetitionManager {
   }
 
   /**
+   * Calculates final leaderboard, enriches with PnL data, and persists to database
+   * @param competitionId The competition ID
+   * @param totalAgents Total number of agents in the competition
+   * @returns Number of leaderboard entries that were saved
+   */
+  private async calculateAndPersistFinalLeaderboard(
+    competitionId: string,
+    totalAgents: number,
+  ): Promise<number> {
+    // Get the leaderboard (calculated from final snapshots)
+    const leaderboard = await this.getLeaderboard(competitionId);
+
+    const enrichedEntries = [];
+
+    // Note: the leaderboard array could be quite large, avoiding Promise.all
+    // so that these async calls to get pnl happen in series and don't over
+    // use system resources.
+    for (let i = 0; i < leaderboard.length; i++) {
+      const entry = leaderboard[i];
+      if (entry === undefined) continue;
+
+      const { pnl, startingValue } =
+        await this.agentManager.getAgentPerformanceForComp(
+          entry.agentId,
+          competitionId,
+        );
+
+      enrichedEntries.push({
+        agentId: entry.agentId,
+        competitionId,
+        rank: i + 1, // 1-based ranking
+        pnl,
+        startingValue,
+        totalAgents,
+        score: entry.value, // Portfolio value in USD is saved as `score`
+      });
+    }
+
+    // Persist to database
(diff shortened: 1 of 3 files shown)


apps/api/src/services/competition-manager.service.ts
```diff
@@ -354,6 +363,53 @@ export class CompetitionManager {
...
+    competitionId: string,
+    totalAgents: number,
+  ): Promise<number> {
+    // Get the leaderboard (calculated from final snapshots)
+    const leaderboard = await this.getLeaderboard(competitionId);
+
+    const enrichedEntries = [];
+
+    // Note: the leaderboard array could be quite large, avoiding Promise.all
+    // so that these async calls to get pnl happen in series and don't over
```
joewagner, line 382: :heart: :heart: :heart:

apps/api/src/services/competition-manager.service.ts
```diff
@@ -558,87 +586,164 @@ export class CompetitionManager {
...
-      //  the state of several tables. We should refactor and ideally only have
-      //  one, which doesn't have to do calculations for every (uncached) request.
-      const competition = await findById(competitionId);
-      const competitionStatus = competition?.status;
+  private async getSavedLeaderboard(
+    competitionId: string,
+  ): Promise<LeaderboardEntry[]> {
+    const leaderboardEntries =
+      await findLeaderboardByTradingComp(competitionId);
+    return leaderboardEntries.map((entry) => ({
```
joewagner: This is more of a question than suggestion, but any reason we to this `map` to remove the extra entry fields vs. reducing the fields in the SQL select statement?

dtbuchholz (approved):

joewagner (approved):
I wasn't tagged, but figured I'd try to help do reviews before eod.
This looks good, I left a few comments but nothing blocking merge.

apps/api/src/services/competition-manager.service.ts
stbrody: Good call, working on a fix for that now

apps/api/src/services/competition-manager.service.ts
```diff
@@ -558,87 +586,164 @@ export class CompetitionManager {
...
+      agentId: entry.agentId,
+      value: entry.score,
```
stbrody: Addressing Joe's comment above made me notice this... is it weird that we're taking the `score` from the `competitionsLeaderboard` table and returning it as `value` here?  I would have assumed `value` meant total USD portfolio value and score meant AgentRank score, but we seem to be conflating those here... is this a bug?  @KGmajor @dtbuchholz

stbrody: I guess maybe this is intentional? https://github.com/recallnet/js-recall/blob/09153fba9795b79abcfeffd73c28324147420e37/apps/api/src/services/competition-manager.service.ts#L421

dtbuchholz: @stbrody we dont calculate the "agent rank" in a competition; that's a global leaderboard concept.

a per-comp leaderboard's "score" is just the best porfolio value. i.e., this is just confusing column names / variable naming convention. aka value == score in this scenario.

and leaderboard here means the per-comp leaderboard, not the global one. another confusing term...

stbrody: huh okay, sounds good

PR #1100, opened 2025-08-21
chore: Don't deconstruct every field when passing the full object will suffice

stbrody:
AI seems to really like to spell out every field in an object, which bloats the code and makes it harder to follow.  Please if possible add some checks into your workflow to keep things simple when possible


dtbuchholz (approved):
🚀

PR #1102, opened 2025-08-22
feat(api): privy wallet integration

dtbuchholz:
closes [APP-10](https://linear.app/recall-labs/issue/APP-10/login-wallet-ux-improvements-eg-privy)


apps/api/e2e/tests/email-subscription.test.ts
```diff
@@ -0,0 +1,27 @@
...
+      email: "alice@example.com",
+    });
+
+    const authResult = await client.authenticateWithPrivy(testUser);
+    expect(authResult.success).toBe(true);
+
+    const user = (await client.getUserProfile()) as UserProfileResponse;
+    expect(user.success).toBe(true);
+    // TODO: ideally, this is `true` but tests cannot currently override the behavior of the email service
+    expect(user.user.isSubscribed).toBe(false);
```
joewagner: Can we add a few assertions here for user info, e.g. name and email.

apps/api/package.json
```diff
@@ -66,9 +66,9 @@
     "competition"
   ],
   "author": "",
-  "license": "ISC",
+  "license": "MIT OR Apache-2.0",
```
joewagner, line 70: No opinion here, but double checking that we are changing licensing on purpose?

joewagner (commented):
I'm still working through the review but I have 2 inline question/comments, and it seems like we should add to one or two of the READMEs with brief instructions on how to start the dev server.
I tried staring and I get the `[AuthMiddleware] Authentication required. Invalid Privy token or no API key provided. Use Authorization: Bearer YOUR_API_KEY` error message when I try to signup.

I have a PRIVY_APP_ID and PRIVY_APP_SECRET in my env file, but still get the error.  I'm guessing I need to get a `PRIVY_JWKS_PUBLIC_KEY`, but not sure how to get that.

apps/api/src/lib/privy/utils.ts
```diff
@@ -0,0 +1,168 @@
...
+ * @param request - The request object to extract the token from.
+ * @returns The Privy identity token.
+ */
+export function extractPrivyIdentityToken(request: {
+  headers?: IncomingHttpHeaders;
+  cookies?: { get: (name: string) => { value: string } | undefined };
+}): string | undefined {
+  return request.headers?.cookie
+    ?.split("; ")
+    .find((c: string) => c.startsWith("privy-id-token="))
```
joewagner, line 24: This might be a naive question, but where/when is this cookie set?  
I see that the tests have the AxiosInstance directly write the cookie to the request headers. I'm assuming our cookies will be http only, so the way the tests work isn't possible for an actual client.  And I haven't found anywhere in the api where the cookie is set.
Am I missing the `Set-Cookie` response header, or is the upcoming PR going to add this?

joewagner:
In general this is looking good, It's hard to give a :+1:  until the comps app part is working too.  
My main concern is explained in [this comment](https://github.com/recallnet/js-recall/pull/1102/files#r2301472767), and without a working client it's hard to test, so maybe I'm missing something on how the flow works.

dtbuchholz:
note: i opened a draft PR for the frontend stuff. still working through some kinks and logic cleanups, but it should be enough to test against the backend https://github.com/recallnet/js-recall/pull/1122. the only thing you'll need to do is add an env var to `apps/comps`: `NEXT_PUBLIC_PRIVY_APP_ID`

dtbuchholz:
closing in favor of https://github.com/recallnet/js-recall/pull/1102

PR #1107, opened 2025-08-25
refactor: rewards should point to competitions and not epochs

brunocalza:
# Context

Closes [APP-136](https://linear.app/recall-labs/issue/APP-136/rewards-database-schema-changes)

We already had some rewards tables from [previous work ](https://github.com/recallnet/js-recall/pull/436). But those tables were pointing to `epochs` table. As of latest spec, we are implementing rewards by competition for TGE.

This PR:
- Removes epoch from `rewards`, `rewards_tree` and `rewards_roots` table, and add `competition_id`. This is a breaking change, but those tables are not used. So, not a big deal.
- Refactors `rewards-repository`, `rewards.service`, `rewards.test.ts` and scripts to accommodate that change

### DIFF
--- a/apps/api/e2e/tests/rewards.test.ts
+++ b/apps/api/e2e/tests/rewards.test.ts
@@ -6,8 +6,8 @@ import { beforeEach, describe, expect, test } from "vitest";
 
 import { db } from "@/database/db.js";
 import { insertRewards } from "@/database/repositories/rewards-repository.js";
+import { competitions } from "@/database/schema/core/defs.js";
 import {
-  epochs,
   rewards,
   rewardsRoots,
   rewardsTree,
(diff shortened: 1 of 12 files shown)


apps/api/e2e/tests/rewards.test.ts
```diff
@@ -17,26 +17,31 @@ import { RewardsService, createLeafNode } from "@/services/rewards.service.js";
...
+        status: "active",
+        type: "trading",
+        startDate: new Date(),
+        endDate: new Date(Date.now() + 7 * 24 * 60 * 60 * 1000), // 7 days from now
       })
       .returning();
 
-    if (!epoch) {
-      throw new Error("Failed to create test epoch");
+    if (!competition) {
```
joewagner: We have been trying to clean up tests, and there is a PR that landed yesterday, so you will probably have conflicts in this file.
  
In general, this should be an assertion instead of a conditional that throws, i.e. 
```
expect(competition).toBeDefined();
expect(competition?.id).toBe(competitionId);
```  
Also, any `console.log`s should be removed.

joewagner (approved):
Looks good!  I left one note inline concerning how you handle merge/rebase conflicts

apps/api/e2e/tests/rewards.test.ts
brunocalza: Thanks! I'll make that change

PR #1110, opened 2025-08-25
feat: Implement getAgentMetrics in terms of getBulkAgentMetrics

stbrody:
They currently have wildly different implementations which creates a ton of code duplication and introduces the potential for bugs to arise from subtle differences between the two.

This does mean now that loading the metrics for a single agent will use an "in" statement in the underlying SQL query, but I'd like to be believe that the postgres query optimizer is smart enough that there's no perf overhead for an "in" query with only a single entry in the "in" statement...

Also includes some general code cleanups to the bulk query path to improve layering separation and add more explicit type information

### DIFF
--- a/apps/api/src/database/repositories/leaderboard-repository.ts
+++ b/apps/api/src/database/repositories/leaderboard-repository.ts
@@ -24,6 +24,7 @@ import {
 } from "@/database/schema/trading/defs.js";
 import { repositoryLogger } from "@/lib/logger.js";
 import { createTimedRepositoryFunction } from "@/lib/repository-timing.js";
+import type { RawAgentMetricsQueryResult } from "@/types/agent-metrics.js";
 import {
   COMPETITION_AGENT_STATUS,
   COMPETITION_STATUS,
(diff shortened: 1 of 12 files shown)


apps/api/src/database/repositories/leaderboard-repository.ts
```diff
@@ -348,15 +278,25 @@ async function getBulkAgentMetricsImpl(agentIds: string[]): Promise<
       )
       .groupBy(competitionsLeaderboard.agentId);
 
+    // Query 8: Get all agent scores for rank calculation in service layer
+    const allAgentScoresQuery = dbRead
```
joewagner, line 282: We are loading every agent score row here.  Is that ok?
It looks like we only have 44 scored agents currently, but we will in theory have 100,000+ agents in the future right?

apps/api/src/services/agent-metrics-helper.ts
```diff
@@ -0,0 +1,214 @@
...
+          pnl: row.pnl,
+        },
+      ]),
+    );
+
+    // Calculate global ranks from all agent scores
+    const globalRanksMap = new Map<string, number>();
+
+    // Sort all agents by ordinal (descending) to determine ranks
+    const sortedAgents = [...rawResults.allAgentScores].sort(
```
joewagner, line 78: Related to the above question: 
With our current prod data the length of `allAgentScores` is ~44, but potentially it will get very large right?  Maybe we need to handle this differently?

joewagner (approved):
This is a great improvement on what we have right now, so I say go for it if you want to just get it in and interate.
I do think that we have a future problem the will arise when the size of the agent_score table grows.  I mention this with inline comments.

apps/api/src/database/repositories/leaderboard-repository.ts
dtbuchholz: since the original logic technically does this, i created a follow up https://linear.app/recall-labs/issue/APP-202/optimize-and-limit-paginate-getagentmetrics-logic. i think we'll want to address this before the "always weekly comps" happens.

dtbuchholz:
(merging since spencer is ooo)

PR #1114, opened 2025-08-26
fix(api): remove full table scan for agents

joewagner:
closes https://linear.app/recall-labs/issue/REC-576/remove-legacy-api-key-validation-fallback-after-migration


smrz2001 (approved):
🚀

PR #1116, opened 2025-08-26
fix(api): remove full table scan query methods

joewagner:
## Overview

After the issues with the competition today, I went ahead and removed any attempts to get all agents with a single database query (`select * from agents;`).  
The exception is the scripts.  For scripts I replaced the get all agents with the new get agents with pagination, but the pagination allows for 1M agents, so the script still gets all the agents.

## Summary from Claude:

**Performance Optimization:** Replaced `getAllAgents()` method that was fetching all 145,000 agents from the database with targeted queries:

1. **Competition Controller** (`getLeaderboard`): Now only fetches agents participating in the specific competition using `getAgentsByIds()`

2. **Admin Controller**: 
   - Competition leaderboard: Only fetches competition participants
   - `listAllAgents` endpoint: Added pagination (default 50, max 1000) instead of fetching 100,000+ agents

3. **Scripts**: Replaced service layer calls with direct database queries using `findAll()` since scripts run independently

4. **New Repository Function**: Added `findByIds()` to efficiently fetch multiple agents by their IDs using SQL `IN` clause

**Result:** Dramatically reduced database load and memory usage, especially in production with 145,000 agents.

### DIFF
--- a/apps/api/scripts/competition-status.ts
+++ b/apps/api/scripts/competition-status.ts
@@ -1,6 +1,7 @@
 import * as dotenv from "dotenv";
 import * as path from "path";
 
+import { findAll as findAllAgents } from "@/database/repositories/agent-repository.js";
 import {
   findAll,
   getAgentPortfolioSnapshots,
@@ -127,8 +128,12 @@ async function showCompetitionStatus() {
     // Get agents participating in the competition
     const participatingAgentIds = await getCompetitionAgents(competition.id);
 
-    // Get all agents (for mapping IDs to names)
-    const allAgents = await services.agentManager.getAllAgents();
+    // Get all agents (for mapping IDs to names) - direct DB access for script
+    const allAgents = await findAllAgents({
+      limit: 1000000,
+      offset: 0,
+      sort: "createdAt:desc",
+    });
     const agentMap = new Map(allAgents.map((agent) => [agent.id, agent]));
 
     // Map participating agent IDs to agent objects
--- a/apps/api/scripts/competition-status.ts
+++ b/apps/api/scripts/competition-status.ts
@@ -132,7 +132,7 @@ async function showCompetitionStatus() {
     const allAgents = await findAllAgents({
       limit: 1000000,
       offset: 0,
-      sort: "createdAt:desc",
+      sort: "-createdAt",
     });
     const agentMap = new Map(allAgents.map((agent) => [agent.id, agent]));
 
--- a/apps/api/src/controllers/admin.controller.ts
+++ b/apps/api/src/controllers/admin.controller.ts
@@ -33,6 +33,7 @@ import {
   AdminGetCompetitionSnapshotsParamsSchema,
   AdminGetCompetitionSnapshotsQuerySchema,
   AdminGetPerformanceReportsQuerySchema,
+  AdminListAllAgentsQuerySchema,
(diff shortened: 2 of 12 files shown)


apps/api/scripts/competition-status.ts
```diff
@@ -127,8 +128,12 @@ async function showCompetitionStatus() {
     // Get agents participating in the competition
     const participatingAgentIds = await getCompetitionAgents(competition.id);
 
-    // Get all agents (for mapping IDs to names)
-    const allAgents = await services.agentManager.getAllAgents();
+    // Get all agents (for mapping IDs to names) - direct DB access for script
+    const allAgents = await findAllAgents({
```
joewagner, line 132: This leaves the script logic unaffected, but allows us to remove methods that were causing the 524 errors.  You'll see most of the changes to the scripts are similar to this one.

apps/api/src/controllers/admin.controller.ts
```diff
@@ -902,8 +903,9 @@ export function makeAdminController(services: ServiceRegistry) {
           users.map((user) => [user.id, user.name || "Unknown User"]),
         );
 
-        // Get all agents to map agent IDs to agent names and owners
-        const agents = await services.agentManager.getAllAgents();
```
joewagner, line 906: This controller only needed the agents for the comp in the query string, but it was sending the full table scan.  It is only accessible to admins, so it's not the root of the issue, but should be fixed anyway.

apps/api/src/controllers/admin.controller.ts
```diff
@@ -1157,8 +1159,27 @@ export function makeAdminController(services: ServiceRegistry) {
      */
     async listAllAgents(req: Request, res: Response, next: NextFunction) {
       try {
-        // Get all agents from the database
-        const agents = await services.agentManager.getAllAgents();
```
joewagner, line 1161: I added pagination to this method.  It's used by the performance reports endpoint, and you'll see changes to the openapi spec for that endpoint.

apps/api/src/controllers/competition.controller.ts
```diff
@@ -103,8 +103,13 @@ export function makeCompetitionController(services: ServiceRegistry) {
             competitionId,
           );
 
-        // Get all agents for mapping IDs to names
-        const agents = await services.agentManager.getAllAgents();
```
joewagner, line 107: we only need the agents for the comp being queried for, so doing a full table scan is not needed.

apps/api/src/database/repositories/agent-repository.ts
```diff
@@ -295,6 +295,29 @@ async function findByIdImpl(id: string): Promise<SelectAgent | undefined> {
...
+async function findByIdsImpl(ids: string[]): Promise<SelectAgent[]> {
+  try {
+    if (ids.length === 0) {
+      return [];
+    }
+
+    const results = await db
+      .select()
+      .from(agents)
+      .where(inArray(agents.id, ids));
```
dtbuchholz, line 312: note: iirc you were hesitant to use `inArray` for performance reasons?

joewagner: Good callout, as long as the array being passed in isn't huge it shouldn't be a real issue.  
With our current data and usage patterns, this wouldn't be more than the number of agents in a competition, which is about 50.

PR #1126, opened 2025-08-28
feat(api): cache comps, leaderboard APIs

dtbuchholz:
Adds basic LRU caching for the frontend-heavy `/competitions` (5 min TTL) and `/leaderboard` (30 min TTL) endpoints, both with 100 max entries. These can be overridden with env vars at startup, and you can disable API caching entirely with the `DISABLE_CACHE_API` (e.g., local development).


apps/api/src/config/index.ts
```diff
@@ -338,6 +338,28 @@ export const config = {
...
       10,
     ),
+    api: {
+      leaderboard: {
+        maxCacheSize: parseInt(
+          process.env.CACHE_API_LEADERBOARD_MAX_CACHE_SIZE || "100",
+          10,
+        ),
+        ttlMs: parseInt(
+          process.env.CACHE_API_LEADERBOARD_TTL_MS || "300000", // 5 minutes
```
dtbuchholz: not sure what the default we'd want—idk if the "max size" really matters here, but i use a 5 min cache on all requests to `/leaderboard` and most of the `/competitions` routes

apps/api/src/controllers/competition.controller.ts
```diff
@@ -485,6 +511,23 @@ export function makeCompetitionController(services: ServiceRegistry) {
           ? CompetitionStatusSchema.parse(req.query.status)
           : undefined;
         const pagingParams = PagingParamsSchema.parse(req.query);
+        // Only cache frontend requests (no user or agent context)
+        const isPublicRequest = checkIsPublicRequest(req);
```
dtbuchholz: i figure we could _try_ just caching the API for unauthenticated requests. users / agents / admins wont hit the cache tho...idk if that's the ideal solution to our problem.

apps/api/src/controllers/competition.controller.ts
```diff
@@ -485,6 +511,23 @@ export function makeCompetitionController(services: ServiceRegistry) {
...
+
+        // Build a cache key off stable query params for public requests, only if cache is enabled
+        const isCacheEnabled = checkIsCacheEnabled();
```
dtbuchholz: for tests to pass, i disabled the caching (and also disable caching for development mode)

apps/api/src/controllers/request-helpers.ts
```diff
@@ -160,3 +162,53 @@ export function parseAdminSearchQuery(
...
+export function checkIsCacheEnabled() {
+  return !config.cache.api.disableCaching;
+}
+
+/**
+ * Check if the cache should be created for the request
+ * @param req Express request
+ * @returns True if the cache should be created, false otherwise
+ */
+export function checkShouldCacheResponse(req: Request) {
```
dtbuchholz, line 188: we cache anything from an unauthenticated or authenticated user. agent / API key requests will get the freshest data.

apps/api/src/controllers/competition.controller.ts
```diff
@@ -24,14 +25,58 @@ import { AgentQuerySchema } from "@/types/sort/agent.js";
...
+    max: config.cache.api.competitions.maxCacheSize,
+    ttl: config.cache.api.competitions.ttlMs,
+  }),
+  // Used for: `/competitions/:id/agents`
+  agents: new LRUCache<string, object>({
+    max: config.cache.api.competitions.maxCacheSize,
+    ttl: config.cache.api.competitions.ttlMs,
+  }),
+  // Used for: `/competitions/:id/agents/:agentId/trades`
+  agentTrades: new LRUCache<string, object>({
```
dtbuchholz, line 48: note: since the default max size is 100, i suppose the "agent trades" and "trades" will fill up quickly. we can revisit this for more fain grained or larger max size configs.

apps/api/src/controllers/competition.controller.ts
```diff
@@ -485,6 +530,20 @@ export function makeCompetitionController(services: ServiceRegistry) {
           ? CompetitionStatusSchema.parse(req.query.status)
           : undefined;
         const pagingParams = PagingParamsSchema.parse(req.query);
+
+        // Cache only public (unauthenticated or authenticated user) requests (and disable in test/dev mode)
+        const shouldCacheResponse = checkShouldCacheResponse(req);
```
dtbuchholz, line 535: general pattern:
1. check if you should cache (aka it's a public/user request, and caching is _not_ disabled)
2. generate some key like `competitions:user:{limit:10,offset:10}`
3. get/set the key
4. if you're an agent or admin, you dont hit the cache (these are non-frontend API requests)

PR #1132, opened 2025-08-28
feat: blockchain indexing and boosts DB structure

ukstv:
...and remove epochs, and votes_available, votes_performed - not relevant anymore


apps/api/src/database/schema/voting/defs.ts
```diff
@@ -210,3 +100,48 @@ export const rewardsRoots = pgTable(
     uniqueIndex("uq_rewards_roots_competition_id").on(table.competitionId),
   ],
 );
+
+/** Wallet balances (canonicalized wallet string is the PK) */
+export const boostBalances = pgTable(
+  "boost_balances",
+  {
+    wallet: varchar("wallet", { length: 64 }).primaryKey().notNull(),
```
dtbuchholz: why is this length 64? and not 42

dtbuchholz: related: i wonder what is the best way to handle wallet address uniqueness? 

namely, in our `core` defs, we don't enforce upper/lowercase uniqueness natively in PG. we rely on the repository logic to handle lowercasing the values (prone to human error).

e.g., we could do something like this on our existing `users` table and also in boost logic:
```sql
CREATE OR REPLACE FUNCTION enforce_lowercase_wallet()
RETURNS trigger AS $$
BEGIN
  NEW.wallet_address := lower(NEW.wallet_address);
  RETURN NEW;
END;
$$ LANGUAGE plpgsql;

CREATE TRIGGER wallet_address_lowercase
BEFORE INSERT OR UPDATE ON users
FOR EACH ROW
EXECUTE FUNCTION enforce_lowercase_wallet();

ALTER TABLE users
  ALTER COLUMN wallet_address TYPE text;

CREATE UNIQUE INDEX users_wallet_address_lower_idx
  ON users (lower(wallet_address));
```
but idk what the best practice is here. do we store all addresses as lowercase? do we need to store the checksummed in one col, and the lowercased in another? do we use bytesa?

apps/api/src/database/repositories/boost.repository.ts
```diff
@@ -0,0 +1,183 @@
...
+    this.#db = database;
+  }
+
+  increase(args: BoostDiffArgs): Promise<BoostDiffResult> {
+    const amount = args.amount;
+    if (amount < 0n) {
+      throw new Error("amount must be non-negative");
+    }
+    const idemKey = args.idemKey ?? crypto.randomUUID();
+    const wallet = args.wallet.toLowerCase();
```
dtbuchholz: e.g., you (correctly) lowercase the wallet address here (like we do elsewhere in our codebase). but just as a safeguard, maybe it's time to handle this in PG?

(it's a totally separate topic from this PR...i just wanted to flag it.)

apps/api/src/indexing/blockchain-types.ts
```diff
@@ -0,0 +1,33 @@
+// Raw log from blockchain
+export type RawLog = {
+  address: `0x${string}`;
```
dtbuchholz, line 13: if we _want_, `viem` exports a `Hex` type (as `0x${string}`). (idk if that makes the type checker less friendly, though.)

dtbuchholz:
one nit: where it makes sense, could we add proper TSDoc comments / docstrings? i think it tends to help LLMs/agents better understand thigns. we've have been historically inconsistent at doing so in this codebase...so we might as well include it. or, a follow up "delegate-to-agent" PR is fine w me.

apps/api/src/database/schema/indexing/defs.ts
```diff
@@ -0,0 +1,117 @@
...
+export { indexingEvents, stakes, stakeChanges };
+export type { IndexingEvent, StakeChangeRow, StakeChangeInsert, StakeRow };
+
+const indexingEvents = pgTable(
+  "indexing_events",
+  {
+    id: uuid().primaryKey().notNull(),
+    // Raw event data (stored during indexing)
+    rawEventData: jsonb("raw_event_data").notNull(), // Store the complete raw event payload
+    type: varchar("type", { length: 50 }).notNull(), // Event type (stake, unstake, relock, withdraw)
```
asutula: Thoughts about using an enum?

apps/api/src/database/schema/indexing/defs.ts
```diff
@@ -0,0 +1,117 @@
...
+
+const stakeChanges = pgTable(
+  "stake_changes",
+  {
+    id: uuid("id").primaryKey().notNull(),
+    // who/what
+    stakeId: bigint("stake_id", { mode: "bigint" })
+      .notNull()
+      .references(() => stakes.id),
+    wallet: varchar("wallet", { length: 64 }).notNull(),
```
asutula: Does this duplicate the `wallet` column in `stakes`?

apps/api/src/database/schema/indexing/defs.ts
```diff
@@ -0,0 +1,117 @@
...
+    // delta — signed: +stake, 0 for relock, -move from staked to withdrawable, -withdraw, etc.
+    deltaAmount: bigint("delta_amount", { mode: "bigint" }).notNull(),
+    kind: varchar("kind", { length: 24 }).notNull(), // See EventType for possible values
```
asutula, line 181: Enum?

apps/api/src/database/schema/indexing/defs.ts
```diff
@@ -0,0 +1,117 @@
...
+
+// Export types for use in TypeScript
+type StakeRow = InferSelectModel<typeof stakes>;
+type IndexingEvent = InferSelectModel<typeof indexingEvents>;
+type StakeChangeRow = Omit<InferSelectModel<typeof stakeChanges>, "kind"> & {
+  kind: EventType;
+};
+type StakeChangeInsert = Omit<InferInsertModel<typeof stakeChanges>, "kind"> & {
+  kind: EventType;
+};
```
asutula: Typically we've put these in a `types.ts` file and also created relations in `relations.ts`.

apps/api/src/database/schema/voting/defs.ts
asutula: Seems more simple than that https://claude.ai/share/b12fffda-ebd1-40af-a744-baccfd2e4231

apps/api/src/database/schema/voting/defs.ts
```diff
@@ -210,3 +100,48 @@ export const rewardsRoots = pgTable(
     uniqueIndex("uq_rewards_roots_competition_id").on(table.competitionId),
   ],
 );
+
+/** Wallet balances (canonicalized wallet string is the PK) */
+export const boostBalances = pgTable(
```
asutula, line 129: I guess the plan is to give users their boost balance _per competition_. Would it be smart to add a column and foreign key referencing the competitions table? That way we don't have to do maintenance around resetting all balances to 0 between competitions and also we could support concurrent competitions if needed.

apps/api/src/database/schema/voting/types.ts
asutula, line 1: Add types for the new boost-related tables?

apps/api/src/database/repositories/boost.repository.ts
```diff
@@ -0,0 +1,183 @@
...
+            balance: schema.boostBalances.balance,
+          });
+
+        return {
+          balanceAfter: updated!.balance,
+          changeId: change.id,
+          idemKey: idemKey,
+        };
+      } else {
+        // Already applied earlier: do NOT touch balance, just return current value
```
asutula, line 196: Maybe I don't understand the use of the `boostBalances` table... Is the idea to store the current balance? If so, wouldn't you want to update it here?

asutula (commented):
Nice work. Left a handfulll of comments/questions/possible improvements. The indexing stuff is quite dense and unfamiliar to me.

apps/api/src/database/schema/voting/defs.ts
ukstv: These are the new tables, so wallet, txHash, blockHash are stored as bytea. AI says bytea reduces index size ~ for our size which positively affects cache hits and perf. Ah, unlike PG, compute is easier to scale, so hex <-> bytes conversion is better done on the app side.

For all the other tables, the migration path is the tricky part.

apps/api/src/database/repositories/boost.repository.ts
ukstv: See https://github.com/recallnet/js-recall/pull/1132#discussion_r2310562790

Oh, idemKey is also Uint8Array / bytea.

apps/api/src/indexing/blockchain-types.ts
ukstv: It barely makes reasoning better, but def adds reasoning cycles: "What is Hex here again?" I am reluctant to use `Hex` here :/

apps/api/src/database/schema/indexing/defs.ts
ukstv: IMO, DB enums are a pain to maintain/migrate.

apps/api/src/database/schema/indexing/defs.ts
ukstv: It does, for a purpose. Joins are a bit more involved that just a single loop through the index for when you intend to show history for a user.

apps/api/src/database/schema/indexing/defs.ts
ukstv: Please, no. Rationale: https://github.com/recallnet/js-recall/pull/1132#discussion_r2310567696

> IMO, DB enums are a pain to maintain/migrate.

apps/api/src/database/schema/indexing/defs.ts
ukstv: As you wish. Changed.

apps/api/src/database/schema/voting/types.ts
ukstv: It's trivial to add them, but there are no callers yet :)

apps/api/src/database/repositories/boost.repository.ts
ukstv: `boostBalances` table is the current boost balance per user. 

The logic in increase works like this:
1. Insert into the change log (boost_changes).
  - If this succeeds → it’s the first time we’ve seen this (wallet, idemKey) pair, and we return non-empty `change`.
  - If it fails (duplicate) → we’ve already recorded this exact logical operation, and `change` variable is `undefined`.
2. If the change was new:
  - Upsert into boost_balances:
    - If the wallet doesn’t exist, create a row with the initial balance.
    - If it does exist, increment its balance (balance = balance + amount).
3. If the change already existed:
  - We’ve already applied this idemKey in the past, so we do not touch the balance again.
  - Just return the current balance as-is.

I suppose the mental trick is about idempotency key `idemKey`. A user can click multiple times on the same button which leads to multiple concurrent requests to increase/decrease the balance. As long as the user clicks have the same `idemKey`, we'd only apply a single operation.

apps/api/src/database/schema/indexing/defs.ts
asutula: Drizzle does it for you ;)

ukstv: I do not actually trust Drizzle that much :) It has been proven stupid.

ukstv:
Closed in favor of https://github.com/recallnet/js-recall/pull/1147

PR #1133, opened 2025-08-28
fix: ensure agent table rankings exclude inactive agents 

stephhuynh18:
## Problem
Agent table rankings were inconsistent with leaderboard rankings. The leaderboard correctly excluded inactive/disqualified agents from total counts, but the agent table included all agents regardless of status.

## Solution
Updated `getAgentRankingsInCompetitionsImpl` to filter by active status, matching the leaderboard logic in `getLatestPortfolioSnapshotsImpl`.

## Changes
- Modified ranking query to exclude inactive/disqualified agents
- Added E2E test verifying consistent behavior after agent disqualification

## Testing
✅ E2E test passes - active agents now see consistent rankings between leaderboard and agent table


dtbuchholz (approved):
lgtm

PR #1134, opened 2025-08-28
feat: add rewards calculation package

brunocalza:
# Context

Closes [APP-139](https://linear.app/recall-labs/issue/APP-139/rewards-implement-calculations)
Closes [APP-322](https://linear.app/recall-labs/issue/APP-322/rewards-implement-rewards-calculations-for-competitors)

This PR adds a standalone `rewards` package that implements the [spec](https://www.notion.so/Simplified-Rewards-Mechanism-for-TGE-25bdfc9427de80eba580ff11e97edaa8?d=6bedfc9427de83c3a7fc8358628c90d2).

This PR is not about:
- Building a merkle tree. That is done inside [`apps/api/src/services/competition-reward.service.ts`](https://github.com/recallnet/js-recall/blob/main/apps/api/src/services/competition-reward.service.ts). That service will use this package to calculate rewards before building the tree.
- Storing merkle roots in smart contracts

# Details

- A [`calculateRewards`](https://github.com/recallnet/js-recall/blob/main/apps/api/src/services/competition-reward.service.ts) function is exposed, with the following parameters:
  - **prizePool** - The total prize pool amount to distribute in WEI
  - **boostAllocations** - Array of boost allocations from users
  - **leaderBoard** - Ordered list of competitors (winner first)
  - **window** - Time window for boost allocations
  - **prizePoolDecayRate** - Decay rate for prize pool distribution (defaults to PrizePoolDecayRate)
  - **boostTimeDecayRate** - Decay rate for boost time calculations (defaults to BoostTimeDecayRate)
  - **hook** -  Optional function to be called to inspect the calculations
- Three helper functions are used to support `calculateRewards`
  - **splitIntoDailyIntervals** - Split the boot allocation window into 1 day intervals, so we can apply a time-decay
  - **splitPrizePool** - Split the prize pool according to a geometric decay
  - **makeBoostDecayFn** - Calculates the effective boost weight given a boost allocation timestamp
 - Helper scripts are added:
   - **verify.ts** - given a test scenario, it runs the calculations and prints all intermediate calculations. It can be used to debug and verify calculations
   - **generate.ts** - generate a test scenario. It is used to generate scenarios with high user count to test performance. It is used together with **verify.ts**
 - Unit tests are added
 - A spec.md is added
 - A performance test with 300k users was executed
   -  It took 8s and consumed 800MB of memory 
 - ~A ISSUES.md file is added with potential problems with current implementation~
 
 # What is expected from review (besides regular review stuff) ?
 - Does the implementation conform to the spec?
 - Any major flaws or risks that are not being captured in ISSUES.md?
 
 # Follow-up work
 - Integrate this package inside  [`apps/api/src/services/competition-reward.service.ts`](https://github.com/recallnet/js-recall/blob/main/apps/api/src/services/competition-reward.service.ts)


packages/rewards/ISSUES.md
```diff
@@ -0,0 +1,79 @@
...
+- **Mitigation**: Define tie-breaking rules
+
+### 3. **Reward Distribution Completeness**
+
+- **Issue**: The spec mentions that total payouts may not equal the prize pool due to time decay
+- **Risk**: Users might not receive the full prize pool
+- **Impact**: User dissatisfaction
+- **Mitigation**: Document this behavior clearly and consider redistribution strategies
+
+### 4. **No Audit Trail**
```
brunocalza: this is not true anymore there's `verify.ts` script that you can pass a test case (e.g. examples/test-case.json) and verify intermediary steps using [hooks](https://github.com/recallnet/js-recall/pull/1134/files#diff-133466dc0a999c4dfa6b42ca8af662f2af0f563719867324ef01a1f76467cd04R40) in `calculateRewards`

packages/rewards/ISSUES.md
```diff
@@ -0,0 +1,79 @@
+# Analysis
```
brunocalza: This is a report with potential issues we have with this implementation. Most of them are probably fine and can be dealt with later

packages/rewards/ISSUES.md
```diff
@@ -0,0 +1,79 @@
...
+- **Mitigation**: Add bounds checking and fallback calculations
+
+### 2. **Tie Handling**
+
+- **Issue**: No logic for handling tied competitors
+- **Risk**: Undefined behavior when competitors have equal scores
+- **Impact**: Unpredictable results
+- **Mitigation**: Define tie-breaking rules
+
+### 3. **Reward Distribution Completeness**
```
brunocalza: If no one boosts a competitor, we may redistribute its prize pool to other competitors. I did no implement this because it's more complex. But def an improvement

packages/rewards/ISSUES.md
```diff
@@ -0,0 +1,79 @@
...
+## Issues
+
+### 1. **Mathematical Formula Edge Cases**
+
+- **Issue**: The decay formula `w_i = ((1-r) * r^(i-1)) / (1-r^k)` can have numerical instability
+- **Risk**: When `r` approaches 1 or 0, or when `k` is very large, the formula can produce unexpected results
+- **Impact**: Could break reward distribution entirely
+- **Mitigation**: Add bounds checking and fallback calculations
+
+### 2. **Tie Handling**
```
brunocalza: I can't remember if ties are possible in our setup. If so, we need to fix this

packages/rewards/ISSUES.md
```diff
@@ -0,0 +1,79 @@
...
+- **Mitigation**: Document this behavior clearly and consider redistribution strategies
+
+### 4. **No Audit Trail**
+
+- **Issue**: No logging or verification of calculations
+- **Risk**: Difficult to debug or verify results
+- **Impact**: Trust and debugging issues
+- **Mitigation**: Add comprehensive logging
+
+### 5. **Deterministic but Unverifiable**
```
brunocalza: same as https://github.com/recallnet/js-recall/pull/1134/files#r2308340044

packages/rewards/ISSUES.md
```diff
@@ -0,0 +1,79 @@
...
+
+- **Issue**: While calculations are deterministic, there's no way to verify them externally
+- **Risk**: Users can't independently verify their rewards
+- **Impact**: Trust issues
+- **Mitigation**: Provide verification tools or APIs
+
+## Performance and Scalability Issues
+
+### 6. **O(n²) Algorithm Complexity**
```
brunocalza: it's more a O(m * n), where m: users, n: competitors. I don't think it's a big issue right now

packages/rewards/ISSUES.md
```diff
@@ -0,0 +1,79 @@
...
+
+- **Issue**: The main calculation loops through all users for each competitor
+- **Risk**: Performance degrades quadratically with number of users/competitors
+- **Impact**: Could become slow with large competitions
+- **Mitigation**: Optimize algorithm or add performance monitoring
+
+### 7. **Memory Usage**
```
brunocalza: I have measured the memory usage to calculate rewards for 300k users, it takes roughly 800MB. It's quite a lot, but we can handle it. But it is definitely something to be aware of.

packages/rewards/ISSUES.md
```diff
@@ -0,0 +1,79 @@
...
+- **Mitigation**: Stream processing or memory-efficient data structures
+
+### 8. **Decimal.js Memory Overhead**
+
+- **Issue**: Each calculation creates many `Decimal` objects
+- **Risk**: High memory usage and potential garbage collection issues
+- **Impact**: Performance degradation
+- **Mitigation**: Object pooling or alternative decimal implementation
+
+### 9. **No Performance Tests**
```
brunocalza: I did a manual perfomance test with 300k users. It took 8s to calculate the rewards. I think it's good enough. Be aware that this will increase when we add the merkle tree building and database storage. Let's see.

packages/rewards/examples/test-case.json
```diff
@@ -0,0 +1,70 @@
+{
```
brunocalza, line 1: this is the test case for spec.md. you can run

```
pnpm verify ../examples/test-case.json
```

packages/rewards/spec.md
```diff
@@ -0,0 +1,373 @@
+# Model
```
brunocalza: This is the formal spec

packages/rewards/src/index.test.ts
```diff
@@ -0,0 +1,137 @@
+import { describe, expect, it } from "vitest";
+
+import { calculateRewards } from "./index.js";
+import type {
+  BoostAllocation,
+  BoostAllocationWindow,
+  Leaderboard,
+} from "./types.js";
+
+describe("calculateRewards", () => {
```
brunocalza: here we test the same [example](https://github.com/recallnet/js-recall/blob/ffb4879757a18cd0093a92d44d17e704cb63858a/packages/rewards/spec.md#full-example) as of spec.md

carsonfarmer:
Hey @brunocalza, I'm preparing to review this in the am. Based on recent lack conversations, would you say this is still up to date and ready for review?

packages/rewards/ISSUES.md
joewagner: I believe ties are possible

packages/rewards/ISSUES.md
joewagner: This is nice for review purpose. It seems problematic to commit this the the main branch, since the LLMs/Agents will read this and assume everything here is true.  This will affect how they work on the code in the future, and it will require keeping this document in sync with any changes to the code base.

packages/rewards/scripts/generate.ts
```diff
@@ -0,0 +1,456 @@
...
+ * Display usage information
+ */
+function showUsage(): void {
+  console.log(`
+Usage: tsx scripts/generate.ts [options]
+
+Options:
+  --competitors <number>     Number of competitors (default: 3)
+  --users <number>          Number of users (default: 5)
+  --prize-pool <string>     Prize pool amount in wei (default: 1000000000000000000000)
```
joewagner, line 393: How much wei in a token?  I'm assuming this unit is 1/10^18 tokens, is that right?

packages/rewards/scripts/verify.ts
```diff
@@ -0,0 +1,264 @@
...
+    // Get filename from command line arguments
+    const args = process.argv.slice(2);
+    if (args.length === 0) {
+      console.error("Usage: node verify.ts <filename>");
+      console.error("Example: node verify.ts ../examples/sample-data.json");
+      process.exit(1);
+    }
+
+    const filename = args[0]!; // Non-null assertion since we check length above
+    const sampleDataPath = join(__dirname, filename);
```
joewagner: This requires the path to be relative to _this_ file.  Maybe overly opinionated, but it seems more intuitive to make the path relative to the current directory where the command is run.
i.e. use `process.cwd()` instead of `__dirname`.

This has the benefit of allowing tab autocomplete for file names, etc...

packages/rewards/spec.md
```diff
@@ -0,0 +1,373 @@
...
+| 1      | 0.50195 | 501.95     |
+| 2      | 0.25098 | 250.98     |
+| 3      | 0.12549 | 125.49     |
+| 4      | 0.06274 | 62.74      |
+| 5      | 0.03122 | 31.22      |
+| 6      | 0.01562 | 15.62      |
+| 7      | 0.00781 | 7.81       |
+| 8      | 0.00391 | 3.91       |
+| 9      | 0.00195 | 1.95       |
+| 10     | 0.00098 | 0.98       |
```
joewagner, line 41: I'm not sure if this is just written wrong, or maybe there is something wrong with the calculation, but these values add up to `1002.65` which is more than the total rewards available.

packages/rewards/ISSUES.md
joewagner: Maybe we can move these to fixme or todo comments?

packages/rewards/ISSUES.md
```diff
@@ -0,0 +1,79 @@
+# Analysis
+
+This document outlines all potential issues with the `@recallnet/rewards` package implementation.
+
+## Issues
+
+### 1. **Mathematical Formula Edge Cases**
+
+- **Issue**: The decay formula `w_i = ((1-r) * r^(i-1)) / (1-r^k)` can have numerical instability
+- **Risk**: When `r` approaches 1 or 0, or when `k` is very large, the formula can produce unexpected results
```
joewagner: Maybe add some checks to ensure that `r` is between 0.9 and 0.1,  and `k` is less than 100?

packages/rewards/ISSUES.md
carsonfarmer: In theory, it is possible to have ties. However, if we use raw scores rather than just rank placement, the openskill library is able to handle this for us IIRC.

carsonfarmer: For now (in this initial version), we'll probably want to just split the pot at that tranche between the competitors. So we'll want a ticket or todo for that.

carsonfarmer: For computing the updated recall rank, this should not be an issue.

packages/rewards/ISSUES.md
carsonfarmer: I think this is a feature, not a bug. If it comes up later, or if we decide we want to redistribute the whole pot each time, we'll come back and update as a whole new feature.

packages/rewards/ISSUES.md
carsonfarmer: So related to @joewagner's comment, perhaps we just move these possible issues into TODO and FIXME comments in the code where appropriate, and then just don't commit this file?

packages/rewards/ISSUES.md
carsonfarmer: There you go. Let's just drop these then.

packages/rewards/ISSUES.md
carsonfarmer: I think this is fine for now. This isn't going to be our bottleneck for a while at least.

packages/rewards/ISSUES.md
carsonfarmer: Yeh, we're going to have plenty of ram for these calcs.

packages/rewards/ISSUES.md
carsonfarmer: Should be fine. For now this is going to run once a week.

packages/rewards/ISSUES.md
```diff
@@ -0,0 +1,79 @@
...
+### 9. **No Performance Tests**
+
+- **Issue**: No tests for large-scale scenarios
+- **Risk**: Performance issues discovered only in production
+- **Impact**: Production performance problems
+- **Mitigation**: Add performance benchmarks
+
+## Low Impact Issues
+
+### 10. **Leap Year and DST Issues**
```
carsonfarmer: Maybe we stick in a TODO or FIXME comment, but I'm not at all worried about this for now.

packages/rewards/examples/test-case.json
joewagner: Add a test case for a tie?

carsonfarmer: If we decide to deal with ties, we should probably add a separate test case or add a tie in here?

packages/rewards/scripts/generate.ts
```diff
@@ -0,0 +1,456 @@
+#!/usr/bin/env node
```
carsonfarmer, line 1: This (and the related verify.ts) file are awesome for proving things out and testing manually. But it would be really nice to have an integration test included, within a tests folder. That way, we can run the unit and integration tests in CI.

Thoughts on how much work it would be to add this?

packages/rewards/scripts/verify.ts
```diff
@@ -0,0 +1,264 @@
+#!/usr/bin/env node
+import { readFileSync } from "fs";
+import { dirname, join } from "path";
+import { fileURLToPath } from "url";
```
carsonfarmer:
```suggestion
import { readFileSync } from "node:fs";
import { dirname, join } from "node:path";
import { fileURLToPath } from "node:url";
```

packages/rewards/scripts/generate.ts
```diff
@@ -0,0 +1,456 @@
...
+import { writeFileSync } from "fs";
+import { dirname, join } from "path";
+import { fileURLToPath } from "url";
```
carsonfarmer:
```suggestion
import { writeFileSync } from "node:fs";
import { dirname, join } from "node:path";
import { fileURLToPath } from "node:url";
```

packages/rewards/scripts/generate.ts
```diff
@@ -0,0 +1,456 @@
...
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
```
joewagner: Same as below, I think you might want to use cwd?

packages/rewards/src/helpers.ts
```diff
@@ -0,0 +1,87 @@
+import Decimal from "decimal.js-light";
```
carsonfarmer, line 1: Do we strictly need this? Just for more stable tests?

packages/rewards/spec.md
joewagner: If we don't already have a test for this, but if not we should have one.
I will say I've been manually using the generate and verify scripts and the values always add up as expected.

joewagner: FYI: Carson recalculated using the spec in Notion and it looks like these numbers are wrong, and the formula is correct.

carsonfarmer:
```suggestion
| 1      | 0.50049 | 500.49     |
| 2      | 0.25024 | 250.24     |
| 3      | 0.12512 | 125.12     |
| 4      | 0.06256 | 62.56      |
| 5      | 0.03128 | 31.28      |
| 6      | 0.01564 | 15.64      |
| 7      | 0.00782 | 7.82       |
| 8      | 0.00391 | 3.91       |
| 9      | 0.00196 | 1.96       |
| 10     | 0.00098 | 0.98       |
```

packages/rewards/spec.md
```diff
@@ -0,0 +1,373 @@
...
+| 7          | 1      | 1      | 0      | 2           |
+| 8          | 0      | 0      | 1      | 1           |
+| 9          | 0      | 1      | 0      | 1           |
+| 10         | 0      | 0      | 1      | 1           |
+
+Payouts per placement
+
+| Rank | Pool   | User A | User B | User C |
+| ---- | ------ | ------ | ------ | ------ |
+| 1    | 501.95 | 150.6  | 250.98 | 100.39 |
```
joewagner: Another note here: 
`150.6  + 250.98 + 100.39 = 501.97` so there's two cents more being paid out than there is in the pool.  It seems like a lot of this document might have small deviations from correct values, which would make sense if llms are trying to do the math.

packages/rewards/src/helpers.test.ts
```diff
@@ -0,0 +1,737 @@
...
+      end: new Date("2024-01-03T00:00:00Z"),
+    };
+    const decayRate = 0.5;
+    const boostDecayFn = makeBoostDecayFn(window, decayRate);
+
+    const testCases = [
+      {
+        name: "timestamp before window start should return 0",
+        timestamp: new Date("2023-12-31T23:59:59Z"),
+        expected: new Decimal(0),
```
joewagner, line 435: I'm not sure how this works yet, but in practice we are going to give anyone who boosts before the offical start of the competition full value for their reward.  
My intuition is that this means the boost decay function would return `1` for a tiemstamp before the start date.  Maybe I'm not understanding how this function is used?

joewagner (commented):
@brunocalza Awesome work here!  I added a handful of coments and questions.  
Carson and I went through this in a call, so I think added comments too.
Happy to discuss in a call or in the comments.

packages/rewards/ISSUES.md
brunocalza: yeah, good call out. i'll just keep them in a notion doc

packages/rewards/examples/test-case.json
brunocalza: yeah definitely

packages/rewards/scripts/generate.ts
brunocalza: yeah, 10^18

packages/rewards/scripts/verify.ts
brunocalza: this is annoying indeed. thanks for the suggestion. will change it

packages/rewards/spec.md
brunocalza: Yeah thanks for checking that. I think I did not use any decimal library to calculate this, and got some errors

packages/rewards/src/helpers.test.ts
brunocalza: The window is actually the voting window and not the competition window. So we would never get a timestamp before the start date, at least we shouldn't. In case we do get, the function returns `0` actually ( not `1`), there's a check if the timestamp does not fall into the interval

```
if (index === -1) {
  return new Decimal(0);
}
 ```

joewagner: Ok, that makes sense.  Just need to make sure we allow for full value rewards if people boost anytime before the start of the competition.  In practice I think that means the caller of this function is going to have to normalize timestamps so that all dates before the comp start are set to the exact start of the window. Which won't be a problem, just calling it out here.

brunocalza: hum, you get a full value if you vote in the first day of the voting window, even if before competition. but if the voting window spans more than a day before the competition starts, you don't get a full value if you vote on second day. 

Are you suggesting that independently of how many days, if you vote before competition starts, you should get full value? That is a slight modification of the spec, but can be done, I guess. I would need to know when the competition starts as well.  wdyt @carsonfarmer ?

carsonfarmer (changes requested):
I'm mostly done with my review. Will submit now so as not to lose any comments!

packages/rewards/src/helpers.test.ts
carsonfarmer: I think we should keep it "as is". We have a voting window. This window is technically independent of the competition start date/time. However, in practice, we'll likely want to "line up" the voting window with the start of the competition, so that it all makes sense?

cc @KGmajor to make sure this is how you are imagining things?

KGmajor: Right now, we actually start voting on the day before a competition begins. The competition API has specific fields for the competition start and also the voting window. I _think_ we'd want to keep maximizing voting time so we can get as many people participating as possible.

packages/rewards/ISSUES.md
brunocalza: sounds good! I'll update the spec and the code to account for ties

packages/rewards/scripts/generate.ts
brunocalza: i'll have an integrate test in the comps app when all pieces are connected together: merkle tree, database, smart contract, and rewards calculations

packages/rewards/src/helpers.ts
brunocalza: I'd say yes. If we use `number` things get can non-deterministic. I think there's a way of doing these calculations using `bigint`, but that requires a bit more thought

joewagner (approved):
Great work!

carsonfarmer (approved):
These changes look great @brunocalza.

PR #1136, opened 2025-08-28
Isolate db schema code

asutula:
In preparation for implementing parts of our API in another app, to start with, we need to make the Drizzle schema shared code in its own package. This PR doesn't do that yet, but it untangles all the code in `apps/api/src/database/schema` from all other code in `apps/api`. This will allow us to move it to a new package in the next PR. In the process of doing this, it was clear there were places where the db schema should be the source of truth (mostly with enums that are used in other places in the app), so I refactored a bit to make it that way.

### DIFF
--- a/apps/api/scripts/backfill-stats.ts
+++ b/apps/api/scripts/backfill-stats.ts
@@ -8,7 +8,6 @@ import {
   getBulkAgentCompetitionRankings,
 } from "@/database/repositories/competition-repository.js";
 import { ServiceRegistry } from "@/services/index.js";
-import { COMPETITION_AGENT_STATUS, COMPETITION_STATUS } from "@/types/index.js";
 
 const services = new ServiceRegistry();
 
@@ -79,7 +78,7 @@ async function calculateBulkAgentStats(
         agentId,
       );
 
-      if (agent?.status !== COMPETITION_AGENT_STATUS.ACTIVE) {
+      if (agent?.status !== "active") {
         console.log(`Agent ${agentId} was deactivated, skipping`);
         continue;
       }
@@ -155,9 +154,7 @@ async function backfillCompetitionPnl() {
     );
 
     const competitions = await services.competitionManager.getAllCompetitions();
-    const endedCompetitions = competitions.filter(
-      (c) => c.status === COMPETITION_STATUS.ENDED,
-    );
+    const endedCompetitions = competitions.filter((c) => c.status === "ended");
     console.log(
       `${colors.magenta}----------------------------------------${colors.reset}`,
     );
--- a/apps/api/src/database/schema/core/defs.ts
+++ b/apps/api/src/database/schema/core/defs.ts
@@ -16,42 +16,40 @@ import {
   varchar,
 } from "drizzle-orm/pg-core";
 
-import {
-  ACTOR_STATUS_VALUES,
-  COMPETITION_AGENT_STATUS_VALUES,
-  COMPETITION_STATUS_VALUES,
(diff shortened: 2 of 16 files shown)


apps/api/scripts/backfill-stats.ts
```diff
@@ -8,7 +8,6 @@ import {
   getBulkAgentCompetitionRankings,
 } from "@/database/repositories/competition-repository.js";
 import { ServiceRegistry } from "@/services/index.js";
-import { COMPETITION_AGENT_STATUS, COMPETITION_STATUS } from "@/types/index.js";
```
asutula, line 11: These objects that mirror enums we've defined in Drizzle and Zod aren't needed because Typescript properly types things as a union of string literals.

apps/api/scripts/backfill-stats.ts
```diff
@@ -79,7 +78,7 @@ async function calculateBulkAgentStats(
         agentId,
       );
 
-      if (agent?.status !== COMPETITION_AGENT_STATUS.ACTIVE) {
+      if (agent?.status !== "active") {
```
asutula, line 81: You can just use these string literals in a typesafe way.

apps/api/src/database/schema/core/defs.ts
```diff
@@ -16,42 +16,40 @@ import {
...
 /**
  * Statuses for users, agents, and admins.
  */
-export const actorStatus = pgEnum("actor_status", ACTOR_STATUS_VALUES);
+export const actorStatus = pgEnum("actor_status", [
+  "active",
+  "inactive",
+  "suspended",
+  "deleted",
+]);
```
asutula, line 29: The database schema should be the source of truth for enums like this. Other places in the code that need the enum values can import them from here.

apps/api/src/database/schema/core/defs.ts
```diff
@@ -16,42 +16,40 @@ import {
   varchar,
 } from "drizzle-orm/pg-core";
 
-import {
-  ACTOR_STATUS_VALUES,
-  COMPETITION_AGENT_STATUS_VALUES,
-  COMPETITION_STATUS_VALUES,
-  COMPETITION_TYPE_VALUES,
-  MAX_HANDLE_LENGTH,
-} from "@/types/index.js";
```
asutula, line 25: Before this, the db schema was depending on types code in the api app to know what enum values to use. Now we're reversing that relationship.

apps/api/src/types/index.ts
```diff
@@ -368,29 +377,11 @@ export interface Admin {
...
-
-/**
- * Cross-chain trading types values.
- */
-export const CROSS_CHAIN_TRADING_TYPE = {
-  DISALLOW_ALL: "disallowAll",
-  DISALLOW_X_PARENT: "disallowXParent",
-  ALLOW: "allow",
-} as const;
-
```
asutula, line 388: Since the database schema files are now the source of truth, and we can use string literal unions all over the code, we don't need these and other definitions like them.

apps/api/src/types/index.ts
```diff
@@ -368,29 +377,11 @@ export interface Admin {
...
 /**
  * Zod schema for the cross-chain trading type.
  */
 export const CrossChainTradingTypeSchema = z.enum(
-  CROSS_CHAIN_TRADING_TYPE_VALUES,
+  crossChainTradingType.enumValues,
```
asutula, line 384: Use the enum defined in the database schema layer instead.

PR #1137, opened 2025-08-28
fix(comps): safari table height virtualization bug

derrekcoleman:
Fixes an issue with Safari's handling of Tanstack's virtualizer that resulted in TableRows rendering with `0` height.

Minimum solution is just deleting
```
position: "absolute",
transform: `translateY(${virtualRow.start}px)`,
```
The other deletions seemed natural to include in my testing.

⚠️ **Please review that these changes don't have unintended consequences.** 
I'm not very familiar with this library or CSS, so please consider edge-cases the deleted styles could have affected.

Before:
Screenshot 2025-08-28 at 4 15 11 PM

After:
Screenshot 2025-08-28 at 4 14 48 PM


carsonfarmer:
Ugh, I really want that forked data thing that neon provides.

derrekcoleman:
@carsonfarmer, my local workaround for that is to add `/api/` in apps/comps/next.config.mjs:
```
          destination: `${process.env.NEXT_PUBLIC_API_BASE_URL}/api/:path*`,
```
(your version is the same, without `/api/` near EOL)

If I had more confidence in the setup, I'd say that's just a fix to push for everyone.

ramirotw (approved):
tested locally, works fine. 

The virtualizer was added to support infinite scrolling, which was how the old designs handled loading more rows. Now that we’re using standard pagination, this extra complexity is no longer needed

derrekcoleman:
Thanks, Ramiro!

PR #1140, opened 2025-08-29
refactor(api): remove full scan & calculate summary metrics

dtbuchholz:
The `/competitions/:id` endpoint is currently doing a full scan of the `trades` table because we don't pass a limit to the method `services.tradeSimulator.getCompetitionTrades(competitionId)`. We only make that call because we need summary metrics (total trades, total volume, and # of unique tokens).

This PR adds a new method to simply calculate those metrics instead of doing "query all + in-memory computation"/

### DIFF
--- a/apps/api/src/database/repositories/trade-repository.ts
+++ b/apps/api/src/database/repositories/trade-repository.ts
@@ -160,16 +160,16 @@ async function getCompetitionTradeMetricsImpl(competitionId: string) {
         unnest(ARRAY[b.from_token, b.to_token]) WITH ORDINALITY AS u(token, ord);
     `);
 
-    const r = query.rows[0] ?? {
+    const result = query.rows[0] ?? {
       total_trades: 0,
       total_volume: 0,
       unique_tokens: 0,
     };
 
     return {
-      totalTrades: Number(r.total_trades), // bigint → number (OK unless > 2^53)
-      totalVolume: Number(r.total_volume), // float8 already a number in many drivers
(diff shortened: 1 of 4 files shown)


apps/api/src/database/repositories/trade-repository.ts
```diff
@@ -134,6 +134,48 @@ async function getAgentTradesImpl(
...
 }
 
+/**
+ * Get the count of trades for a competition (count, total volume, unique tokens)
+ * @param competitionId Competition ID
+ * @returns Count of trades
+ */
+async function getCompetitionTradeMetricsImpl(competitionId: string) {
+  try {
+    const query = await db.execute(sql`
```
dtbuchholz, line 144: note: my original query was this, which LLMs said wasnt as effiecient as the refactored version:
```ts
const [row] = await db
  .select({
    totalTrades: sql<number>`COUNT(*)`,
    totalVolume: sql<number>`COALESCE(SUM(${trades.tradeAmountUsd}), 0)`,
    uniqueTokens: sql<number>`
  (
    SELECT COUNT(*) FROM (
      SELECT ${trades.fromToken} AS token
      FROM ${trades}
      WHERE ${trades.competitionId} = ${competitionId}
      UNION
      SELECT ${trades.toToken} AS token
      FROM ${trades}
      WHERE ${trades.competitionId} = ${competitionId}
    ) t
  )
`,
  })
  .from(trades)
  .where(eq(trades.competitionId, competitionId))
  .limit(1);

```

PR #1144, opened 2025-09-02
Boost API

asutula:


apps/api/src/services/index.ts
```diff
@@ -126,12 +128,13 @@ class ServiceRegistry {
...
     );
+    this._userRepository = new UserRepository(db);
 
     // Initialize BoostService with its dependencies
     this._boostService = new BoostService(
       this._boostRepository,
-      this._competitionService,
-      this._userService,
+      this._competitionRepository,
+      this._userRepository,
```
stbrody, line 142: can we add a TODO comment to put this back to being the services?  I think as a general rule services should be able to call into other services, but not other repositories other than the one it "owns"

apps/comps/app/rpc/[[...rest]]/route.ts
```diff
@@ -0,0 +1,30 @@
...
+      privyClient,
+      boostService,
+      db,
+    },
+  });
+
+  return response ?? new Response("Not found", { status: 404 });
+}
+
+export const HEAD = handleRequest;
```
Comment (stbrody) [inline apps/comps/app/rpc/[[...rest]]/route.ts:25]: does the request handler handle these different HTTP methods differently internally?  How is the HTTP method specified now?  Is it within the Request object?

apps/comps/lib/repositories.ts
```diff
@@ -0,0 +1,9 @@
+import { BoostRepository } from "@recallnet/db/repositories/boost";
+import { CompetitionRepository } from "@recallnet/db/repositories/competition";
+import { UserRepository } from "@recallnet/db/repositories/user";
+
+import { db } from "./db";
+
+export const boostRepository = new BoostRepository(db);
+export const competitionRepository = new CompetitionRepository(db, db);
```
stbrody: Is this losing the ability to route reads to the read replica by using the same db handle in both args here?  That seems like a potential regression we'd want to avoid?

apps/comps/rpc/clients/client-side.ts
```diff
@@ -0,0 +1,25 @@
...
+  createSafeClient,
+  onError,
+} from "@orpc/client";
+import { RPCLink } from "@orpc/client/fetch";
+import { RouterClient } from "@orpc/server";
+
+import { router } from "@/rpc/router";
+
+const link = new RPCLink({
+  url: "http://localhost:3001/rpc",
```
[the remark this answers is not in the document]
stbrody: This does seem like it needs to be fixed before we can merge this

apps/comps/rpc/clients/client-side.ts
```diff
@@ -0,0 +1,25 @@
...
+  headers: () => ({}),
+  // fetch: <-- provide fetch polyfill fetch if needed
```
stbrody: what does this comment mean?

apps/comps/rpc/clients/client-side.ts
```diff
@@ -0,0 +1,25 @@
...
+  interceptors: [
+    onError((error) => {
+      console.error(error);
+    }),
+  ],
+});
+
+// Create a client for your router
+export const client: RouterClient<typeof router> = createORPCClient(link);
```
stbrody, line 23: I don't fully understand the difference between these two, but do we want the un-safe client exported?  Or should we only export the safe one?

apps/comps/rpc/context/base.ts
```diff
@@ -0,0 +1,23 @@
+import { os } from "@orpc/server";
+import { PrivyClient } from "@privy-io/node";
+import { cookies } from "next/headers";
+
+import { Database } from "@recallnet/db/types";
+import { BoostService } from "@recallnet/services/boost";
+
+export const base = os
```
stbrody, line 34: some comments explaining what this doing and how to use/maintain it might be nice?  Probably for a lot of this new plumbing to be honest

apps/comps/rpc/middleware/user.ts
```diff
@@ -0,0 +1,31 @@
...
+    const userId = context.privyUser?.id;
+    if (!userId) {
+      return await next({
+        context: {
+          user: undefined,
+        },
+      });
+    }
+
+    const user = await context.db.query.users.findFirst({
```
stbrody, line 24: Lets add a TODO comment to update this to use the userService. As-is we're going to lose some of the caching around user objects (unless nextjs handles this automagically for us?).

apps/comps/rpc/router/boost/balance.ts
```diff
@@ -0,0 +1,29 @@
+import { z } from "zod";
+
+import { BoostError, type BoostService } from "@recallnet/services/boost";
```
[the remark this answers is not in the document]
stbrody: plus one, surprised this didn't fail the lint check actually

apps/comps/rpc/router/boost/boost-agent.ts
```diff
@@ -0,0 +1,50 @@
+import { z } from "zod";
+
+import { BoostError } from "@recallnet/services/boost";
+
+import { base } from "@/rpc/context/base";
+import { authMiddleware } from "@/rpc/middleware/auth";
+import { assertNever } from "@/rpc/router/utils/assert-never";
+import { Database } from "@/rpc/types";
```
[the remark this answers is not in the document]
stbrody: +1

apps/comps/rpc/router/boost/boost-agent.ts
```diff
@@ -0,0 +1,50 @@
...
+    OUTSIDE_BOOST_WINDOW: {},
+    COMPETITION_MISSING_VOTING_DATES: {},
+  })
+  .handler(async ({ input, context, errors }) => {
+    const res = await context.boostService.boostAgent({
+      userId: context.user.id,
+      competitionId: input.competitionId,
+      agentId: input.agentId,
+      amount: input.amount,
+      idemKey: Buffer.from(input.idemKey, "base64"),
```
stbrody: nitpick, but you should be able to do this transformation with zod in the parsing phase

packages/db/src/repositories/competition.ts
```diff
@@ -105,10 +105,10 @@ const competitionOrderByFields: Record<string, AnyColumn> = {
 export class CompetitionRepository {
   readonly #db: Database;
   readonly #dbRead: Database;
-  readonly #logger: Logger;
+  readonly #logger?: Logger;
   readonly #snapshotCache;
 
-  constructor(database: Database, readDatabase: Database, logger: Logger) {
+  constructor(database: Database, readDatabase: Database, logger?: Logger) {
```
stbrody: I still don't totally understand why this needs to be optional now?  Does that mean we're not providing a logger for the endpoints that migrate to serverless?  That seems like it would be bad no?

packages/services/package.json
```diff
@@ -0,0 +1,53 @@
...
+      "import": {
+        "types": "./dist/boost.d.ts",
+        "default": "./dist/boost.js"
+      }
+    }
+  },
+  "files": [
+    "dist"
+  ],
+  "scripts": {
```
stbrody, line 25: are we missing a format script for using prettier?

packages/services/src/boost.ts
```diff
@@ -0,0 +1,157 @@
...
+  private boostRepository: BoostRepository;
+  private competitionRepository: CompetitionRepository;
+  private userRepository: UserRepository;
+  private logger?: Logger;
+
+  constructor(
+    boostRepository: BoostRepository,
+    competitionRepository: CompetitionRepository,
+    userRepository: UserRepository,
+    logger?: Logger,
```
stbrody: seems like the logger should be required here too?

packages/services/src/boost.ts
```diff
@@ -0,0 +1,157 @@
...
+  OutsideCompetitionBoostWindow,
+}
+
+/**
+ * Boost Service
+ * Manages boost operations including validation and business logic
+ */
+export class BoostService {
+  private boostRepository: BoostRepository;
+  private competitionRepository: CompetitionRepository;
```
stbrody, line 29: TODO to use services

packages/services/src/boost.ts
```diff
@@ -0,0 +1,157 @@
...
+   * @param params The boost parameters
+   * @returns The result of the boost operation
+   */
+  async boostAgent(params: BoostAgentParams) {
+    const { userId, competitionId, agentId, amount, idemKey } = params;
+    return ResultAsync.fromPromise(
+      this.userRepository.findById(userId),
+      () => BoostError.RepositoryError as const,
+    )
+      .andThen((user) => {
```
stbrody, line 157: Trying to understand this new paradigm with `neverthrow`:

how would this be different if this whole `boostAgent` function was in a single `ResultAsync.fromPromise` call, without any of the `andThen` calls?  Do we lose any type safety?  The final `andThen` is quite a bit larger than the first 2, why can all that logic fit in one vs being broken up like the other 2 are?  I guess it has something to do with the first 2 doing async i/o?  Do you need one functional unit for every async call?  What happens if you just try to use `await` within the body of one of these?

apps/comps/lib/privy-client.ts
```diff
@@ -0,0 +1,6 @@
+import { PrivyClient } from "@privy-io/node";
+
+export const privyClient = new PrivyClient({
+  appId: "",
+  appSecret: "",
```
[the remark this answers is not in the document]
stbrody: addressed in the follow-up pr

apps/comps/rpc/clients/client-side.ts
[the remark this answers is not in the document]
asutula: Did it in the upstream pr.

apps/api/src/services/index.ts
asutula: Hmm, we should talk about that. I'm not sure I agree. In fact, I'd almost argue the opposite that services should only depend on repositories, any repositories they need to get their job done, but not other services. This avoids circular dependencies between services, and code that might be duplicated between services could be located in a shared location. I'll make a TODO comment like you said, but we should discuss the final solution.

apps/comps/app/rpc/[[...rest]]/route.ts
Comment (asutula) [reply apps/comps/app/rpc/[[...rest]]/route.ts:25]: Explaining this with my ~80% level of oRPC knowledge... I believe that in the default rpc mode, all requests are sent as `POST`. If you choose to leverage the OpenAPI support in oRPC, you can , in your procedure definitions, add configuration to specify what http method the procedure should use, along with other configuration options only applicable in OpenAPI mode. This allows you to mold a more REST-ish API that is OpenAPI compatible, and requires that the rpc handler here respond to all those methods. Perhaps we can remove the ones we don't need for now.

https://orpc.unnoq.com/docs/openapi/getting-started

apps/comps/lib/repositories.ts
asutula: I need to understand our read replica situation. I don't even know where it exists or how to connect to it. I'll find out and we can properly configure it and use it here.

apps/comps/rpc/clients/client-side.ts
asutula: Copy pasta from their example code. Deleted.

apps/comps/rpc/clients/client-side.ts
asutula: The unsafe client throws errors and returns whatever value upon success. The safe client catches errors internally and returns a tuple of `[value, error]`. So it just depends on the context in which you are using the client and what makes sense. Our usual use case is actually using neither of these and using the `tanstack` client from our react components. These clients might be more useful for things like writing a cli or Node app that communicates with our rpc api.

apps/comps/rpc/router/boost/boost-agent.ts
asutula: Good idea. Done.

packages/db/src/repositories/competition.ts
asutula: It shouldn't be optional, I just did that in a hurry and forgot to come back to it. Will add pino loggers to the project now.

packages/services/package.json
asutula: I saw those scripts pop up in all the packages over the time I was out. It isn't necessary. You just run format from the root of the repo.

packages/services/src/boost.ts
asutula: Yes. Doing.

packages/services/src/boost.ts
asutula: If you wrap the entire `boostAgent` function in a single `ResultAsync.fromPromise`, then you loose the ability to be granular about what errors to throw, so your type safety is less useful.

It seems that each `andThen` block should only contain one async operation. Doing other synchronous work in the block before returning the promise seems to make sense and is fine. I'm now curious about doing synchronous work that thows an error. Will look into that.

You can't use `await` inside of `andThen`. It won't let you pass an async function to `andThen`.

The wiki for the library is informative. This page especially https://github.com/supermacro/neverthrow/wiki/Working-with-ResultAsync

packages/db/src/repositories/competition.ts
asutula: Ok services are in. I created a quick pino setup in the comps app that looks very similar to the setup in the api app. I'll make an issue to refactor that code into a shared package.

packages/services/src/boost.ts
```diff
@@ -0,0 +1,188 @@
...
+  /**
+   * Boost an agent in a competition
+   * @param params The boost parameters
+   * @returns The result of the boost operation
+   */
+  async boostAgent(params: BoostAgentParams) {
+    const { userId, competitionId, agentId, amount, idemKey } = params;
+    return ResultAsync.fromPromise(
+      this.userRepository.findById(userId),
+      () => BoostError.RepositoryError as const,
```
stbrody: is this losing the error message of whatever Error might be thrown by `userRepository.findById`?  Seems like we should be able to map the error type while keeping the message passed through so we don't lose context?

packages/services/src/boost.ts
stbrody: hmm, but you can almost certainly use `await` within an async function passed to `ResultAsync.fromPromise`.

I still would be pretty curious to see what this function looks like with a single async function given to the top-level `ResultAsync.fromPromise`.  Feels like it could be easier to read that way, but would depend a bit on how the error handling winds up looking like.  Would be an interesting experiment to try.  Anyway, certainly not a blocker

packages/services/src/boost.ts
```diff
@@ -0,0 +1,188 @@
...
+    );
+  }
+
+  /**
+   * Boost an agent in a competition
+   * @param params The boost parameters
+   * @returns The result of the boost operation
+   */
+  async boostAgent(params: BoostAgentParams) {
```
stbrody: the guide you linked to on working with `ResultAsync` suggests that this function (and really all of the functions in this file now) should *not* be async

asutula: Oh good call, there should be a lint rule for that. Fixing.

packages/services/src/boost.ts
asutula: Yea, we can preserve it by using some object with a message for our error type (not this enum). I can take a quick crack at that. Really, we need to properly handle errors from the db layer. Been reading about it here and other places https://github.com/drizzle-team/drizzle-orm/discussions/916

asutula: Took a stab at this. Not sure if I love the solution or not, but we can experiment with it.

PR #1148, opened 2025-09-02
fix: ci & db-schema package publishing

dtbuchholz:
1. CI is failing to build the docker image because we explicitly include only a subset of packages—and the new db-schema package isn't included
2. we want to keep our db-schema package off of npm


smrz2001 (approved):
🚀

PR #1149, opened 2025-09-02
fix(comps): safari table height virtualization bug

derrekcoleman:
[Previous PR](https://github.com/recallnet/js-recall/pull/1137) should have been against `prod-latest` branch, which we use for front-end deployments, not `main`.


carsonfarmer (approved):
Based on the previous PR, I'm happy to 👍 this one. However, we should have a plan for dealing with the weird history if we're doing two separate PRs.

PR #1154, opened 2025-09-03
fix: used old leaderboard stats

andrewxhill:


dtbuchholz:
note: the new JSON doesnt lead to the right behavior in the UI:
- the score bars aren't filled, and the `0` value should be rendered as `Top: <score>`
Screenshot 2025-09-03 at 3 01 09 PM
- when you click it, the summary bar isnt populated correctly
Screenshot 2025-09-03 at 3 00 50 PM

mzkrasner (dismissed):
lgtm! Spot checked a few - appears to match against the files you shared in slack.

Edit - just saw @dtbuchholz's note about how it behaves in UI. Investigating..

PR #1163, opened 2025-09-06
chore(comps): remove external address-utils package

dtbuchholz:
- We want to remove the `@recallnet/address-utils` package entirely since [it's only 1 method](https://github.com/recallnet/js-recall/blob/main/packages/address-utils/src/display.ts)
- Thus, this brings the method into the `app/comps` dir and cleans up the external dep


carsonfarmer (approved):
Very nice!

PR #1166, opened 2025-09-06
test(api): refactor helpers to use named params

dtbuchholz:
The test helper methods to create/start comps have ordered params that are too hard to keep track of. This makes them named params.

### DIFF
--- a/apps/api/e2e/utils/api-client.ts
+++ b/apps/api/e2e/utils/api-client.ts
@@ -417,34 +417,54 @@ export class ApiClient {
   /**
    * Create a competition in PENDING state
    */
-  async createCompetition(
-    name: string,
-    description?: string,
-    tradingType?: CrossChainTradingType,
-    sandboxMode?: boolean,
-    externalUrl?: string,
-    imageUrl?: string,
-    type?: string,
-    endDate?: string,
-    votingStartDate?: string,
-    votingEndDate?: string,
-    joinStartDate?: string,
-    joinEndDate?: string,
-    maxParticipants?: number,
-    tradingConstraints?: TradingConstraints,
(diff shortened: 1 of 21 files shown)


apps/api/e2e/utils/test-helpers.ts
```diff
@@ -217,27 +218,33 @@ export async function registerUserAndAgentAndGetClient({
...
-  externalUrl?: string,
-  imageUrl?: string,
-  tradingConstraints?: TradingConstraints,
-): Promise<StartCompetitionResponse> {
-  const result = await adminClient.startCompetition(
-    name,
-    `Test competition description for ${name}`,
-    agentIds,
-    undefined, // tradingType
+export async function startTestCompetition({
```
dtbuchholz, line 221: here / below is where we change the helper methods, and then everything above is just to conform to this.

apps/api/e2e/utils/api-client.ts
```diff
@@ -417,34 +417,54 @@ export class ApiClient {
...
-    endDate?: string,
-    votingStartDate?: string,
-    votingEndDate?: string,
-    joinStartDate?: string,
-    joinEndDate?: string,
-    maxParticipants?: number,
-    tradingConstraints?: TradingConstraints,
-    rewards?: Record<number, number>,
-  ): Promise<CreateCompetitionResponse | ErrorResponse> {
+  async createCompetition({
```
stbrody, line 420: This might be a bit cleaner with a named type, `CreateCompetitionArgs`.

dtbuchholz: yeah, fair point. (i had auto-merge enabled on the PR, so i didnt refactor it...can handle it separately.)

PR #1168, opened 2025-09-08
fix: update cursorrules detailing efficient patterns, function complexity & design

mzkrasner:
# Cursor Rules Enhancement Based on PR Feedback

## Overview

Updated organization-wide cursor rules to codify code quality patterns. These changes ensure future development naturally follows best practices for code readability, efficiency, and maintainability.

### 1. `.cursor/rules/org-general-practices.mdc`

**Added two new sections:**

#### Function Design & Complexity Management
- Extract helper methods when functions exceed ~30-40 lines or have distinct logical sections
- Single responsibility principle: Each function should do ONE thing well
- Fail-fast pattern: Check error conditions early and return immediately
- Avoid redundant checking across different functions
- Clear method contracts: Helper names should indicate their assumptions

#### Efficient Code Patterns
- Don't compute what you don't need: Avoid building intermediate data structures just to check properties
- Choose appropriate loops: Use `for` when iteration count is known, `while` for conditional iteration
- Early exit strategies: Prefer `.some()`, `.every()`, `.find()` over `.filter().length` or `.map()` when you just need a boolean
- Separation of concerns: Separate detailed logging from control flow logic

### 2. `.cursor/rules/org-typescript-standards.mdc`

**Added new section: Code Efficiency Patterns**

#### Array Method Selection
- Use `.some()` to check if ANY element matches (stops on first match)
- Use `.every()` to check if ALL elements match (stops on first non-match)
- Use `.find()` to get the first matching element (stops on first match)
- Avoid `.filter().length === 0` or `.filter().length > 0` when `.some()` or `.every()` would suffice
- Don't build arrays just to check emptiness or count

#### Type Narrowing with Early Returns
- Use early returns to narrow types and reduce nesting
- After null/undefined checks with early returns, TypeScript knows the value is defined
- Prefer guard clauses at the start of functions over deeply nested if-else blocks
- Includes example code showing the preferred pattern

#### Helper Function Extraction
- Extract complex type transformations into well-typed helper functions
- Use generics in helpers to maintain type safety
- Return discriminated unions or Result types from helpers to make error handling explicit
- Keep helper functions pure when possible (no side effects, deterministic output)

### 3. `.cursor/rules/api-specific-config.mdc`

**Added new section: Service Method Design Patterns**

- **Method Complexity:** If a service method has 3+ distinct steps (especially if labeled as "Step 1", "Step 2", etc.), consider extracting each step into a private helper method
- **Retry Logic Separation:** Extract retry logic with exponential backoff into dedicated helper methods
- **Batch Operation Helpers:** Create reusable helpers for batch operations that can be shared across methods
- **Clear Method Contracts:** Helper methods should clearly indicate what they assume about their inputs
- **Fail-Fast in Service Methods:** Validate inputs and check preconditions at the start of public methods, return early on invalid conditions

## Rationale

These additions were based on specific feedback patterns observed during code review:

1. **Long functions with numbered steps** - Now have clear guidance on when to extract helpers
2. **Redundant null checking across functions** - Now addressed with fail-fast and boundary validation patterns
3. **Building arrays just to check emptiness** - Now explicitly discouraged with better alternatives provided
4. **While loops with manual increments** - Now guided toward cleaner for-loop syntax when appropriate
5. **Complex retry logic inline with business logic** - Now directed to extract into dedicated helpers

## Impact

These rules ensure that future code contributions will:
- Be more readable and maintainable
- Avoid common inefficiencies
- Follow consistent patterns across the codebase
- Reduce the likelihood of similar PR feedback in the future

## Key Themes Now Codified

- Extract helper methods when functions have 3+ distinct steps
- Fail-fast pattern - Check errors early and return
- Avoid redundant operations - Don't build arrays just to check emptiness
- Use appropriate array methods - `.some()` instead of `.filter().length`
- Single responsibility - Each function does ONE thing
- Clear method contracts - Names indicate assumptions
- Separation of concerns - Logging vs control flow
- Choose appropriate loops - `for` when count is known

These additions complement existing rules without duplication and provide specific, actionable guidance that will help developers and AI assistants write cleaner, more efficient code.

### DIFF
--- a/.cursor/rules/api-specific-config.mdc
+++ b/.cursor/rules/api-specific-config.mdc
@@ -78,6 +78,19 @@ alwaysApply: false
 - Keep manager methods focused on single responsibilities
 - Always handle errors gracefully and return consistent result objects
 
+### Service Method Design Patterns
+- **Method Complexity:** If a service method has 3+ distinct steps (especially if labeled as "Step 1", "Step 2", etc.), consider extracting each step into a private helper method
+- **Retry Logic Separation:** Extract retry logic with exponential backoff into dedicated helper methods rather than inline in business logic
+- **Batch Operation Helpers:** Create reusable helpers for batch operations (batch fetching, batch processing) that can be shared across methods
+- **Clear Method Contracts:** Helper methods should clearly indicate what they assume about their inputs:
+  - Use descriptive names like `processWithValidatedData` vs just `processData`
+  - Document preconditions in method comments
+  - Consider using TypeScript branded types or type predicates for validated data
+- **Fail-Fast in Service Methods:** 
+  - Validate inputs and check preconditions at the start of public methods
+  - Return early on invalid conditions rather than nesting the entire method body
+  - Let helper methods assume valid inputs (since the public method already validated)
+
 ### Code Reuse & Method Implementation
 - **Before implementing any new method**: Always search the codebase for similar existing functionality using semantic search or grep
 - **If a similar method exists but is not adequate**: You must explicitly explain why the existing method cannot be used or extended, including:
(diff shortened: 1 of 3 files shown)


stbrody (approved):
nice

PR #1170, opened 2025-09-08
feat: Optimize database indices

stbrody:
Mostly adding compound indices to support sorting in various queries

### DIFF
--- a/packages/db-schema/src/core/defs.ts
+++ b/packages/db-schema/src/core/defs.ts
@@ -1,4 +1,3 @@
-import { sql } from "drizzle-orm";
 import {
   boolean,
   char,


apps/api/drizzle/0038_talented_ben_urich.sql
```diff
@@ -0,0 +1,12 @@
+-- Enable pg_trgm extension if not already enabled
+CREATE EXTENSION IF NOT EXISTS pg_trgm;--> statement-breakpoint
+DROP INDEX "idx_votes_user_competition";--> statement-breakpoint
+DROP INDEX "trading_comps"."idx_trades_agent_id";--> statement-breakpoint
+CREATE INDEX "idx_agents_name_gin" ON "agents" USING btree ("name" gin_trgm_ops);--> statement-breakpoint
```
stbrody: this one is the most interesting, it's adding a new postgres extension to support case-insensitive queries against the agent names

PR #1171, opened 2025-09-08
feat: Optimize database indices

stbrody:
Mostly adding compound indices to support sorting in various queries, as well as a special Gin index using a postgres-native extension for case-insensitive querying

### DIFF
--- a/packages/db-schema/src/core/defs.ts
+++ b/packages/db-schema/src/core/defs.ts
@@ -125,7 +125,7 @@ export const agents = pgTable(
     index("idx_agents_api_key").on(table.apiKey),
     index("idx_agents_api_key_hash").on(table.apiKeyHash),
     // NOTE: There is an extra index not listed here. There is a GIN index on the "name" that is
-    // created via custom migration (0038_talented_ben_urich) due to Drizzle limitations. That index
+    // created via custom migration (0038_left_ultragirl) due to Drizzle limitations. That index
     // supports case-insensitive queries on agent names.
     foreignKey({
       columns: [table.ownerId],


apps/api/drizzle/0038_talented_ben_urich.sql
```diff
@@ -0,0 +1,12 @@
+-- Enable pg_trgm extension if not already enabled
+CREATE EXTENSION IF NOT EXISTS pg_trgm;--> statement-breakpoint
+DROP INDEX "idx_votes_user_competition";--> statement-breakpoint
+DROP INDEX "trading_comps"."idx_trades_agent_id";--> statement-breakpoint
+CREATE INDEX IF NOT EXISTS "idx_agents_name_gin" ON "agents" USING gin ("name" gin_trgm_ops);--> statement-breakpoint
```
stbrody: this one is the most interesting, it's adding a new postgres extension to support case-insensitive queries against the agent names

dtbuchholz:
run this and commit the changes:
```
git fetch origin && git restore --source origin/main --staged --worktree drizzle/ && git clean -fd drizzle/ && pnpm db:gen-migrations
```

apps/api/drizzle/meta/0038_snapshot.json
```diff
@@ -0,0 +1,3731 @@
+{
```
stbrody, line 1: the Gin index is notably NOT included in this snapshot, because if it is then drizzle tried to drop it on every future call to `db:gen-migrations`.

Cursor seems to think this is fine and safe, but I don't know enough about how drizzle works to really know for sure if there are any risks of this approach.  Cursor claims this is standard practice though when people want to use advanced database features that drizzle doesn't natively support

asutula (approved):
Cool!

PR #1173, opened 2025-09-09
fix(api): slippage causing negative balances

dtbuchholz:
closes [APP-344](https://linear.app/recall-labs/issue/APP-344/bug-negative-prices-in-trades-table). here's where the `calculateSlippage` method is used, causing the `effectiveFromValueUSD` (used in `toAmount`) and resulting price in the trades table to be negative:
```ts
private calculateTradeAmounts(
    fromAmount: number,
    fromValueUSD: number,
    toPrice: number,
    fromToken: string,
    toToken: string,
  ): { toAmount: number; exchangeRate: number } {
    let toAmount: number;
    let exchangeRate: number;
    let effectiveFromValueUSD: number;

    if (toPrice === 0) {
      // ...
    } else {
      // Normal trade with slippage
      const { effectiveFromValueUSD: calculatedEffectiveValue } =
        calculateSlippage(fromValueUSD);
      effectiveFromValueUSD = calculatedEffectiveValue;
      toAmount = effectiveFromValueUSD / toPrice;
      exchangeRate = toAmount / fromAmount;
      serviceLogger.debug(`[TradeSimulator] Trade calculation details:
          From Token (${fromToken}):
          - Amount: ${fromAmount}
          - USD Value: $${fromValueUSD.toFixed(6)}

          To Token (${toToken}):
          - Price: $${toPrice}
          - Calculated Amount: ${toAmount.toFixed(6)}

          Exchange Rate: 1 ${fromToken} = ${exchangeRate.toFixed(6)} ${toToken}
      `);
    }

    return { toAmount, exchangeRate };
  }
```

### DIFF
--- a/apps/api/src/lib/__tests__/trade-utils.test.ts
+++ b/apps/api/src/lib/__tests__/trade-utils.test.ts
@@ -13,17 +13,17 @@ describe("trade-utils", () => {
 
     it("should apply minimal slippage for small trades", () => {
       const result = calculateSlippage(100); // $100 trade
-      expect(result.actualSlippage).toBeGreaterThan(0);
-      expect(result.actualSlippage).toBeLessThan(0.001); // Less than 0.1%
-      expect(result.effectiveFromValueUSD).toBeGreaterThan(99);
+      expect(result.actualSlippage).toBeGreaterThanOrEqual(0.0005); // At least MIN_SLIPPAGE (5 bps)
+      expect(result.actualSlippage).toBeLessThan(0.002); // Less than 0.2%
+      expect(result.effectiveFromValueUSD).toBeGreaterThan(99.8);
       expect(result.effectiveFromValueUSD).toBeLessThan(100);
     });
 
     it("should apply moderate slippage for medium trades", () => {
       const result = calculateSlippage(10000); // $10k trade
-      expect(result.actualSlippage).toBeGreaterThan(0.005); // More than 0.5%
-      expect(result.actualSlippage).toBeLessThan(0.01); // Less than 1%
-      expect(result.effectiveFromValueUSD).toBeGreaterThan(9900);
+      expect(result.actualSlippage).toBeGreaterThan(0.01); // More than 1%
+      expect(result.actualSlippage).toBeLessThan(0.04); // Less than 4%
+      expect(result.effectiveFromValueUSD).toBeGreaterThan(9600);
       expect(result.effectiveFromValueUSD).toBeLessThan(10000);
     });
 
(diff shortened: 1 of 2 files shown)


apps/api/src/lib/trade-utils.ts
```diff
@@ -16,13 +16,31 @@ export interface SlippageResult {
  * @returns SlippageResult containing slippage calculations
  */
 export function calculateSlippage(fromValueUSD: number): SlippageResult {
-  // Apply slippage based on trade size
-  const baseSlippage = (fromValueUSD / 10000) * 0.05; // 0.05% per $10,000 (10x lower than before)
-  const actualSlippage = baseSlippage * (0.9 + Math.random() * 0.2); // ±10% randomness (reduced from ±20%)
+  // Apply slippage based on trade size with logarithmic scaling
```
dtbuchholz, line 37: not sure what the ideal slippage formula is. we'll also throw an error if `effectiveFromValueUSD` is negative

stbrody (commented):
tbh I'm not great at reasoning about this kind of stuff.  This feels like something that would benefit from @carsonfarmer taking a look at

apps/api/src/lib/trade-utils.ts
```diff
@@ -16,13 +16,31 @@ export interface SlippageResult {
...
+  // This prevents slippage from exceeding reasonable limits for large trades
+  const MAX_SLIPPAGE = 0.15; // 15% maximum slippage
+
+  // Logarithmic scaling: slippage increases with trade size but at a decreasing rate
+  // Base formula: slippage = k * log(1 + value/scale)
+  // where k = 0.02 (2% base) and scale = 10000 ($10k reference)
+  const baseSlippage = Math.min(
+    0.02 * Math.log10(1 + fromValueUSD / 10000),
```
carsonfarmer: nit: We should probably use `log1p` (with base conversion) for clarity here (i.e., making it clear the intent is log(1+x)) and better behavior for tiny trades. Something like `Math.log1p(fromValueUSD / 10000) / Math.LN10`

apps/api/src/lib/trade-utils.ts
```diff
@@ -16,13 +16,31 @@ export interface SlippageResult {
  * @returns SlippageResult containing slippage calculations
  */
 export function calculateSlippage(fromValueUSD: number): SlippageResult {
-  // Apply slippage based on trade size
-  const baseSlippage = (fromValueUSD / 10000) * 0.05; // 0.05% per $10,000 (10x lower than before)
-  const actualSlippage = baseSlippage * (0.9 + Math.random() * 0.2); // ±10% randomness (reduced from ±20%)
+  // Apply slippage based on trade size with logarithmic scaling
+  // This prevents slippage from exceeding reasonable limits for large trades
```
carsonfarmer, line 38:
```suggestion
  if (!Number.isFinite(fromValueUSD) || fromValueUSD < 0) {
    throw new Error("fromValueUSD must be a finite, non-negative number");
  }

  // This prevents slippage from exceeding reasonable limits for large trades
```

apps/api/src/lib/trade-utils.ts
```diff
@@ -16,13 +16,31 @@ export interface SlippageResult {
...
+    0.02 * Math.log10(1 + fromValueUSD / 10000),
+    MAX_SLIPPAGE,
+  );
+
+  // Apply randomness: ±10% variation
+  const randomMultiplier = 0.9 + Math.random() * 0.2;
+  const actualSlippage = Math.min(
+    baseSlippage * randomMultiplier,
+    MAX_SLIPPAGE,
+  );
```
carsonfarmer, line 60: Check me on this, but I think this is a bit clearer and you'll avoid the possibility of a negative.
```suggestion
  const actualSlippage = Math.max(
    0, Math.min(baseSlippage * randomMultiplier, MAX_SLIPPAGE)
  );
```

apps/api/src/lib/trade-utils.ts
```diff
@@ -16,13 +16,31 @@ export interface SlippageResult {
...
   const slippagePercentage = actualSlippage * 100;
 
   // Calculate final amount with slippage
   const effectiveFromValueUSD = fromValueUSD * (1 - actualSlippage);
+  if (effectiveFromValueUSD < 0) {
+    throw new Error("effectiveFromValueUSD is negative");
+  }
```
carsonfarmer: I think if you do the clamping above as I suggested, you no longer need this?

```suggestion
```

apps/api/src/lib/__tests__/trade-utils.test.ts
```diff
@@ -0,0 +1,117 @@
+import { describe, expect, it } from "vitest";
+
+import { calculateSlippage } from "../trade-utils.js";
+
+describe("trade-utils", () => {
```
carsonfarmer, line 5: This all looks reasonable and good. Also, yes tests!

Do we need to add tests for negative and non-integer input values? Or, maybe extremely large values (e.g., Number.MAX_SAFE_INTEGER)?

PR #1177, opened 2025-09-09
feat: add rewards endpoints

brunocalza:
# Context

Closes [APP-318](https://linear.app/recall-labs/issue/APP-318/rewards-api-implementation)

Adds the following 3 endpoints:
1.  `GET /user/rewards/total`
2.  `GET /user/rewards/proofs`
3.  ~`POST /user/rewards/claim/`~

These will be used to implement the CLAIM flow button in the navbar:

image

### DIFF
--- a/apps/api/e2e/tests/rewards-api.test.ts
+++ b/apps/api/e2e/tests/rewards-api.test.ts
@@ -1,0 +1,195 @@
+import { hexToBytes } from "viem";
+import { beforeEach, describe, expect, test, vi } from "vitest";
+
+import { BoostRepository } from "@recallnet/db/repositories/boost";
+import { CompetitionRepository } from "@recallnet/db/repositories/competition";
+import { RewardsRepository } from "@recallnet/db/repositories/rewards";
+import { competitions } from "@recallnet/db/schema/core/defs";
+import { InsertReward } from "@recallnet/db/schema/voting/types";
+import { RewardsService, createLeafNode } from "@recallnet/services";
+
+import { db } from "@/database/db.js";
+import { ApiClient } from "@/e2e/utils/api-client.js";
+import {
+  RewardProof,
+  RewardsProofsResponse,
+  RewardsTotalResponse,
+} from "@/e2e/utils/api-types.js";
+import {
+  createPrivyAuthenticatedClient,
+  createTestClient,
+} from "@/e2e/utils/test-helpers.js";
+import { logger } from "@/lib/logger.js";
+
+import { TestPrivyUser } from "../utils/privy.js";
+
+// Mock the RewardsAllocator class
+const mockRewardsAllocator = {
+  allocate: vi.fn().mockResolvedValue({
+    transactionHash:
+      "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef",
+    blockNumber: BigInt(12345),
+    gasUsed: BigInt(100000),
+  }),
+};
+
+let testUserAddress: string;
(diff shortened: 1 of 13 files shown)


apps/api/e2e/tests/rewards-api.test.ts
```diff
@@ -0,0 +1,290 @@
...
+  let rewardsService: RewardsService;
+  let testCompetitionId: string;
+  let testRewards: InsertReward[];
+
+  // Test constants for the allocate method
+  const testTokenAddress =
+    "0x1234567890123456789012345678901234567890" as `0x${string}`;
+  const testStartTimestamp = Math.floor(Date.now() / 1000); // Current timestamp
+
+  beforeEach(async () => {
```
brunocalza, line 51: Before each test, we add some entries in the `rewards` table. We also call the mocked `allocate` call that will add entries to `rewards_tree` and `rewards_roots` tables. With that data in the database, we can test the 3 endpoints.

ukstv (approved):

apps/api/e2e/tests/rewards-api.test.ts
```diff
@@ -0,0 +1,290 @@
...
+    // Verify testClient and testUser are properly initialized
+    expect(testClient).toBeDefined();
+    expect(testUser).toBeDefined();
+    expect(testUser.walletAddress).toBeDefined();
+
+    // Get total claimable rewards using the test client
+    const response = await testClient.getTotalClaimableRewards();
+
+    expect(response.success).toBe(true);
+    if (response.success) {
```
joewagner: We don't need to make this check, the assertion above is enough

apps/api/e2e/tests/rewards-api.test.ts
```diff
@@ -0,0 +1,290 @@
...
+      expect(response.totalClaimableRewards).toBe("2");
+    }
+  });
+
+  test("authenticated user can get rewards with proofs", async () => {
+    // Get rewards with proofs using the test client
+    const response = await testClient.getRewardsWithProofs();
+
+    expect(response.success).toBe(true);
+    if (response.success) {
```
joewagner: Don't need this check either.  We should start looking for these as part of review since it seems like llms like to do this as a way to make the type checker happy.  A better way to do that is use the chaining operator `?` or just cast the thing since the assertion will ensure its the right type.

apps/api/e2e/tests/rewards-api.test.ts
```diff
@@ -0,0 +1,290 @@
...
+      });
+    }
+  });
+
+  test("authenticated user can claim all rewards", async () => {
+    // Claim all rewards using the test client
+    const response = await testClient.claimAllRewards();
+
+    expect(response.success).toBe(true);
+    if (response.success) {
```
joewagner: same here

apps/api/src/controllers/rewards.controller.ts
```diff
@@ -0,0 +1,131 @@
...
+     * @param res Express response
+     * @param next Express next function
+     */
+    async getTotalClaimableRewards(
+      req: Request,
+      res: Response,
+      next: NextFunction,
+    ) {
+      try {
+        const address = req.wallet as string;
```
joewagner: We should use Zod for this instead of casting.  That will catch problems that arise from the string "undefined" being passed into functions.

There's two other places in this file that have the same issue, line 68 and 103.

joewagner (commented):
Nice! 
I put a few nits inline, but looks good

PR #1187, opened 2025-09-09
Full test coverage for rewards sub-package

carsonfarmer:


carsonfarmer:
I tagged a bunch of folks, but I only really need one person to give me a 👍 . This touches some work and tests that both @brunocalza and @ukstv have recently done, so wanted to make sure it looks good to them.

.github/workflows/code-quality.yml
```diff
@@ -17,9 +17,10 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
-        # with:
-        #   submodules: true
-        #   fetch-depth: 0
+        with:
```
brunocalza, line 20: Using PAT is not recommended, the way to have access to private submodules is with SSH and Deploy keys. Check it out in [this](https://github.com/recallnet/js-recall/blob/main/.github/workflows/api-ci.yml#L120-L130) pipeline. Although this would not matter if we make the staking contracts repo public

.gitmodules
```diff
@@ -1,7 +1,3 @@
-[submodule "packages/contracts/contracts"]
-	path = packages/contracts/contracts
-	url = git@github.com:recallnet/contracts.git
-
 [submodule "packages/staking-contracts/contracts"]
 	path = packages/staking-contracts/contracts
-	url = git@github.com:recallnet/recall-staking-contracts.git
```
brunocalza, line 7: This looks wrong. Changing the submodule from remote to local?

brunocalza (changes requested):
Looks good overall! I am almost sure the change on `.gitmodules` should be reverted though

.gitmodules
carsonfarmer: Apparently this works fine on remote and local setups. But actually, I'm probably going to just make the repo public and avoid all this shenanigans in the first place.

carsonfarmer: https://git-scm.com/docs/git-submodule

> This may be either an absolute URL, or (if it begins with ./ or ../), the location relative to the superproject’s default remote repository (Please note that to specify a repository foo.git which is located right next to a superproject bar.git, you’ll have to use ../foo.git instead of ./foo.git - as one might expect when following the rules for relative URLs - because the evaluation of relative URLs in Git is identical to that of relative directories).

.github/workflows/code-quality.yml
carsonfarmer: Interesting, do you know why?

brunocalza: Hmm, I could swear this was an official recommendation from Github docs, but can't find it. But the way I see is that PAT are tied to personal accounts, and ideally you should isolate things by giving the repo its own access rights through an SSH key that is independent of any person

.gitmodules
brunocalza: I see. It is relative to the superproject’s remote

.github/workflows/code-quality.yml
carsonfarmer: Good point. I used a PAT for textileio-machine. But even that is sort of "not right". The reason I _like_ PATs is that they can have expiration dates. But, actually, we'll just avoid this whole thing by making the repo public anyway, in which case, I'll remove this stuff before I merge anyway.

PR #1190, opened 2025-09-10
feat(api): add subscribe/unsubscribe endpoints | remove from comps

dtbuchholz:
closes [APP-7](https://linear.app/recall-labs/issue/APP-7/add-loops-to-comps-app). note: we decided to remove the "Notify me" in the footer. we already built the hooks and API. they're useful to merge because in the future, we can add buttons in the user profile page to subscribe/unsubscribe from the mailing list.

### DIFF
--- a/apps/api/.env.test
+++ b/apps/api/.env.test
@@ -81,3 +81,4 @@ DISABLE_CACHE_API=true
 LOOPS_API_KEY=faux_loops_api_key
 LOOPS_MAILING_LIST_ID=faux_loops_mailing_list_id
 LOOPS_BASE_URL=http://127.0.0.1:4010
+


apps/api/src/controllers/user.controller.ts
```diff
@@ -479,6 +479,97 @@ export function makeUserController(services: ServiceRegistry) {
...
       }
     },
+
+    /**
+     * Subscribe to Loops mailing list
+     * @param req Express request with userId from session
+     * @param res Express response
+     * @param next Express next function
+     */
+    async subscribe(req: Request, res: Response, next: NextFunction) {
```
dtbuchholz, line 489: user auth gates the `/user/subscribe` and `/user/unsubscribe` endpoints. if the boolean `isSubscribed` conflicts with the intent, we throw an error; else, we update the user.

apps/api/src/lib/loops/mock.ts
```diff
@@ -0,0 +1,8 @@
+import { LoopsPayload, LoopsResponse } from "@/services/email.service.js";
+
+export async function updateContactMock(
```
dtbuchholz: added a mock Loops API for testing purposes

apps/api/src/controllers/user.controller.ts
```diff
@@ -479,6 +479,97 @@ export function makeUserController(services: ServiceRegistry) {
...
+        const { email, isSubscribed } = user;
+        if (!email) {
+          // Note: this should never happen post-Privy migration since Privy guarantees an email
+          throw new ApiError(404, "User email not found");
+        }
+        if (isSubscribed) {
+          throw new ApiError(400, "User already subscribed");
+        }
+        // Subscribe to Loops mailing list
+        const result = await services.emailService.subscribeUser(email);
```
stbrody, line 509: is this idempotent?  What happens if we call this and the user is already subscribed?  Trying to think through what happens if the server crashes in between this call and the next call to update the database and mark the user as subscribed

apps/api/src/controllers/user.controller.ts
```diff
@@ -479,6 +479,97 @@ export function makeUserController(services: ServiceRegistry) {
...
+        const { email, isSubscribed } = user;
+        if (!email) {
+          // Note: this should never happen post-Privy migration since Privy guarantees an email
+          throw new ApiError(404, "User email not found");
+        }
+        if (!isSubscribed) {
+          throw new ApiError(400, "User already unsubscribed");
+        }
+        // Unsubscribe from Loops mailing list
+        const result = await services.emailService.unsubscribeUser(email);
```
stbrody, line 557: same concern re: idempotency

apps/api/src/services/email.service.ts
```diff
@@ -156,6 +156,15 @@ export class EmailService {
   private async updateContact(
     payload: LoopsPayload,
   ): Promise<{ success: boolean; error?: string }> {
+    // In test mode, use a local mock to avoid external HTTP calls
+    if (config.server.nodeEnv === "test") {
+      const { updateContactMock } = await import("@/lib/loops/mock.js");
```
stbrody: let's avoid inline imports like this.  We can just import this regularly at the top of the file

apps/api/src/services/email.service.ts
```diff
@@ -156,6 +156,15 @@ export class EmailService {
   private async updateContact(
     payload: LoopsPayload,
   ): Promise<{ success: boolean; error?: string }> {
+    // In test mode, use a local mock to avoid external HTTP calls
```
stbrody: I don't love changing real code to accommodate tests...
If this isn't here, how many tests fail?  How much of a pain would it be to update the test setup for affected tests to mock out the loops service in the test harness setup rather than here in the production code?

apps/api/openapi/openapi.json
```diff
@@ -1795,12 +1795,47 @@
       "get": {
         "tags": ["Admin"],
         "summary": "List all agents",
-        "description": "Get a list of all agents in the system",
+        "description": "Get a paginated list of all agents in the system",
```
stbrody, line 1798: I don't see any changes to the backend that correspond with the changes to these API definitions?  Am I missing where we're adding the pagination logic?

Also nitpick but this change seems fairly unrelated to the subscribe/unsubscribe changes so would ideally be in its own PR

dtbuchholz: when you run the `pnpm generate` command, it updates the openapi spec. if someone didnt run it after making API changes, they get included here, too

stbrody: oh interesting, I thought this file was the source of truth for the API and code gets generated off of it, this is actually also autogenerated code?  Where's the source of truth for this then?

I do wish our repo was structured to put all the autogenerated code into a single `autogenerated` directory so it was really obvious what is generated code or not...

dtbuchholz: the openapi spec is generated from the inline comments on all `routes/` files. the comments on those route handlers are also "up to the author" and don't necessarily reflect the actual responses. LLMs make it easy to auto-complete those comments, but there's plenty of room for human error / missing fields / wrong status codes.

our ideal flow is definitely what you alluded to: the spec lives in a separate package => we spec out new API changes => `apps/api` routes conforms to it. see (legacy issue) [REC2-205](https://linear.app/recall-labs/issue/REC2-205/spec-first-openapi-development-pattern)

apps/api/src/controllers/user.controller.ts
dtbuchholz: in terms of Loops—it'll update the contact regardless and wont fail if, e.g., the user is already subscribed and we try to subscribe them again. but good point. i'll just change the logic so that we dont throw a 400 on "already subscribed" and let it pass through.

same w/ below.

apps/api/src/services/email.service.ts
dtbuchholz: in theory, vitest should let us mock package imports and override behavior _only_ in tests. i had trouble with this in the past and couldnt figure it out—e.g., in places like our API middleware. i can give it a try.

dtbuchholz: note: because our e2e tests spawn different servers, we have to set up a mock Loops server for all tests. i've added that in the most recent changes and removed the inline / mock imports.

apps/api/e2e/setup.ts
```diff
@@ -145,9 +149,24 @@ export async function setup() {
     process.env.METRICS_PORT = "3003";
   }
 
+  // Ensure Loos variables are set (to trigger the `EmailService` logic properly)
```
stbrody, line 152:
```suggestion
  // Ensure Loops variables are set (to trigger the `EmailService` logic properly)
```

PR #1193, opened 2025-09-11
feat: configure automatic deployment to prod

sam701:
- Add release-prod workflow to update `compapp-prod*` and `compapp-sandbox` containers.
- Rename `release` to `release-staging`.

Test run with the current image (`textile/js-recall:sha-4040f96`): https://github.com/recallnet/js-recall/actions/runs/17642049560

### DIFF
--- a/.github/workflows/release-prod.yml
+++ b/.github/workflows/release-prod.yml
@@ -5,10 +5,8 @@ permissions:
 
 on:
   push:
-    branches:
-      - alexei/app-401-configure-automatic-deployment-to-prod
-    # tags:
-    #   - '*'
+    tags:
+      - '*'
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release-prod.yml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  trigger-api-deployment:
+    if: ${{ github.repository_owner == 'recallnet' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
dtbuchholz, line 21: Do we need to fetch the submodules too? It's not used in the backend API app yet, but I _think_ it will be soon

https://github.com/recallnet/js-recall/blob/main/.github/workflows/coverage.yml#L21

sam701: We don’t actually need to fetch submodules - we are only checking out the repo to get the docker image name from the git commit hash.

PR #1196, opened 2025-09-11
feat: Award Boost

ukstv:
1. When Stake happens the indexer awards Boost for the active competition.
2. Every 3 hours a script goes through every upcoming competition and staked stake position to award Boost.


apps/api/scripts/award_boost.ts
```diff
@@ -0,0 +1,108 @@
...
+    const upcomingCompetitionsRes =
+      await competitionManager.getUpcomingCompetitions();
+    const upcomingCompetitions = upcomingCompetitionsRes.competitions;
+    logger.info(
+      `${upcomingCompetitions.length} competitions need Boost awarded to the users`,
+    );
+    // Iterate over competitions
+    for (const competition of upcomingCompetitions) {
+      const votingEndDate = competition.votingEndDate;
+      if (!votingEndDate) {
```
stbrody, line 37: This is interesting because it introduces a new constraint on competitions.  I believe in the past we have sometimes opened voting for a competition without having a voting end date set yet, and just had the admin (ie Kristal) close voting manually at some later point.  Now with this change, its actually important that all competitions get created with votingStartDate and votingEndDate decided upfront.

FYI @KGmajor to confirm that that is an okay constraint to introduce.

If so, we should probably make those fields required in the createCompetition flow

apps/api/src/database/repositories/boost.repository.ts
```diff
@@ -146,17 +150,18 @@ class BoostRepository {
    * - `amount` must be `>= 0n`. Zero-amount credits will only create a journal row
    *   on first application (useful for marking events without changing balance).
    */
-  increase(args: BoostDiffArgs): Promise<BoostDiffResult> {
+  increase(args: BoostDiffArgs, tx?: DbTransaction): Promise<BoostDiffResult> {
     const amount = args.amount;
     if (amount < 0n) {
       throw new Error("amount must be non-negative");
     }
     const idemKey = args.idemKey ?? randomBytes(32);
```
stbrody, line 158: should we let idemKey be missing?  Seems pretty important to always have an idemKey to avoid constantly granting more and more boost, maybe that would be better as a required argument?

apps/api/scripts/award_boost.ts
stbrody: We also might need to think about what happens if a competition is created but then gets rescheduled, including its voting start and end dates...

apps/api/src/services/boost-award.service.ts
```diff
@@ -0,0 +1,81 @@
...
+class BoostAwardService {
+  readonly #boostRepository: BoostRepository;
+  readonly #divisor: bigint;
+
+  constructor(boostRepository: BoostRepository) {
+    this.#boostRepository = boostRepository;
+    this.#divisor = 10n ** BigInt(DECIMALS);
+  }
+
+  // FIXME Play with scenarios
```
stbrody, line 36: leftover dev comment?

apps/api/scripts/award_boost.ts
KGmajor: Good call out. We do start every competition with a voting start and end date, so that's no problem. I don't anticipate that we will change that, since we want everything to be routine and on the schedule. 

Competition types might get rescheduled, but we will always host a competition that week. So the only thing that will change is likely the name, description, and type. For example, if Live Trading was broken, we would simply fall back to a paper trading competition and run that instead - updating the competition information and not the dates.

apps/api/src/database/repositories/boost.repository.ts
ukstv: This thing is _internal_ by definition. It is up to a caller to define what an idemKey is semantically and if ever it should be used.

PR #1200, opened 2025-09-12
Full stack boost API

asutula:
Everything from db schema, to repository code to controllers to api client to react hooks and tests are done.

Closes APP-287 and APP-325

### DIFF
--- a/apps/api/src/database/repositories/boost.repository.ts
+++ b/apps/api/src/database/repositories/boost.repository.ts
@@ -2,7 +2,7 @@ import { and, eq, sql, sum } from "drizzle-orm";
 import { randomBytes } from "node:crypto";
 import { z } from "zod";
 
-import { Transaction } from "@recallnet/db-schema/types";
+import type { Transaction } from "@recallnet/db-schema/types";
 import * as schema from "@recallnet/db-schema/voting/defs";
 import {
   SelectAgentBoost,
--- a/apps/api/scripts/award_boost.ts
+++ b/apps/api/scripts/award_boost.ts
@@ -1,0 +1,117 @@
+import * as dotenv from "dotenv";
+import cron from "node-cron";
+import * as path from "path";
+
+import { BlockchainAddressAsU8A } from "@/lib/coders.js";
+import { createLogger } from "@/lib/logger.js";
+import { ServiceRegistry } from "@/services/index.js";
+
+// Load environment variables
+dotenv.config({ path: path.resolve(process.cwd(), ".env") });
+
+const services = new ServiceRegistry();
+const logger = createLogger("AwardBoost");
+
+/**
+ * Assign boost to users for upcoming competitions
+ */
+async function awardAllBoost() {
+  const startTime = Date.now();
+  logger.info("Starting award Boost task...");
+
+  const competitionManager = services.competitionManager;
+  const stakesRepository = services.stakesRepository;
+  const userManager = services.userManager;
+  const boostAwardService = services.boostAwardService;
+
+  try {
+    const upcomingCompetitionsRes =
+      await competitionManager.getUpcomingCompetitions();
+    const upcomingCompetitions = upcomingCompetitionsRes.competitions;
+    logger.info(
+      `${upcomingCompetitions.length} competitions need Boost awarded to the users`,
+    );
+    // Iterate over competitions
+    for (const competition of upcomingCompetitions) {
+      const votingEndDate = competition.votingEndDate;
+      if (!votingEndDate) {
+        logger.warn(`No voting end date for competition: ${competition.id}`);
+        continue;
+      }
+      const votingStartDate = competition.votingStartDate;
+      if (!votingStartDate) {
+        logger.warn(`No voting start date for competition: ${competition.id}`);
+        continue;
+      }
+      // Iterate over Stakes pages
+      let stakes = await stakesRepository.allStaked();
+      while (stakes.length > 0) {
+        // Iterate over Stakes
+        for (const stake of stakes) {
+          const walletString = BlockchainAddressAsU8A.decode(stake.wallet);
+          const user = await userManager.getUserByWalletAddress(walletString);
+          if (!user) {
+            logger.warn(`User not found for wallet: ${stake.wallet}`);
+            continue; // For some reason no user found for this wallet
+          }
+          // Now is the time to award boost
+          const diffResult = await boostAwardService.awardForStake(
+            {
+              id: stake.id,
+              wallet: BlockchainAddressAsU8A.decode(stake.wallet),
+              amount: stake.amount,
+              stakedAt: stake.stakedAt,
+              canUnstakeAfter: stake.canUnstakeAfter,
+            },
+            {
+              id: competition.id,
+              votingEndDate: votingEndDate,
+              votingStartDate: votingStartDate,
+            },
+          );
+          if (diffResult.type === "applied") {
+            logger.info(
+              `Awarded Boost to user ${user.id} for stake ${stake.id}, balance is ${diffResult.balanceAfter}`,
+            );
+          }
+        }
+
+        // Fetch the next page of Stakes
+        const lastStakeId = stakes[-1]?.id;
+        if (lastStakeId) {
+          stakes = await services.stakesRepository.allStaked(lastStakeId);
+        } else {
+          stakes = [];
+        }
+      }
+    }
+
+    const duration = Date.now() - startTime;
+    logger.info(`Awarding Boost completed successfully in ${duration}ms!`);
+  } catch (error) {
+    logger.error(
+      "Error awarding Boost:",
+      error instanceof Error ? error.message : String(error),
+    );
+
+    throw error;
+  }
+}
+
+const isRunOnce = process.argv.includes("--run-once");
+if (isRunOnce) {
+  logger.info("Running award Boost task once");
+  try {
+    await awardAllBoost();
+  } catch {
+    process.exit(1);
+  } finally {
+    process.exit(0);
+  }
+} else {
+  // Schedule the task to run every 3 hours
+  cron.schedule("0 */3 * * *", async () => {
+    logger.info("Running award Boost task");
+    await awardAllBoost();
+  });
+}
--- a/apps/api/scripts/award_boost_faux.ts
+++ b/apps/api/scripts/award_boost_faux.ts
@@ -1,0 +1,91 @@
+import * as dotenv from "dotenv";
+import * as path from "path";
+import { parse } from "ts-command-line-args";
+
+import { createLogger } from "@/lib/logger.js";
+import { ServiceRegistry } from "@/services/index.js";
+
+// Load environment variables
+dotenv.config({ path: path.resolve(process.cwd(), ".env") });
+
+const services = new ServiceRegistry();
+const logger = createLogger("AwardBoost");
+
+type Args = {
+  competitionId: string;
+  boostAmount: number;
+  reason?: string;
+};
+
+/**
+ * Assign boost to users for upcoming competitions
+ */
+async function awardBoostFaux(args: Args) {
+  const startTime = Date.now();
+  const competitionManager = services.competitionManager;
+  const userManager = services.userManager;
+  const boostAwardService = services.boostAwardService;
+  logger.info("Starting award Boost task...");
+  const competitionId = args.competitionId;
+  const boostAmount = BigInt(args.boostAmount); // Beware! No decimals.
+  const idemReason = args.reason ?? "script";
+
+  try {
+    const competition = await competitionManager.getCompetition(competitionId);
+    if (!competition) {
+      throw new Error(`Competition not found: ${args.competitionId}`);
+    }
+
+    const users = await userManager.getAllUsers();
+    for (const user of users) {
+      // Now is the time to award boost
+      const diffResult = await boostAwardService.awardNoStake(
+        competitionId,
+        user.id,
+        user.walletAddress,
+        boostAmount,
+        idemReason,
+      );
+      if (diffResult.type === "applied") {
+        logger.info(
+          `Awarded ${boostAmount} Boost to user ${user.id} for competition ${competitionId}, balance is ${diffResult.balanceAfter}`,
+        );
+      }
+    }
+
+    const duration = Date.now() - startTime;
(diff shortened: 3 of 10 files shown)


apps/api/src/controllers/competition.controller.ts
```diff
@@ -858,6 +858,7 @@ export function makeCompetitionController(services: ServiceRegistry) {
           competitionId,
           registeredParticipants: competition.registeredParticipants,
           maxParticipants: competition.maxParticipants,
+          // TODO: Each agent contains a bigint totalBoost. How to serialize?
```
ramirotw: I'd say the easiest way is to just stringify it

asutula: Done.

asutula: And it gets converted back to bigint in the api client.

apps/api/src/controllers/boost.controller.ts
```diff
@@ -0,0 +1,149 @@
...
+          Buffer.from(val, "base64");
+          return true;
+        } catch {
+          return false;
+        }
+      },
+      { message: "Invalid base64 string" },
+    )
+    .transform((val) => Buffer.from(val, "base64")),
+});
```
asutula, line 29: The `idemKey` gets sent as base64 encoded string, this schema parses it for us and converts it to byte array for storage in the db.

apps/api/src/controllers/boost.controller.ts
```diff
@@ -0,0 +1,149 @@
...
+          userId,
+          competitionId,
+        });
+
+        res.status(200).json({
+          success: true,
+          boosts: Object.fromEntries(
+            Object.entries(boosts).map(([key, value]) => [
+              key,
+              value.toString(),
```
asutula, line 101: Need to convert any `bigint`s to `string`.

apps/api/src/controllers/boost.controller.ts
```diff
@@ -0,0 +1,149 @@
...
+        // TODO: Make the window size configurable.
+        // 3 days before competition start.
+        const windowStart =
+          competition.startDate.getTime() - 3 * 24 * 60 * 60 * 1000;
+        if (!(windowStart < now && now < competition.endDate.getTime())) {
+          throw new ApiError(
+            400,
+            "Can't boost in a competition outside of the boost time window.",
+          );
+        }
```
asutula: Doing this check, but can make the window size configurable when the team decides about it.

apps/api/src/controllers/competition.controller.ts
```diff
@@ -858,7 +858,10 @@ export function makeCompetitionController(services: ServiceRegistry) {
           competitionId,
           registeredParticipants: competition.registeredParticipants,
           maxParticipants: competition.maxParticipants,
-          agents,
+          agents: agents.map((agent) => ({
+            ...agent,
+            boostTotal: agent.boostTotal.toString(),
+          })),
```
asutula: Include the agent's boost total in the competition with all the rest of the agent properties.

apps/api/src/database/repositories/agent-repository.ts
```diff
@@ -238,6 +240,7 @@ async function findByCompetitionImpl(
         competitions,
         eq(competitionAgents.competitionId, competitions.id),
       )
+      .innerJoin(agentBoostTotals, eq(agentBoostTotals.agentId, agents.id))
```
asutula: Join to agent boost totals table so we can return that value with the agent info.

apps/api/src/database/repositories/boost.repository.ts
```diff
@@ -53,8 +58,42 @@ type BoostDiffArgs = {
...
 type BoostDiffResult =
-  | { changeId: string; balanceAfter: bigint; idemKey: Uint8Array } // change applied
-  | { balance: bigint; idemKey: Uint8Array }; // noop (already applied)
+  | {
+      type: "applied";
+      changeId: string;
+      balanceAfter: bigint;
+      idemKey: Uint8Array;
+    } // change applied
+  | { type: "noop"; balance: bigint; idemKey: Uint8Array }; // noop (already applied)
```
asutula, line 69: Added the `type` string literal property to make this a discriminated union, which is much easier to deal with.

apps/api/src/database/repositories/boost.repository.ts
```diff
@@ -146,7 +185,7 @@ class BoostRepository {
    * - `amount` must be `>= 0n`. Zero-amount credits will only create a journal row
    *   on first application (useful for marking events without changing balance).
    */
-  increase(args: BoostDiffArgs): Promise<BoostDiffResult> {
+  increase(args: BoostDiffArgs, tx?: Transaction): Promise<BoostDiffResult> {
```
asutula, line 190: Added the ability to pass in a tx.

apps/api/src/database/repositories/boost.repository.ts
```diff
@@ -378,10 +424,249 @@ class BoostRepository {
...
+          and(
+            eq(schema.agentBoostTotals.agentId, agentId),
+            eq(schema.agentBoostTotals.competitionId, competitionId),
+          ),
+        )
+        .limit(1)
+        .for("update"); // locks this row until tx ends
+
+      // Decrease the user's boost balance
+      const diffRes = await this.decrease(
```
asutula, line 635: Use our existing `decrease` function.

apps/comps/lib/api-client.ts
```diff
@@ -557,6 +577,87 @@ export class ApiClient {
...
+    amount,
+  }: {
+    competitionId: string;
+    agentId: string;
+    currentAgentBoostTotal: bigint;
+    amount: bigint;
+  }): Promise<{ success: boolean; agentTotal: bigint }> {
+    const idemKey = btoa(
+      `${competitionId}-${agentId}-${currentAgentBoostTotal.toString()}`,
+    );
```
asutula, line 666: Here is how we calculate a good idempotency key. We force the caller to pass in the target agent's current boost balance as they see it. Combine that with the competition id and agent id, and that should be good.

packages/db-schema/src/indexing/defs.ts
asutula, line 1: `tokenAmount` column type for stakes and stake changes tables.

packages/db-schema/src/voting/defs.ts
```diff
@@ -190,7 +189,7 @@ export const boostChanges = pgTable(
         onDelete: "restrict",
         onUpdate: "cascade",
       }),
-    deltaAmount: bigint("delta_amount", { mode: "bigint" }).notNull(), // earn:+X, spend:-X
+    deltaAmount: tokenAmount("delta_amount").notNull(), // earn:+X, spend:-X
```
asutula, line 182: Update these two places for boost balance/change to `tokenAmount`.

packages/db-schema/src/voting/defs.ts
```diff
@@ -211,3 +210,55 @@ export const boostChanges = pgTable(
     createdIdx: index("boost_changes_created_at_idx").on(t.createdAt),
   }),
 );
+
+export const agentBoostTotals = pgTable(
+  "agent_boost_totals",
```
asutula, line 205: This table hold an always updated list of agent boost totals per competition id.

packages/db-schema/src/voting/defs.ts
```diff
@@ -211,3 +210,55 @@ export const boostChanges = pgTable(
...
+    index("agent_boost_totals_competition_id_idx").on(t.competitionId),
+    uniqueIndex("agent_boost_totals_agent_id_competition_id_idx").on(
+      t.agentId,
+      t.competitionId,
+    ),
+  ],
+);
+
+export const agentBoosts = pgTable(
+  "agent_boosts",
```
asutula, line 231: This is a log of all boost actions by agent and competition. It references the `boostChanges` table.

packages/db-schema/src/voting/defs.ts
```diff
@@ -132,7 +131,7 @@ export const boostBalances = pgTable(
     competitionId: uuid("competition_id")
```
brunocalza, line 121: Question, not related to the PR, about `wallet: bytea("wallet").notNull(),`: isn't it better to reference `user_id`, so that it's the `user` that has boost balance, and not the wallet? I fear that in the future we may support user changing wallets and then we would have to update this

apps/api/src/database/repositories/boost.repository.ts
```diff
@@ -378,10 +424,249 @@ class BoostRepository {
...
+   */
+  async userBoosts(
+    { userId, competitionId }: { userId: string; competitionId: string },
+    tx?: Transaction,
+  ) {
+    const executor = tx || this.#db;
+    const res = await executor
+      .select({
+        agentId: schema.agentBoosts.agentId,
+        boostTotal: sum(schema.boostChanges.deltaAmount).mapWith(BigInt),
```
brunocalza: This `sum` will be negative, right? Let me see if I understood things correctly. Every time users spend boost on an agent, we add a negative entry to `boost_changes`, so when we try to get the total a user has spent on agents by sum, we would get a negative number. Is that so? If so, the UI must flip the sign?

packages/db-schema/src/voting/defs.ts
brunocalza: Is this an optimization so that we don't have to aggregate data from agent_boosts every time?

apps/api/src/controllers/boost.controller.ts
```diff
@@ -0,0 +1,152 @@
...
+
+import { db } from "@/database/db.js";
+import { BoostRepository } from "@/database/repositories/boost.repository.js";
+import { ApiError } from "@/middleware/errorHandler.js";
+import ServiceRegistry from "@/services/index.js";
+
+import { ensureUuid } from "./request-helpers.js";
+
+const BoostAgentSchema = z.object({
+  amount: z.coerce.bigint(),
```
brunocalza, line 12: So this is sent as a string and converted to bigint?

packages/db-schema/src/voting/defs.ts
asutula: I was wondering about this. @ukstv might have some thoughts and @jovirecall as well.

apps/api/src/database/repositories/boost.repository.ts
asutula: You are absolutely right (lol I'm an llm)!. I'll multiply by -1.

packages/db-schema/src/voting/defs.ts
asutula: Yes.

apps/api/src/controllers/boost.controller.ts
asutula: Right, it comes as a string in a JSON payload of the POST request. I use this schema to validate the payload and also coerce to the bigint type we want.

apps/api/src/database/repositories/agent-repository.ts
asutula: This should be `leftJoin`. Will push that update.

asutula:
Made a simple UI integration to test this all out and found a couple bugs. Fixed them. Seems to work nicely.

asutula:
Used Amp to write extremely thorough integration tests for the boost repository. Integrated those tests into CI. Made some important updates to the db schema, which I'll comment on in the code diff.

apps/api/src/database/repositories/__tests__/boost.repository.agentBoostTotals.integration.test.ts
asutula, line 1: Each of the files in this directory contains integration tests for a single boost repository function. Amp went all the way in and tested every possibly things, which I guess I good since these are very important code paths involving real money.

packages/db-schema/src/util.ts
```diff
@@ -13,3 +13,14 @@ export function tokenAmount(name: string) {
...
+export const bytea = customType<{
+  data: Uint8Array | Buffer; // what your app uses
+  driverData: Buffer; // what node-postgres returns
+  notNull: false;
+  default: false;
+}>({
+  dataType: () => "bytea",
+  toDriver: (v) => (v instanceof Buffer ? v : Buffer.from(v)),
+  fromDriver: (v) => v, // Buffer
+});
```
asutula, line 26: Moved this custom type in here to a shared location.

packages/db-schema/src/voting/defs.ts
```diff
@@ -14,22 +12,11 @@ import {
...
-  dataType: () => "bytea",
-  toDriver: (v) => (v instanceof Buffer ? v : Buffer.from(v)),
-  fromDriver: (v) => v, // Buffer
-});
+import { agents, competitions, users } from "../core/defs.js";
+import { blockchainAddress, bytea, tokenAmount } from "../util.js";
 
 export const epochs = pgTable("epochs", {
-  id: uuid().primaryKey().notNull(),
+  id: uuid().primaryKey().notNull().defaultRandom(),
```
asutula, line 19: Saw you can set a default uuid in postgres. Makes it nice not having to provide it in code.

packages/db-schema/src/voting/defs.ts
asutula: Important update here. I did switch this to referencing user. I moved the `wallet` column to the `boostChanges` table below, so every change is lined to a wallet address and associated with the user. If, for some reason, we allow the user to change their wallet address or link multiple wallets, this will continue to work and their boost across all wallet addresses linked to them will be reflected in this table. cc @ukstv

packages/db-schema/src/voting/defs.ts
```diff
@@ -183,31 +171,84 @@ export const boostBalances = pgTable(
...
-    id: uuid("id").primaryKey().notNull(),
+    id: uuid("id").primaryKey().notNull().defaultRandom(),
     balanceId: uuid("balance_id")
       .notNull()
       .references(() => boostBalances.id, {
         onDelete: "restrict",
         onUpdate: "cascade",
       }),
-    deltaAmount: bigint("delta_amount", { mode: "bigint" }).notNull(), // earn:+X, spend:-X
+    wallet: bytea("wallet").notNull(),
```
asutula, line 181: Always track the wallet with each change event. These are connected to the user via the `balanceId` column. One day, we might make a table to hold wallet addresses per user, then reference that table from this one to save some bytes, but for now this is fine.

ukstv (approved):

PR #1206, opened 2025-09-12
feat: Make endCompetition transactional and idempotent

stbrody:
The goal of this PR is twofold: to make it so that `endCompetition` is resilient if the server crashes midway through processing it, and to make it resilient to `endCompetition` being run in parallel on the same competition by two different app servers / serverless functions at once.

The most straightforward way to do this would have been to do all the writes of `endCompetition` in a single transaction, but I was worried about how big and long running that transaction would be.  I was especially worried about locking out access to the `competition` table for a long time while that transaction runs, since that table is queried in many places throughout the app.  So I took a kind of hybrid approach, making the second half of `endCompetition` fully happen in a single transaction, but the first half of `endCompetition` happens outside of a transaction, but is updated to be resilient to being re-run if the operation fails/crashes midway through.  Hopefully this represents a good balance between keeping the code not too complex, making things safe and resilient, and keeping the performance impact of ending a competition on the rest of the app down.

The final transaction for the second half of endCompetition is still pretty big though, we'll definitely want to keep an eye on the performance impacts of it.

### DIFF
--- a/apps/api/scripts/backfill-agent-score.ts
+++ b/apps/api/scripts/backfill-agent-score.ts
@@ -8,6 +8,7 @@ import {
   competitionAgents,
   competitions,
 } from "@recallnet/db-schema/core/defs";
+import type { SelectCompetitionsLeaderboard } from "@recallnet/db-schema/core/types";
 import {
   agentScore,
   agentScoreHistory,
@@ -118,7 +119,7 @@ async function backfillCompetitionScores(competitionId: string) {
   console.log(chalk.gray(`Processing ${activeAgents.length} active agents`));
 
   // Get the leaderboard for this competition (might be incomplete)
-  const leaderboard =
+  const leaderboard: SelectCompetitionsLeaderboard[] | null =
     await competitionRepo.findLeaderboardByCompetition(competitionId);
 
   // Create a map for easy lookup of leaderboard positions
--- a/apps/api/src/database/repositories/agentscore-repository.ts
+++ b/apps/api/src/database/repositories/agentscore-repository.ts
@@ -11,13 +11,12 @@ import {
   InsertAgentScoreHistory,
   SelectAgentScore,
 } from "@recallnet/db-schema/ranking/types";
+import type { Transaction as DatabaseTransaction } from "@recallnet/db-schema/types";
 
 import { db } from "@/database/db.js";
 import { repositoryLogger } from "@/lib/logger.js";
 import { createTimedRepositoryFunction } from "@/lib/repository-timing.js";
 import { AgentMetadata } from "@/types/index.js";
-
-type DatabaseTransaction = Parameters<Parameters<typeof db.transaction>[0]>[0];
 
 /**
  * Agent Rank Repository
@@ -97,16 +96,19 @@ async function getAllAgentRanksImpl(
 async function batchUpdateAgentRanksImpl(
   dataArray: Array<Omit<InsertAgentScore, "id" | "createdAt" | "updatedAt">>,
   competitionId: string,
+  tx?: DatabaseTransaction,
 ): Promise<InsertAgentScore[]> {
   if (dataArray.length === 0) {
     repositoryLogger.debug("No agent ranks to update in batch");
     return [];
   }
 
+  const executor = tx || db;
+
   try {
     repositoryLogger.debug(`Batch updating ${dataArray.length} agent ranks`);
 
-    return await db.transaction(async (tx) => {
+    return await executor.transaction(async (tx) => {
       // Prepare rank data with IDs
       const rankDataArray: InsertAgentScore[] = dataArray.map((data) => ({
         id: uuidv4(),
--- a/apps/api/src/database/repositories/competition-repository.ts
+++ b/apps/api/src/database/repositories/competition-repository.ts
@@ -142,9 +142,6 @@ interface Snapshot24hResult {
 const snapshotCache = new Map<string, [number, Snapshot24hResult]>();
 const MAX_CACHE_AGE = 1000 * 60 * 5; // 5 minutes
 
-// Type for database transaction
-type DatabaseTransaction = Parameters<Parameters<typeof db.transaction>[0]>[0];
-
 /**
  * Find all competitions
  */
--- a/apps/api/src/database/repositories/competition-repository.ts
+++ b/apps/api/src/database/repositories/competition-repository.ts
@@ -9,6 +9,7 @@ import {
   inArray,
   isNotNull,
   lte,
+  or,
   sql,
 } from "drizzle-orm";
 import { unionAll } from "drizzle-orm/pg-core";
@@ -27,6 +28,7 @@ import {
   InsertCompetitionAgent,
   InsertCompetitionsLeaderboard,
   SelectCompetition,
+  SelectCompetitionsLeaderboard,
   UpdateCompetition,
 } from "@recallnet/db-schema/core/types";
 import {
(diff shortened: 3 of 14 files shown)


apps/api/scripts/backfill-agent-score.ts
```diff
@@ -118,7 +119,7 @@ async function backfillCompetitionScores(competitionId: string) {
   console.log(chalk.gray(`Processing ${activeAgents.length} active agents`));
 
   // Get the leaderboard for this competition (might be incomplete)
-  const leaderboard =
+  const leaderboard: SelectCompetitionsLeaderboard[] | null =
```
stbrody, line 122: I had some lint failures on missing types pop up even on lines I didn't change... not quite sure why this wound up coming up during this change, but seems like reasonable changes so 🤷.  So this PR has a few random extra type annotations added throughout that isn't actually related to the core changes here, hopefully its not too distracting to reviewers

apps/api/src/services/providers/__tests__/batch-functionality.test.ts
```diff
@@ -1,11 +1,14 @@
...
+import { describe, expect, it, vi } from "vitest";
 
 import { PriceTracker } from "@/services/price-tracker.service.js";
 import { BlockchainType } from "@/types/index.js";
 
 import { DexScreenerProvider } from "../dexscreener.provider.js";
 import { MultiChainProvider } from "../multi-chain.provider.js";
 
+// Set timeout for all tests in this file to 15 seconds
+vi.setConfig({ testTimeout: 15_000 });
```
stbrody, line 10: this is unrelated to the changes here, but the tests in this file were failing sporadically in CI so hopefully this will make them more stable.

apps/api/src/services/providers/__tests__/multi-chain.provider.test.ts
```diff
@@ -28,7 +31,6 @@ describe("MultiChainProvider", () => {
 
   beforeEach(() => {
     provider = new MultiChainProvider();
-    vi.setConfig({ testTimeout: 30_000 }); // Increase timeout for API calls
```
stbrody, line 31: there was a CI failure where one of the tests in this file timed out after 5 seconds, even though this is supposedly setting the timeout to 30 seconds.  Claude thinks its because the `beforeEach` runs too late, after `vi` has already configured itself, so it suggested moving the timeout up to the top of the file, which is what I did here and in the dexscreener test

andrewxhill:
## AI Review Summary

✅ **Approved** - This PR successfully implements transactional and idempotent competition ending with proper concurrency handling.

### Key Features Reviewed
- **New "ending" Status**: Added intermediate state between "active" and "ended" to prevent race conditions
- **Atomic Operations**: Uses database transactions to ensure consistency during competition finalization
- **Idempotent Design**: Can safely retry failed ending operations without corruption
- **Comprehensive Testing**: Includes dedicated test suite for auto-end functionality

### Code Quality Assessment
- **Database Layer**: Well-structured repository functions with proper transaction handling
- **Service Layer**: Clean separation of concerns with atomic state transitions
- **Error Handling**: Robust error handling for concurrent execution scenarios
- **Type Safety**: Proper TypeScript types after schema regeneration

### Security & Reliability
- **Race Condition Protection**: Uses SELECT FOR UPDATE to prevent concurrent competition endings
- **Transaction Safety**: Critical operations wrapped in database transactions
- **Rollback Support**: Failed operations don't leave database in inconsistent state

### Test Coverage
- Existing E2E tests for auto-end functionality in `auto-end-competition.test.ts`
- Tests verify both successful ending and proper handling of not-yet-ready competitions
- Tests simulate cron-like execution scenarios

### Build Status
- ✅ Linting passes
- ✅ TypeScript compilation succeeds (after schema rebuild)
- ✅ No critical issues detected

This implementation follows financial system best practices with proper atomicity and idempotence - exactly what's needed for reliable competition management.

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

apps/api/src/database/repositories/agentscore-repository.ts
```diff
@@ -97,7 +97,15 @@ async function getAllAgentRanksImpl(
...
   dataArray: Array<Omit<InsertAgentScore, "id" | "createdAt" | "updatedAt">>,
   competitionId: string,
+  tx?: DatabaseTransaction,
 ): Promise<InsertAgentScore[]> {
+  // If no transaction provided, create one and call recursively
+  if (!tx) {
+    return await db.transaction(async (newTx) => {
+      return await batchUpdateAgentRanksImpl(dataArray, competitionId, newTx);
+    });
+  }
```
asutula: I've been doing this with a simpler syntax:
```
const executor = tx || db;
return executor.transaction(async (tx) => { ... })

File: apps/api/src/database/repositories/competition-repository.ts [hunk 5]
```diff
@@ -61,6 +64,16 @@ import { PartialExcept } from "./types.js";
  * Handles database operations for competitions
  */
 
+// Type for database transaction
+type DatabaseTransaction = Parameters<Parameters<typeof db.transaction>[0]>[0];
```
asutula: This type is already exported from the `db-schema` package if you want to use it:
https://github.com/recallnet/js-recall/blob/main/packages/db-schema/src/types.ts#L8

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -282,6 +295,82 @@ async function updateOneImpl(
...
+        .select()
+        .from(competitions)
+        .where(
+          and(
+            eq(competitions.id, competitionId),
+            eq(competitions.status, "active"),
+          ),
+        )
+        .for("update")
+        .limit(1);
```
asutula: This `select ... for update` seems unnecessary because you are simply updating the competition row with new values. You're not using any value from the `select` to calculate a new value to set in the update below. If you simply run the update below, postgres already locks the row at the moment it is being updated, and that is all you need in this case.

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -282,6 +295,82 @@ async function updateOneImpl(
...
+/**
+ * Mark competition as ended (ending -> ended) within a transaction
+ * This acts as a guard to ensure only one process can finalize a competition
+ * @param competitionId Competition ID
+ * @param tx Database transaction
+ * @returns The competition if successfully marked as ended, null otherwise
+ */
+async function markCompetitionAsEndedImpl(
+  competitionId: string,
+  tx: DatabaseTransaction,
```
asutula: Nice to make this optional and use the pattern I showed above.

apps/api/src/database/repositories/agentscore-repository.ts
asutula: No recursive call, less code.

stbrody: Nice, didn't realize you could call `.transaction` on a transaction.  I assume its basically a no-op then?

stbrody:
okay, I addressed most of the comments, but I left the two-phase competition ending because taking portfolio snapshots involves making network calls to the external dexscreener price API and I really don't think that should be happening inside of the database transaction, that will make it way too long running and put a lot of pressure on the database.

Please take another look @asutula

apps/api/src/database/repositories/agentscore-repository.ts
asutula: It starts a new txn checkpoint, which I what you'd want I think. It's like "ok i've been doing work in this txn, and now i'm going to hand it to you to do your work", so starting an new checkpoint for your work is that I'd want.

asutula (approved):
Nice!

PR #1209, opened 2025-09-12
Revert "fix(comps): safari encoding (#1208)"

dtbuchholz:
This reverts commit 0ad0be7a86304dff0e02f9199c51ba201c5dd52a.

### DIFF
--- a/apps/comps/app/api/proxy/[...path]/route.ts
+++ b/apps/comps/app/api/proxy/[...path]/route.ts
@@ -113,7 +113,13 @@ async function proxy(req: Request, path: string[]) {
 
   // 6) Strongly discourage CDN caching of proxied responses
   respHeaders.set("cache-control", "no-store");
-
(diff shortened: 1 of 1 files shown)


apps/comps/app/api/proxy/[...path]/route.ts
```diff
@@ -114,12 +114,6 @@ async function proxy(req: Request, path: string[]) {
   // 6) Strongly discourage CDN caching of proxied responses
   respHeaders.set("cache-control", "no-store");
```
Comment (carsonfarmer) [inline apps/comps/app/api/proxy/[...path]/route.ts]:
```suggestion
  const incoming = new Headers(req.headers);
  if (hasBody && !incoming.get("accept-encoding")) {
    respHeaders.delete("content-encoding");
    respHeaders.delete("content-length");
    respHeaders.delete("transfer-encoding");
    respHeaders.delete("connection");  
  }
```

PR #1231, opened 2025-09-16
fix(comps): missing env results in bad image url

joewagner:
Small fix here.  
I was doing a review and I had not set the `NEXT_PUBLIC_FRONTEND_URL` env, and this causes an endless loop of requests for `/undefined//logo_white_full.svg`.  I'm not sure how this would be handled in prod, but locally the app was DOSing itself.  
This change throws an error and stops next from re-requesting the image repeatedly.


apps/comps/providers/privy-provider.tsx
```diff
@@ -23,7 +27,7 @@ const privyConfig: PrivyClientConfig = {
   appearance: {
     theme: themeBackground, // Use dark theme as base since your CSS uses dark colors
     accentColor: themeForeground,
-    logo: `${process.env.NEXT_PUBLIC_FRONTEND_URL}/logo_white_full.svg`,
```
joewagner, line 26: Notice that  
```
`${process.env.FOO}` === "undefined"  // -> true
```

hence this url is `/undefined/logo_white_full.svg`

joewagner:
Actually this is complex than I thought, I need to fix this up.

apps/comps/providers/privy-provider.tsx
```diff
@@ -23,7 +25,7 @@ const privyConfig: PrivyClientConfig = {
...
+    logo: `${NEXT_PUBLIC_FRONTEND_URL}/logo_white_full.svg`,
     showWalletLoginFirst: false, // Show social login options first
     walletList: [
```
dtbuchholz, line 28: if we want to avoid this annoyance altogether, i uploaded the logo to vercel blobs so we can just set it to this value: `https://5pskttgrmgbdllus.public.blob.vercel-storage.com/logo_white.png`

here's what it looks like (it's the same):

Image

apps/comps/wagmi-config.ts
```diff
@@ -2,13 +2,14 @@ import { getDefaultConfig } from "connectkit";
...
+
 export const clientConfig: () => Config = () => {
   const configParams = getDefaultConfig({
     appName: "Recall Competitions",
     appDescription: "Compete with your AI agents in Recall competitions",
-    appUrl:
-      process.env.NEXT_PUBLIC_FRONTEND_URL || "https://app.recall.network",
-    appIcon: `${process.env.NEXT_PUBLIC_FRONTEND_URL}/favicon.ico`,
+    appUrl: NEXT_PUBLIC_FRONTEND_URL || "https://app.recall.network",
+    appIcon: `${NEXT_PUBLIC_FRONTEND_URL}/favicon.ico`,
```
dtbuchholz, line 12: fwiw, i think we _technically_ don't need the wagmi config, anymore. it was for SIWE/walletconnect/connectkit. but, maybe we can use the same blobs URL here too

apps/comps/providers/privy-provider.tsx
joewagner: Oh nice, I'll update that value for the privy logo.  This turned into a bit bigger PR, since the env us used a bunch of places.
I _think_ all of our preview build would be broken without this (if they weren't broken because of the db connect issues)

dtbuchholz (approved):

apps/comps/wagmi-config.ts
joewagner: Is the favicon already in the blobs store somewhere?

apps/comps/providers/privy-provider.tsx
dtbuchholz: > I think all of our preview build would be broken by this (if they weren't broken because of the db connect issues)

ah yeah that's true...we'd have to dynamically infer the base URL or something like that. not worth it.

apps/comps/wagmi-config.ts
dtbuchholz: ah right. nope. just uploaded it here: https://5pskttgrmgbdllus.public.blob.vercel-storage.com/favicon.ico

PR #1235, opened 2025-09-17
chore: Move business logic from AgentController into service layer

stbrody:


apps/api/src/controllers/agent.controller.ts
```diff
@@ -286,17 +254,11 @@ export function makeAgentController(services: ServiceRegistry) {
         const trades =
           await services.tradeSimulatorService.getAgentTrades(agentId);
 
-        // Sort trades by timestamp (newest first)
-        const sortedTrades = [...trades].sort(
```
stbrody, line 290: sorting is already happening in the underlying SQL statement, so this in-memory sort is redundant

PR #1237, opened 2025-09-17
chore: Move business logic from BoostController to service layer

stbrody:


apps/api/src/database/repositories/boost.repository.ts
```diff
@@ -536,7 +542,7 @@ class BoostRepository {
   async userBoosts(
     { userId, competitionId }: { userId: string; competitionId: string },
     tx?: Transaction,
-  ) {
+  ): Promise<Record<string, bigint>> {
```
asutula, line 545: This is fine, but just wonder why you don't let the return type be inferred?

apps/api/src/services/boost.service.ts
```diff
@@ -0,0 +1,172 @@
...
+   */
+  async getUserBoostBalance(
+    userId: string,
+    competitionId: string,
+  ): Promise<bigint> {
+    try {
+      // Get the user to validate they exist
+      const user = await this.userService.getUser(userId);
+      if (!user) {
+        throw new ApiError(404, "User not found");
```
asutula, line 54: Not saying we should change this now, but it's weird to have http errors in a class that has nothing to do with http. This class could be used in any context. We can revisit in the future.

asutula (approved):

apps/api/src/database/repositories/boost.repository.ts
stbrody: I just think explicit types are generally better than implicit ones

apps/api/src/services/boost.service.ts
stbrody: yeah...  https://github.com/recallnet/js-recall/pull/1243 has this type of problem even more tbh.
It's tricky to leave it totally http agnostic without introducing a whole bunch of redundant work into the controller.  The error handling there in particular was pretty nasty, doing a bunch of string matching to determine what code to use...  At some point we'll probably want a more unified error system throughout the codebase that uses internal error codes, and then maps them to http codes at the edges

PR #1241, opened 2025-09-18
feat(api): wallet watchlist with chainalysis

dtbuchholz:
Implements sanctioned wallet checks whenever a user tries to link their wallet to their account. Closes [APP-399](https://linear.app/recall-labs/issue/APP-399/block-all-wallets-on-the-sdn-list)


apps/api/src/services/watchlist.service.ts
```diff
@@ -0,0 +1,164 @@
...
+
+        // If API is down or returns error, fail-safe (allow access)
+        if (!response.ok) {
+          serviceLogger.warn(
+            {
+              address: normalizedAddress,
+              status: response.status,
+              statusText: response.statusText,
+            },
+            "[WatchlistService] Skipping watchlist address check. Chainalysis API returned error",
```
joewagner: I have the same general thought here as on the claims pr.  If chainalysis isn't working, is it ok to give a pass to a sanctioned address?

joewagner (commented):
This is great.  My only question is in the inline comment.

apps/api/src/services/watchlist.service.ts
dtbuchholz: hm yeah fair point. i've changed this so that we'll just throw. i also added some retry logic...sorta similar to the claims stuff but just as a way to try and make sure Chainalysis doesnt cause user creation problems

apps/api/src/lib/retry-helper.ts
carsonfarmer, line 1: Wonder if we should just use this one (copy it) and then reference in a doc string: https://github.com/vercel/async-retry

carsonfarmer: There is also this one: https://github.com/franckLdx/ts-retry

Since this is pretty simple, it is probably ok to replicate. But do we have this type of functionality elsewhere? I think we do?

apps/api/src/routes/health.routes.ts
```diff
@@ -104,6 +104,7 @@ export function configureHealthRoutes(
    *                     agentService:
    *                       type: string
    *                       description: Status of the agent manager service
+   *                       example: ok
```
carsonfarmer, line 107: nice catch

apps/api/src/services/watchlist.service.ts
```diff
@@ -0,0 +1,203 @@
...
+        try {
+          const response = await fetch(`${this.baseUrl}/${normalizedAddress}`, {
+            method: "GET",
+            headers: {
+              "X-API-Key": this.apiKey,
+              Accept: "application/json",
+            },
+            signal: controller.signal,
+          });
+          clearTimeout(timeoutId);
```
carsonfarmer: Could stick this in a finally clause

apps/api/src/lib/watchlist.ts
```diff
@@ -0,0 +1,203 @@
...
+                address: normalizedAddress,
+                identifications: data.identifications.filter(checkIsSanctioned),
+              },
+              "[WatchlistService] SANCTIONED ADDRESS DETECTED",
+            );
+          }
+
+          return isSanctioned;
+        } catch (error) {
+          clearTimeout(timeoutId);
```
carsonfarmer, line 150: same here

apps/api/src/services/watchlist.service.ts
```diff
@@ -0,0 +1,203 @@
...
+          }
+
+          return isSanctioned;
+        } catch (error) {
+          clearTimeout(timeoutId);
+
+          // Handle network errors and timeouts as retryable
+          if (error instanceof Error) {
+            const message = error.message.toLowerCase();
+            if (
```
carsonfarmer: This seems kinda brittle, no?

apps/api/src/services/user.service.ts
```diff
@@ -90,6 +94,21 @@ export class UserService {
       const normalizedEmbeddedWalletAddress =
         embeddedWalletAddress?.toLowerCase();
 
+      // Check wallet addresses against sanctions list
+      // Note: user registration always uses the wallet address and embedded wallet as the same
+      // value, and then users can update it later. This is more of a safeguard.
+      const isWalletSanctioned =
+        normalizedEmbeddedWalletAddress &&
+        normalizedWalletAddress !== normalizedEmbeddedWalletAddress &&
+        (await this.watchlistService.isAddressSanctioned(
```
carsonfarmer: While probably fine, maybe capture this variable above and then include it here? I am always weary of inlined awaits like this...

apps/api/src/services/watchlist.service.ts
carsonfarmer: We should fail if they try to create a wallet that is sanctioned. But if we wanted to, we could, in theory, just not let them link a sanctioned wallet, but still let them create the user account without a linked wallet?

apps/api/src/lib/retry-helper.ts
dtbuchholz: > Since our primary entry point for wallets is privy, would it make more sense to integrate at that layer?

i originally had the same intuition. and considering the future Vercel-first user pattern, maybe we won't need this service in `apps/api`? i was a bit hesitant, though, because we had a bunch of little issues with Privy and dealing various hooks. but hm yeah, maybe it makes sense to just move this logic to `apps/comps`?

> The watchlist service is essentially stateless, do we need to use the same "services" pattern here? Maybe that's just a style thing?

fair. yeah i just kept it with how the codebase sets everything up. i can move it into `lib` instead.

> The withRetry function has been written 1000 times before, should we a) pull in a library, or b) have our own "helpers" shared library?

sure, works for me

apps/api/src/services/user.service.ts
dtbuchholz: i'll just remove it from `registerUser`, actually. in practice, our user registration _never_ uses a custom linked wallet, and always uses the embedded wallet as the `walletAddress`. this check will never get used.

apps/api/src/services/watchlist.service.ts
dtbuchholz: agreed (addressed by _not_ using this method in `registerUser`, which wont take a custom wallet address anyways)

apps/api/src/lib/retry-helper.ts
dtbuchholz: > Since this is pretty simple, it is probably ok to replicate. But do we have this type of functionality elsewhere? I think we do?

note: instead of using external deps, i had GPT-5 rewrite ours to take patterns from those libs into account. ours is still pretty small. if desired, i can still swap it out for some existing lib.

oh and generally...hm no, i dont see anything in `apps/api` that with the classic HTTP retry logic. there is some retry logic in db-related methods (resetting the db, or taking a porfolio snapshot). 

the only other place that uses retry logic in HTTP calls is in our price provider (e.g., DexScreener API calls), but the logic is pretty simple and only sets a a min/max retry attempts with a fixed delay. given that i'm implementing CoinGecko API calls, i'll use this new helper in that code.

also—created a follow up to consider a `@recallnet/helpers` package: [APP-500](https://linear.app/recall-labs/issue/APP-500/create-helpers-package-for-commonly-used-methods-shared-across)

carsonfarmer: Re:

> i originally had the same intuition. and considering the future Vercel-first user pattern, maybe we won't need this service in apps/api? i was a bit hesitant, though, because we had a bunch of little issues with Privy and dealing various hooks. but hm yeah, maybe it makes sense to just move this logic to apps/comps?

What do you think? cc @asutula. I think we're fine to keep this "as is" for now. We might find we need to update/change as we move more APIs over.

But it would be good to make sure @asutula is aware of this.

PR #1244, opened 2025-09-18
fix(ci): stop random chars from interactive output

dtbuchholz:
## Summary

fixes annoying random characters from being injected into your terminal when pushing to remote.

## Details

with the husky pre-push hook, it will run / exit...but then it logs a bunch of random numbers/characters. these are raw ANSI control codes from interactive output (spinners/progress) leaking out of the Git hook—e.g., pnpm.

image

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,4 +1,5 @@
 #!/bin/sh
+export CI=1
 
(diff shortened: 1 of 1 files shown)


smrz2001 (approved):
🚀

PR #1246, opened 2025-09-18
Shared repositories

asutula:
This touches lots of files, but is mostly a big refactor of where the repository files live. I'll leave code comments to call out important changes worth reading and understanding.


apps/api/scripts/award_boost.ts
```diff
@@ -2,7 +2,8 @@ import * as dotenv from "dotenv";
 import cron from "node-cron";
 import * as path from "path";
 
-import { BlockchainAddressAsU8A } from "@/lib/coders.js";
+import { BlockchainAddressAsU8A } from "@recallnet/db-schema/coders";
```
asutula: Moved the coders to the shared db packages so they can be used in that package and here in the api app.

apps/api/src/controllers/competition.controller.ts
```diff
@@ -638,7 +639,11 @@ export function makeCompetitionController(services: ServiceRegistry) {
 
         res.status(200).json(responseBody);
       } catch (error) {
-        next(error);
+        next(
+          error instanceof ParsingError
+            ? new ApiError(400, error.message)
+            : error,
+        );
```
asutula, line 267: I didn't like a service throwing an http error, so I made an error type in the db package and that error type is thrown. We catch it here in the controller layer and map it to an http error. This is a very unsophisticated version of what we can do moving forward.

apps/api/src/database/repositories/admin-repository.ts
```diff
@@ -1,303 +1,10 @@
...
-async function countImpl(): Promise<number> {
-  try {
-    const [result] = await db.select({ count: drizzleCount() }).from(admins);
-    return result?.count ?? 0;
-  } catch (error) {
-    repositoryLogger.error("Error in count:", error);
-    throw error;
-  }
-}
+const repository = new AdminRepository(db, repositoryLogger);
```
asutula, line 7: The thing to notice about all the "repository" files that still live in the api app is that they are simply timing wrappers around the actual repositories that have been moved to the shared db package. Here we create the repository instance, passing in its dependencies.

apps/api/src/database/repositories/admin-repository.ts
```diff
@@ -309,73 +16,73 @@ async function countImpl(): Promise<number> {
  */
 
 export const create = createTimedRepositoryFunction(
-  createImpl,
+  repository.create.bind(repository),
   "AdminRepository",
   "create",
 );
```
asutula, line 22: For consistency and simplicity and not having to update everywhere the repository functions are used, I left the timing wrapper definitions in place, just swapping out the old function being wrapped for the new one that lives on the repository instance.

It will be up to us to decide how/if to implement timing tracking like this in the comps app. When we do, it will probably not be like this and will use more integrated OpenTelemetry tooling.

packages/db-schema/src/repositories/types/index.ts
asutula, line 1: This `types/index.ts` as well as the other folders/files located in `types/` contain typescript types used by various repositories. In their original location in the api app, these types are quite confusing and used and tangled up in many other parts of the app. For simplicity, just copied the types used by the repository files, and omitted everything else. I left the file/directory structure the same as it is in the original location. I made this choice as to not make major interface changes to the repository functions. We can and should clean up how/when/where we create/export types for the repositories.

packages/db-schema/src/repositories/admin.ts
```diff
@@ -0,0 +1,300 @@
...
+import { InsertAdmin, SelectAdmin } from "../core/types.js";
+import { Database } from "../types.js";
+import { SearchAdminsParams } from "./types/index.js";
+import { PartialExcept } from "./util/types.js";
+
+/**
+ * Admin Repository
+ * Handles database operations for admins
+ */
+export class AdminRepository {
```
asutula: Ok this is a repository. All the repositories in this folder follow this same pattern.

A repository is now modeled as a class. I went back and forth on whether to use classes or plain objects (I don't like classes personally), and went with class because I've seen other people on the team prefer them. Encapsulating repository functions in a class gives us a nice way to pass in any dependency the repository has. Usually this is a database instance and a logger.

packages/db-schema/src/utils/drop-all.ts
asutula, line 1: Used in integration tests.

packages/db-schema/src/utils/push-schema.ts
asutula, line 1: Also used in integration tests.

packages/db-schema/src/errors.ts
```diff
@@ -0,0 +1,7 @@
+export class ParsingError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = "ParsingError";
+    Error.captureStackTrace(this, this.constructor);
+  }
+}
```
asutula: Custom error thrown by some parsing code in `src/repositories/util/query/ts`.

packages/db-schema/src/repositories/util/query.ts
```diff
@@ -17,14 +17,12 @@ export function getSort<T extends PgSelect>(
...
+  if (!sortString || sortString.trim() === "") {
+    throw new ParsingError("cannot sort by undefined");
   }
+  const parts = sortString.split(",");
   if (parts.length > 3) {
-    throw new ApiError(
-      400,
+    throw new ParsingError(
       "compound sorting with more than 3 fields not allowed",
     );
```
asutula: Using that custom error type in here.

apps/api/init_comps_v2.sh
```diff
@@ -0,0 +1,485 @@
+#!/bin/bash
```
stbrody: why is this script being added as part of this PR?  Seems unrelated?

apps/api/package.json
```diff
@@ -16,8 +16,6 @@
     "build": "tsc && tsc-alias && copyfiles -u 0 './drizzle/**/*.{sql,json}' dist",
     "test": "vitest run --project unit",
     "test:watch": "vitest watch --project unit",
-    "test:integration": "vitest run --project integration",
```
stbrody, line 19: why are we removing these?

apps/api/src/controllers/competition.controller.ts
stbrody: I still don't like that this would potentially require every controller to have big catch blocks matching on lots of different errors.  Check out how the comps controller joinCompetition method worked before my recent refactor, for an example of what I'd like to avoid: https://github.com/recallnet/js-recall/blob/1602a0fd838c44f796bf0c39fd9b27fadfc35ae9/apps/api/src/controllers/competition.controller.ts#L949-L1003

I wonder if instead we can make this cleaner by having our own error type that has an abstract `toApiError()` method, and then all the concrete error types that we define (like `ParsingError`) have to extend that base error type and implement `toApiError()` (or could just be `getHttpCode()` that is the abstract method they have to implement and `toApiError()` is concrete in the base class and calls `getHttpCode()`).  That way the catch block in the controllers can just be a single call to convert all possible error types to the appropriate `ApiError` without having to match on every possible kind of error.


Anyway fine to do this for now and revisit later

apps/api/src/database/repositories/admin-repository.ts
stbrody: probably at some point we should have a RepositoryRegistry similar to the ServiceRegistry and construct all the repository instance in one place at startup rather than staticly within the files like this...
But this is good enough for now

packages/db-schema/src/repositories/__tests__/boost.agentBoostTotals.integration.test.ts
```diff
@@ -0,0 +1,1114 @@
+import { and, eq } from "drizzle-orm";
```
stbrody: are these tests migrated from somewhere or are they brand new in this PR?

packages/db-schema/src/repositories/types/index.ts
```diff
@@ -0,0 +1,172 @@
...
+  "svm",
+]);
+
+// Type derived from the Zod schema
+export type SpecificChain = z.infer<typeof SpecificChainSchema>;
+
+/**
+ * Zod schema for the status of a competition.
+ */
+export const CompetitionStatusSchema = z.enum(competitionStatus.enumValues);
```
stbrody: Some of the types in this file, such as this one, I don't see a corresponding change where this type is being moved from somewhere else.  In fact, looking at the competition controller, it seems to be still be importing a same-named type from a different location.  I think we probably don't want to introduce duplicate types here and instead want to migrate the existing calls to use these new versions?

packages/db-schema/src/repositories/types/sort/agent.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+export const AgentSortField = makeSortFieldSchema([
+  ...AGENT_DB_FIELDS,
+  ...AGENT_COMPUTED_FIELDS,
+]).default("rank");
+
+/**
+ * Agent query schema (database fields and computed fields)
+ */
+export const AgentQuerySchema = PagingSchema.extend({
```
stbrody: ditto my last comment, the only user of a type with this name, CompetitionController, is still using the old version of this type

packages/db-schema/src/utils/drop-all.ts
```diff
@@ -0,0 +1,23 @@
+import { sql } from "drizzle-orm";
+import { isPgSchema } from "drizzle-orm/pg-core";
+
+import schema from "../index.js";
+import { Database } from "../types.js";
+
+export async function dropAll(db: Database) {
```
stbrody: consider renaming this to `dropAllSchemas`?  Otherwise its not really clear anymore what "all" is being dropped

stbrody (approved):
this PR makes it hard to see if anything changed between the source and dest repositories, so I didn't review those and just assume its strict code motion.
I also didn't review the seemingly new integration test files.

Otherwise the general structural changes seem fine for now, even if they're kind of incomplete and still leave redundant controller files in the old locations.  Would be nice to clean those up and move everything to reference the new files soon instead.

The most actionable comments are about making sure we don't introduce redundant types and actually migrate callers over to the new types in the new locations.  Feel free to do that as a follow-up pr though if you want to get this one in quickly

apps/api/init_comps_v2.sh
asutula: It was an accident, will remove it.

apps/api/package.json
asutula: Because we don't have any integration tests in the `api` now. I moved the only integration tests over to the `db` package. Leaving this here was causing an error "no test files".

apps/api/src/database/repositories/admin-repository.ts
asutula: Yea I agree, but my goal here was to move the repositories to a shared package while minimizing how existing code uses them.

packages/db-schema/src/repositories/__tests__/boost.agentBoostTotals.integration.test.ts
asutula: Migrated from the `api` app.

packages/db-schema/src/repositories/types/index.ts
asutula: Correct, I just copied the types over as a short term solution for the sake of simplicity. These types are still defined and used in the api app. The way types are defined and used in `api` is pretty confusing, so I didn't want to put a lot more work into preserving that or re-inventing it. I figured we can come up with a good approach for types when we implement the new api and shared services.

PR #1224, opened 2025-09-16
fix(comps): retrieving agent api keys

dtbuchholz:
closes [APP-419](https://linear.app/recall-labs/issue/APP-419/users-cannot-get-api-key-in-comps-ui-due-to-sandboxprod-db-mismatch). a few fixes:
1. legacy users in the sandbox db might _only_ have a wallet address
   - before: when we look up the agent API key, we were using the admin search API for user + agent combinations (`user.walletAddress` +  `agent.handle`). this can lead to problems in case the prod<>sandbox dbs get out of sync with user profile information.
   - now: since agent handles are globally unique, we can be more lenient and simply fetch the sandbox agent API key based on their handles. our existing logic better handles updates to the sandbox if the agent handle changes.
2. agent didn't exist
   - before: somehow, the frontend didnt trigger an agent to get created. when a user tried to get the sandbox API key, there was nothing to fetch.
   - now: we will create a sandbox agent if the agent handle is not found.
3. cant get prod keys, even though logged in
   - before: our old logic had wonky email verification that gated both prod and sandbox keys.
   - now: production API keys are always shown to the user. sandbox API keys are hidden, but only because the user needs to manually "join competition". (note: we can probably refactor this and auto-join agents into the sandbox competition.)

PR #1250, opened 2025-09-19
fix(api): api sandbox key fetching

dtbuchholz:
closes [APP-448](https://linear.app/recall-labs/issue/APP-448/fix-sandbox-api-keys)

### DIFF
--- a/apps/comps/app/api/sandbox/_lib/privy-utils.ts
+++ b/apps/comps/app/api/sandbox/_lib/privy-utils.ts
@@ -1,0 +1,121 @@
+import {
+  LinkedAccountWithMetadata,
+  PrivyClient,
+  User as PrivyUser,
+  WalletWithMetadata,
+} from "@privy-io/server-auth";
+import { NextRequest } from "next/server";
+
+/**
+ * A Privy user with their connected accounts.
+ */
+interface PrivyUserWithAccounts {
+  privyId: string;
+  email: string;
+  walletAddress: string;
+  embeddedWalletAddress: string;
+}
+
+/**
+ * Check if a Privy user is set up with a custom linked wallet. Custom linked wallets are linked
+ * accounts with a wallet client type that is not "privy" (i.e., not an embedded wallet).
+ * @param wallet - The linked account to check.
+ * @returns True if the linked account is a custom linked wallet, false otherwise. If the user is not
+ * set up with a custom linked wallet, returns false.
+ */
+export function isCustomLinkedWallet(
+  wallet: LinkedAccountWithMetadata,
+): wallet is WalletWithMetadata {
+  return wallet.type === "wallet" && wallet.walletClientType !== "privy";
+}
+
+/**
+ * Check if a Privy user is set up with an embedded wallet. Embedded wallets are linked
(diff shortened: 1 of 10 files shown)


apps/comps/app/api/sandbox/_lib/privy-utils.ts
```diff
@@ -0,0 +1,121 @@
+import {
```
dtbuchholz, line 1: the [original design](https://github.com/recallnet/js-recall/pull/1224) for the sandbox setup had some steps where we'd call the "main" API (prod) to get the uesr's profile, and then use that information in "sandbox" API calls.

i had removed that code because i thought our `/search` API didnt need the `user` param (below), but we _do_ want to include the user's wallet address or email in our queries.

so, adding this Privy logic reduces an API call to the main API backend since we can just parse the Privy ID token for the user's wallet and email.

apps/comps/app/api/sandbox/agents/route.ts
```diff
@@ -22,25 +15,24 @@ import {
...
-
-  // Fetch user profile from the base API to get walletAddress
-  const profileData = await mainApiRequest<ProfileResponse>(
-    "/user/profile",
-    sessionCookie,
-  );
-  const { walletAddress } = profileData.user;
-
+  // Extract privy ID from cookie and reject if fully unauthenticated
+  const privyUser = await getPrivyUserFromCookie(request);
```
dtbuchholz, line 19: e.g., here, we use the privy helpers to get the wallet or email.

apps/comps/app/api/sandbox/api-key/route.ts
```diff
@@ -31,25 +31,28 @@ async function handleGetAgentApiKey(request: NextRequest) {
   }
 
   // Find the agent's ID in the sandbox using its handle
-  const searchData = await sandboxAdminRequest<AdminSearchResult>(
-    `/admin/search?agent.handle=${encodeURIComponent(agentHandle)}`,
+  // Note: We check if user already exists in sandbox with wallet address, else, email. This helps
+  // with Privy-related backwards compatibility. The user inclusion in the search query is
+  // *important* to ensure that the requester is the owner of the agent, otherwise, the API key can
+  // be leaked.
+  let searchForUserWithAgent = await sandboxAdminRequest<AdminSearchResult>(
```
dtbuchholz, line 38: part of the reason why API calls to the sandbox fail is because we can't find the `user.email`. this is because legacy sandbox users were never created with an email address, and they only have a wallet address. 

now, we try to get the agent with a `user.walletAddress` or `user.email`. if one of those succeeds, then we can return the agent's API key.

note: i didn't want to alter any backend logic, but we could improve this logic, if needed.

apps/comps/lib/api-client.ts
```diff
@@ -193,14 +192,6 @@ export class ApiClient {
 
   // Authentication endpoints
 
-  /**
-   * Get nonce for message signature
-   * @returns Nonce response
-   */
-  async getNonce(): Promise<NonceResponse> {
```
dtbuchholz, line 200: unrelated cleanups

PR #1253, opened 2025-09-19
feat: integrates boosts and rewards

brunocalza:


packages/services/src/rewards.service.ts
```diff
@@ -45,10 +62,76 @@ export class RewardsService {
...
+
+      const rewards = this.calculate(
+        prizePoolUsers,
+        prizePoolCompetitors,
+        boostAllocations,
+        leaderBoard,
+        boostAllocationWindow,
+      );
+
+      // TODO: add user_id, and agent_id columns to rewards table
```
brunocalza, line 113: i'll see if I can add this in another PR

packages/services/src/rewards.service.ts
```diff
@@ -45,10 +62,76 @@ export class RewardsService {
...
     try {
-      const rewards = await this.calculate();
-      await insertRewards(rewards);
+      const competition = await findById(competitionId);
+      if (!competition) {
+        throw new Error("Competition not found");
+      }
+
+      if (!competition.votingStartDate || !competition.votingEndDate) {
+        throw new Error("Voting start or end date not found");
```
joewagner, line 68: This is a nit for sure, but if we split this into two checks, it will be easier to debug, and the perf is basically the same.

packages/db/src/repositories/boost.ts
```diff
@@ -582,6 +582,66 @@ class BoostRepository {
...
+   *
+   * @param args - The query parameters
+   * @param args.competitionId - ID of the competition context
+   * @param tx - Optional database transaction to use for the query
+   * @returns Promise resolving to array of boost change records with agent information
+   */
+  async userBoostSpending(competitionId: string, tx?: Transaction) {
+    const executor = tx || this.#db;
+    return await executor
+      .select({
```
joewagner, line 616: I might not fully understand what the underlying query is here, but I came up with this as a way to count how many result this is going to return:
```
SELECT COUNT(*) AS count
FROM boost_changes bc
INNER JOIN boost_balances bb 
  ON bc.balance_id = bb.id
INNER JOIN agent_boosts ab 
  ON bc.id = ab.change_id
INNER JOIN agent_boost_totals abt 
  ON ab.agent_boost_total_id = abt.id
WHERE 
  bb.competition_id = '9e0140df-06b6-4fe1-a5e8-5bf523c09808'
  AND bc.delta_amount < 0;
```

When I run that on some data from the production app as of today (Monday), this shows 270,570 results.  This will probably be too big, and it will likely grow a lot in the future competitions.
Unless I've messed up the sql here, we need to do something else.

brunocalza: Yeah, thanks for the call-out. I've been thinking about this for some time. There's a chance this is not a real problem, at least for a couple of weeks. The query takes less than 1 second, and it is not executed by a user. I think we may have a problem in memory usage, on the NodeJS side. We would have 300k items in-memory. In any case, I'd like at least to get this feature merged and working so we can test the flow. And investigate potential performance issues in future issues.

joewagner: Ok, that works for me.  To be clear, `userBoostSpending` is only called by `calculateRewards` and I don't see `calculateRewards` getting called anywhere.  Before we totally commit to adding this, where/what is the planned caller?

brunocalza: Right now it's planned to be an admin call. That is being worked on https://github.com/recallnet/js-recall/pull/1309. If the HTTP call turns out to be a problem, it can be called by a script.

asutula: You could just sum the values from the `agent_boost_totals` table `where compeition_id = <id>`. There will always be i row for each agent competing, so not that many.

asutula (approved):
This looks pretty good to me. Left one comment about that sql query Joe commented on. We're looking at some conflicting changes between your pr and mine at #1303

packages/db/src/repositories/boost.ts
brunocalza: > There will always be i row for each agent competing, so not that many

I can't aggregate, unfortunately, because the time the person boosted is important to the calculation algorithm. There's a decay based on when the boost happened.

asutula: Ah right. ok ok.

PR #1255, opened 2025-09-19
Initialize boost upon link wallet

asutula:
Created a service that increments user boost for all competitions with an active voting window (we're repurposing the voting start and end columns for boost). This is called from the `linkWallet` handler in the user controller. The idea is when a user links their wallet, this new service will be called and they will have boost. The is for pre TGE only.


apps/api/src/config/index.ts
```diff
@@ -399,6 +399,12 @@ export const config = {
     // RPC provider URL for blockchain interactions
     rpcProvider: process.env.RPC_PROVIDER || "",
   },
+  boost: {
+    // Amount of boost (in wei) to grant on wallet linking pre TGE
+    initialBoostAmount: process.env.INITIAL_BOOST_AMOUNT
+      ? BigInt(process.env.INITIAL_BOOST_AMOUNT)
+      : 500000000000000000000n, // 500 boost by default
+  },
```
asutula, line 407: New env var to set how much boost we should allocate to users.

apps/api/src/controllers/user.controller.ts
```diff
@@ -146,6 +146,16 @@ export function makeUserController(services: ServiceRegistry) {
...
 
+        if (!linkedUser) {
+          throw new ApiError(
+            404,
+            "Failed to link wallet to user because no user found",
+          );
+        }
+
+        // Grant initial boost if applicable
+        await services.initBoostService.initBoost(linkedUser);
```
asutula: Here's the call in the controller to the new service.

apps/api/src/database/repositories/competition-repository.ts
```diff
@@ -128,6 +128,12 @@ export const findActive = createTimedRepositoryFunction(
   "findActive",
 );
 
+export const findVotingOpen = createTimedRepositoryFunction(
+  repository.findVotingOpen.bind(repository),
+  "CompetitionRepository",
+  "findVotingOpen",
+);
+
```
asutula, line 136: Needed this new repo function to return all comps with active boost windows.

apps/api/src/services/__tests__/init-boost.service.test.ts
asutula, line 1: Unit tests for the new service. Using `vitest-mock-extended` is much nicer and typesafe for mocks than just plain vitest.

apps/api/src/services/init-boost.service.ts
```diff
@@ -0,0 +1,79 @@
...
+
+type InitBoostResult = {
+  type: "boostGranted" | "alreadyGranted";
+  competitionId: string;
+  balance: bigint;
+};
+
+export interface InitBoostService {
+  initBoost(user: InsertUser): Promise<InitBoostResult[]>;
+}
```
asutula: Here is an interface so we can implement pre and post tge versions of this service. So far I've just implemented the pre tge version.

apps/api/src/services/init-boost.service.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export class PreTGEInitBoostService implements InitBoostService {
```
asutula: Pre tge implementation.

apps/api/src/services/init-boost.service.ts
```diff
@@ -0,0 +1,79 @@
...
+    this.#boostAmount = boostAmount;
+  }
+
+  async initBoost(user: InsertUser): Promise<InitBoostResult[]> {
+    return this.#db.transaction(async (tx) => {
+      const competitions = await this.#competitionRepo.findVotingOpen(tx);
+
+      if (competitions.length === 0) {
+        return [];
+      }
```
asutula: Get all comps with open boost window. Return empty results if none.

apps/api/src/services/init-boost.service.ts
```diff
@@ -0,0 +1,79 @@
...
+
+      return Promise.all(
```
asutula: For each competition...

apps/api/src/services/init-boost.service.ts
```diff
@@ -0,0 +1,79 @@
...
+            },
+            tx,
+          );
+          if (existingBalance > 0n) {
+            return {
+              type: "alreadyGranted",
+              competitionId: competition.id,
+              balance: existingBalance,
+            } as InitBoostResult;
+          }
```
asutula: Check if the user somehow has a boost balance for this competition. If so, return the appropriate result.

apps/api/src/services/init-boost.service.ts
```diff
@@ -0,0 +1,79 @@
...
+            tx,
+          );
+          return {
+            type: "boostGranted",
+            competitionId: competition.id,
+            balance:
+              increaseBoostRes.type === "applied"
+                ? increaseBoostRes.balanceAfter
+                : increaseBoostRes.balance,
+          } as InitBoostResult;
```
asutula: Else, increase the user's boost balance for this competition and return the appropriate result.

packages/db/src/repositories/competition.ts
```diff
@@ -974,6 +975,26 @@ export class CompetitionRepository {
...
+            gte(competitions.votingEndDate, now),
+          ),
+        );
+      return result;
+    } catch (error) {
+      this.#logger.error("Error in findVotingOpen:", error);
+      throw error;
+    }
+  }
+
```
asutula, line 998: New repo function to get comps open for boosting.

apps/api/src/config/index.ts
```diff
@@ -399,6 +399,12 @@ export const config = {
     // RPC provider URL for blockchain interactions
     rpcProvider: process.env.RPC_PROVIDER || "",
   },
+  boost: {
+    // Amount of boost (in wei) to grant on wallet linking pre TGE
+    noStakeBoostAmount: process.env.NO_STAKE_BOOST_AMOUNT
+      ? BigInt(process.env.NO_STAKE_BOOST_AMOUNT)
+      : undefined,
+  },
```
asutula, line 407: Noticed this config value can be `undefined`, signaling we are post TGE.

apps/api/src/controllers/user.controller.ts
```diff
@@ -146,6 +149,18 @@ export function makeUserController(services: ServiceRegistry) {
...
+        if (boostMode === "no-stake") {
+          await services.boostAwardService.initNoStake(
+            linkedUser.id,
+            linkedUser.walletAddress,
+          );
+        } else {
+          await services.boostAwardService.initForStake(
+            linkedUser.walletAddress,
+          );
+        }
```
asutula, line 162: Call the appropriate function.

apps/api/src/indexing/event-processor.ts
```diff
@@ -1,6 +1,8 @@
 import type { Logger } from "pino";
 import { decodeEventLog, hexToBytes } from "viem";
 
+import type { StakesRepository } from "@recallnet/db/repositories/stakes";
+
```
asutula, line 5: Moved the stakes repository to the shared package.

apps/api/src/services/__tests__/boost-award.service.test.ts
asutula, line 1: AI generated, skimmed over by me to make sure they're looking ok. I told it to use `vitest-mock-extended` for much better type safety in mocks.

apps/api/src/services/boost-award.service.ts
```diff
@@ -119,4 +140,82 @@ class BoostAwardService {
       tx,
     );
   }
+
+  async initForStake(
```
asutula, line 140: This function is for use post tge.

apps/api/src/services/boost-award.service.ts
asutula, line 1: The new functions I added in here call the existing award functions, but additionally query for all relevant competitions and stakes, iterating over those and calling the other/existing service functions.

apps/api/src/services/boost-award.service.ts
```diff
@@ -119,4 +140,82 @@ class BoostAwardService {
...
+              competitionId: competition.id,
+              stakeId: stake.id,
+            };
+          });
+        }),
+      );
+    });
+  }
+
+  async initNoStake(
```
asutula, line 186: And the pre-tge version.

apps/api/src/index.ts
```diff
@@ -154,7 +154,10 @@ const docsController = makeDocsController();
 const healthController = makeHealthController();
 const priceController = makePriceController(services);
 const tradeController = makeTradeController(services);
-const userController = makeUserController(services);
+const userController = makeUserController(
+  services,
+  config.boost.noStakeBoostAmount ? "no-stake" : "stake",
```
asutula, line 159: Configure the controller behavior based on the app config being defined or not.

packages/db/src/repositories/competition.ts
```diff
@@ -974,6 +976,26 @@ export class CompetitionRepository {
...
+    try {
+      const now = new Date();
+      const executor = tx || this.#db;
+      const result = await executor
+        .select()
+        .from(competitions)
+        .where(
+          and(
+            lt(competitions.votingStartDate, now),
+            gt(competitions.votingEndDate, now),
```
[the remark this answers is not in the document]
asutula: This is correct behavior. We want to ignore rows with null voting star or end dates.

packages/db/src/repositories/stakes.ts
```diff
@@ -552,4 +550,22 @@ class StakesRepository {
...
+      .select()
+      .from(schema.stakes)
+      .where(
+        and(
+          eq(schema.stakes.wallet, walletBytes),
+          isNull(schema.stakes.unstakedAt),
+        ),
+      )
+      .orderBy(asc(schema.stakes.id));
+  }
```
asutula, line 570: Return all active stakes for a wallet address.

asutula: Moved this file into the shared repository location.

asutula:
@stbrody please excuse the fact that I'm doing things the "old" way in this PR wrt logic in controllers, etc. Just tying to keep the PR scoped to the changes at hand and keep things as simple as possible. We'll clean this stuff up in another step/PR.

turbo.json
```diff
@@ -9,7 +9,8 @@
     "FONTS_REPO_ACCESS_TOKEN",
     "DATABASE_URL",
     "NODE_ENV",
-    "VITE_DATABASE_URL"
+    "VITE_DATABASE_URL",
+    "NO_STAKE_BOOST_AMOUNT"
```
dtbuchholz: shouldnt this be in the project/package specific turbo file, and not the root one?

asutula: Yea, i tried that in the past and had problems getting it to work. Do we have other examples of that? Been meaning to try that again and clean this all up.

apps/api/src/services/boost-award.service.ts
```diff
@@ -76,7 +93,7 @@ class BoostAwardService {
       // Staked after the voting starts
       multiplier = 1n;
     }
-    const boostAmount = (stake.amount * multiplier) / this.#divisor;
```
asutula, line 79: Storing boost with full precision.

apps/api/src/services/boost-award.service.ts
```diff
@@ -22,19 +22,36 @@ type CompetitionPosition = {
...
+    userService: UserService,
+    noStakeBoostAmount?: bigint,
+  ) {
+    this.#db = database;
+    this.#competitionRepo = competitionRepo;
     this.#boostRepository = boostRepository;
+    this.#stakesRepository = stakesRepository;
     this.#userService = userService;
-    this.#divisor = 10n ** BigInt(DECIMALS);
+    this.#noStakeBoostAmount = noStakeBoostAmount || 1000n;
```
dtbuchholz: why do we set a fallback amount here? if i'm reading the config correctly, if`NO_STAKE_BOOST_AMOUNT` env var is set, then `initNoStake` is used. it feels like the fallback value should happen in the config, and not the in this constructor. or, maybe not at all?

if you're in "stake" mode, this _technically_ doesn't matter. it's just a bit confusing.

i wonder if we could have a boolean env var `ENABLE_NO_STAKE_BOOST` dictate the "stake" vs "no-stake" behavior, or if that's overkill.

turbo.json
dtbuchholz: hm i wonder why? but we have [this file](https://github.com/recallnet/js-recall/blob/main/apps/api/turbo.json) in the `apps/api` dir, and it works for all of the backend related env vars.

asutula: Oh that's good news. I'll make an issue to clean them all up then.

apps/api/src/services/boost-award.service.ts
asutula: Yea good call, I can make it optional and throw an error if it isn't set.

turbo.json
asutula: Went ahead and moved the one introduced in this PR

PR #1256, opened 2025-09-22
fix(comps): api sandbox key fetching

dtbuchholz:
closes [APP-448](https://linear.app/recall-labs/issue/APP-448/fix-sandbox-api-keys)


apps/comps/app/api/sandbox/agents/route.ts
```diff
@@ -1,64 +1,34 @@
...
+import { AdminCreateAgentRequest, AdminCreateAgentResponse } from "@/types";
 
 /**
  * POST /api/sandbox/agents
- * Creates an agent in the sandbox environment by:
- * 1. Fetching user profile from the base API using session cookie to get walletAddress
- * 2. Creating agent in sandbox using the admin API
+ * Creates an agent in the sandbox environment using the provided user and agent data
  */
 async function handleCreateAgent(request: NextRequest) {
```
dtbuchholz, line 14: we save a call to the "main" prod API by using the user session context, which already has user profile info

apps/comps/app/api/sandbox/api-key/route.ts
```diff
@@ -23,6 +29,15 @@ async function handleGetAgentApiKey(request: NextRequest) {
     throw new Error("Authentication required");
   }
 
+  // Fetch user profile from the base API
```
dtbuchholz, line 32: note: this was the original code before i tried to fix this with https://github.com/recallnet/js-recall/pull/1224. the only difference here: we search by  `user.walletAddress` or `user.email`. this is needed because we must enforce any odd "sync" behavior with agents and users in the sanbdox

apps/comps/app/create-agent/page.tsx
```diff
@@ -40,27 +42,44 @@ function CreateAgentView() {
...
-          name: data.name,
-          handle: data.handle,
-          description: data.description,
-          imageUrl: data.imageUrl,
-          metadata: {
-            skills: data.skills,
-            repositoryUrl: data.repositoryUrl,
-            x: data.x,
-            telegram: data.telegram,
+          user: {
```
dtbuchholz, line 69: and this change is related to using the session instead of the "main" API requeat

apps/comps/hooks/useUnlockKeys.ts
```diff
@@ -19,6 +23,52 @@ export const useUnlockKeys = (agentHandle: string, agentId?: string) => {
     enabled: !!agentId,
   });
 
+  // Mutation to create sandbox agent if it doesn't exist
```
dtbuchholz, line 26: this logic sort of exists in the diff, but we pull it into its own mutation. it seems there are query / caching issues that were causing the sandbox API keys to not be found.

apps/comps/hooks/useUnlockKeys.ts
```diff
@@ -28,30 +78,12 @@ export const useUnlockKeys = (agentHandle: string, agentId?: string) => {
...
-          const { id, apiKey, name } = createResult.agent;
-          return {
-            success: true,
-            agent: {
-              id,
-              apiKey,
-              name,
-            },
-          } as AdminAgentKeyResponse;
+          return null;
```
dtbuchholz, line 86: and this was also the original behavior before https://github.com/recallnet/js-recall/pull/1224 landed

apps/comps/hooks/useUnlockKeys.ts
```diff
@@ -92,7 +133,27 @@ export const useUnlockKeys = (agentHandle: string, agentId?: string) => {
 
   const mutation = useMutation({
     mutationFn: async () => {
-      if (!sandboxAgentId) throw new Error("Unable to find sandbox agent");
+      let currentSandboxAgentId = sandboxAgentId;
+
+      // If sandbox agent doesn't exist, create it first
+      if (!currentSandboxAgentId) {
+        // Check if we need to refetch first (in case the agent was created elsewhere)
+        const refetchResult = await sandboxKeyQuery.refetch();
```
dtbuchholz, line 141: now, instead of trying to create a "missing" sandbox in the `sandboxKeyQuery`, we handling it in this mutation when the sandbox agent is not found. other queries that use this should refetch once the agent is created. i think should solve the "Unable to find sandbox agent" issue because we were running into cach and query race conditions.

dtbuchholz:
note: in general, i think we need to revisit our sandbox API key creation logic. it looks like the app frontend hasn't effectively managed the "sync" process well because i can see there are logs in vercel for non-existent sandbox agents.

this PR should improve some of existing problems. and hopefully, the serverless setup can make it less flaky. created a follow issue to spend more time on this if we still see problems: [APP-488](https://linear.app/recall-labs/issue/APP-488/refactor-prodsandox-sync-logic)

PR #1258, opened 2025-09-22
Convert atto boost to boost and vice versa in api client

asutula:
Allows us to have high precision accounting the db and easy to read UI.

### DIFF
--- a/apps/comps/lib/api-client.ts
+++ b/apps/comps/lib/api-client.ts
@@ -692,15 +692,17 @@ export class ApiClient {
   }
 }
 
+const attoBoostDivisor = 10 ** 18;
+
 function attoBoostStringToBoost(attoBoost: string) {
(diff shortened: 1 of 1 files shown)


apps/comps/lib/api-client.ts
```diff
@@ -682,11 +687,23 @@ export class ApiClient {
...
       ...res,
-      agentTotal: BigInt(res.agentTotal),
+      agentTotal: attoBoostStringToBoost(res.agentTotal),
     };
   }
 }
 
+function attoBoostStringToBoost(attoBoost: string) {
+  const res = dnum.div(attoBoost, 10 ** 18, {
+    rounding: "ROUND_DOWN",
```
asutula: Important to round down here so we don't ever let the user boost an amount greater than what their balance is.

PR #1259, opened 2025-09-22
chore: Move business logic out of TradeController into service layer

stbrody:
### Overview
This PR refactors the `TradeController` to follow our established pattern of keeping controllers as thin wrappers around service layer functions. All business logic has been moved from the controller into appropriate service classes, making the controller focused solely on request/response handling.

#### Commit 1: Move `getQuote` business logic to service layer
- Extracted all quote calculation logic from `TradeController.getQuote` into `TradeSimulatorService.getTradeQuote`
- Controller now only handles request parameter validation and response formatting
- Added proper TypeScript interfaces for quote parameters and results
- Extracted `SPECIFIC_CHAIN_NAMES` constant to avoid duplication between schemas

#### Commit 2: Move `executeTrade` business logic to new service
- Created new `SimulatedTradeExecutionService` to handle trade execution with competition validation
- Moved all business logic from `TradeController.executeTrade` including:
  - Competition validation (end date checks, agent registration checks)
  - Trade input validation
  - Chain resolution and cross-chain trading rules
  - Price fetching and constraint validation
  - Balance and portfolio limit checks
  - Trade calculation and database updates
- Updated `ServiceRegistry` to properly inject dependencies
- **Important**: Removed `executeTrade` entirely from `TradeSimulatorService` to avoid circular dependencies

### Architecture Decisions

1. **New Service Creation**: Created `SimulatedTradeExecutionService` instead of pushing logic into `TradeSimulatorService` to avoid circular dependency with `CompetitionService`

2. **Method Migration**: Moved all helper methods that were only used by the `executeTrade` call stack:
   - Trading constraint validation methods
   - Chain resolution and validation
   - Price fetching and validation
   - Balance and portfolio checks
   - Trade execution and database updates

3. **Clean Dependencies**: The new service depends on:
   - `CompetitionService` - for competition validation
   - `TradeSimulatorService` - for portfolio value calculations
   - `BalanceService` - for balance operations
   - `PriceTrackerService` - for price data

### Testing Updates
- Updated `trade-simulator.constraints.test.ts` to test `SimulatedTradeExecutionService` instead
- All existing tests pass with minimal changes (just service name updates)
- Maintained 100% backward compatibility

### Next Steps
After this PR, we should have a cleaner separation between HTTP handling (controllers) and business logic (services), making the codebase more maintainable and testable.


apps/api/src/services/simulated-trade-execution.service.ts
```diff
@@ -0,0 +1,843 @@
...
+}
+
+/**
+ * Service for executing simulated trades with competition validation
+ * Handles business logic for trade execution including competition checks
+ */
+export class SimulatedTradeExecutionService {
+  // Maximum percentage of portfolio that can be traded in a single transaction
+  private readonly maxTradePercentage: number = config.maxTradePercentage;
+  private readonly constraintsCache = new Map<string, TradingConstraints>();
```
asutula, line 60: We discussed not having caches in the service layer, just making sure this should be here.

asutula: Maybe this is just called "cache" but isn't the kind of caching we discussed keeping in the controller.

stbrody: hmm good point.  Looking at what's happening here, this is used in `getTradingConstraints`, which seems to have 2 possible sources for the constraints: the database and the config (set via env var).  Given constraints change very infrequently, I wonder why we even have this in the db at all and don't just simply use the config for this always?  We can't even change it at runtime without restarting the app server because of this caching anyway.

@dtbuchholz, what do you think about deleting this `trading_constraints` table entirely and always getting the constraints from the config?

dtbuchholz: we should keep that table. we want to have a fully stateless API, and each comp might have different config values. we started that work some time ago in https://github.com/recallnet/js-recall/pull/1092 but never got around to finishing it.

dtbuchholz: for reference: the original logic didn't have a `trading_constraints` table and only read from the config. we added that table because every comp is different, and we wanted to be able to see or easily update those values without needed to change env vars + stop/start the server.

stbrody: got it, makes sense.

Given that this is happening within executeTrade, our highest-volume endpoint (and the one that we aren't yet sure if we're going to move to serverless anyway), and the fact that the constraints generally won't change for a given competition, this seems like a good candidate for caching, and it feels like this is the right place for the trade constraint validation logic and therefore for the constraint caching.  I think it makes sense to leave this here, but I'm open to reconsidering if you feel strongly @asutula

dtbuchholz (approved):

apps/api/src/services/simulated-trade-execution.service.ts
asutula: Yea sounds reasonable to me.

PR #1265, opened 2025-09-23
chore: Cleanup types in AdminController

stbrody:


apps/api/src/controllers/admin.controller.ts
```diff
@@ -84,7 +69,7 @@ interface AdminAgentRegistrationResponse {
 
 interface AdminSearchResults {
   users: User[];
-  agents: Omit<Agent, "apiKey">[];
+  agents: AgentPublic[];
```
dtbuchholz, line 72: techncally, this will still expose the API key. i validated this in the `admin.test.ts` file for the case named `should search for users and agents based on various criteria`. 

tbh, that's probably okay here since it's an admin search result? we should change the type and OpenAPI spec, if we want to include the `apiKey` field.

apps/api/src/types/index.ts
```diff
@@ -560,6 +561,58 @@ export const AgentPublicSchema = AgentSchema.omit({
...
+    lastLoginAt: dbUser.lastLoginAt ?? undefined,
+  };
+}
+
+/**
+ * Converts a database Agent object to API Agent format
+ * @param dbAgent Database agent object with null values
+ * @returns API agent object with undefined values
+ */
+export function toApiAgent(dbAgent: SelectAgent): Agent {
```
dtbuchholz, line 598: or, maybe the "sanitize" step could happen here, and then we can return the `Agent` or `AgentPublic` as needed (some ai-gen'd code so take w grain of salt):
```ts
/**
 * Converts a database Agent object to API Agent format
 * @param dbAgent Database agent object with null values
 * @param includePrivateFields Whether to include private fields (apiKey, email)
 * @returns API agent object with undefined values
 */
export function toApiAgent(
  dbAgent: SelectAgent,
  includePrivateFields: true,
): Agent;
export function toApiAgent(
  dbAgent: SelectAgent,
  includePrivateFields: false,
): AgentPublic;
export function toApiAgent(dbAgent: SelectAgent): Agent;
export function toApiAgent(
  dbAgent: SelectAgent,
  includePrivateFields: boolean = true,
): Agent | AgentPublic {
  const baseAgent = {
    id: dbAgent.id,
    ownerId: dbAgent.ownerId,
    walletAddress: dbAgent.walletAddress ?? undefined,
    name: dbAgent.name,
    handle: dbAgent.handle,
    description: dbAgent.description ?? undefined,
    imageUrl: dbAgent.imageUrl ?? undefined,
    metadata: dbAgent.metadata ? (dbAgent.metadata as UserMetadata) : undefined,
    status: dbAgent.status as ActorStatus,
    createdAt: dbAgent.createdAt,
    updatedAt: dbAgent.updatedAt,
  };

  if (includePrivateFields) {
    return {
      ...baseAgent,
      apiKey: dbAgent.apiKey,
      email: dbAgent.email ?? undefined,
    } as Agent;
  } else {
    return {
      ...baseAgent,
      isVerified: !!dbAgent.walletAddress,
    } as AgentPublic;
  }
}

/**
 * Converts an array of database Agent objects to API format
 * @param dbAgents Database agent objects with null values
 * @param includePrivateFields Whether to include private fields (apiKey, email)
 * @returns API agent objects with undefined values
 */
export function toApiAgents(
  dbAgents: SelectAgent[],
  includePrivateFields: true,
): Agent[];
export function toApiAgents(
  dbAgents: SelectAgent[],
  includePrivateFields: false,
): AgentPublic[];
export function toApiAgents(dbAgents: SelectAgent[]): Agent[];
export function toApiAgents(
  dbAgents: SelectAgent[],
  includePrivateFields: boolean = true,
): Agent[] | AgentPublic[] {
  if (includePrivateFields) {
    return dbAgents.map((agent) => toApiAgent(agent, true));
  } else {
    return dbAgents.map((agent) => toApiAgent(agent, false));
  }
}
```

apps/api/src/controllers/admin.controller.ts
asutula: Does this mean there is a bug in that `sanitizeAgent` function? If so we should fix it.

asutula (commented):
Commented on that agent sanitization thing. Seems worth fixing.

apps/api/src/types/index.ts
stbrody: A previous version of the code as I was working on this PR had `toApiAgentPublic` function, I could put that back and make it include the sanitize step.  I guess I had assumed the fact that some endpoints return Agent and some return AgentPublic was intentional - are you saying that they should all return the redacted AgentPublic?

stbrody: So there are 3 functions in this file that use `toApiAgent` currently: registerUser, registerAgent, and listAllAgents.

There's definitely a comment in registerUser suggesting that the plaintext display of the apiKey is intentional so that the admin can view it.  I'm not sure about the other two, but since these are all admin endpoints it kind of makes sense that the admin may want to be able to view the api key in all of them?  I believe @KGmajor sometimes sends competitors their keys manually...

apps/api/src/controllers/admin.controller.ts
stbrody: @dtbuchholz can you explain a bit more what you were seeing?  I just pushed an update that expands that test case to include some checks that the apiKey is not returned, and the test is passing

dtbuchholz: in that test, try adding those three lines at the end

```ts
// TEST CASE 4: Search by status - all users should be 
// ...

// All three agents we created should be found as active
// ...
expect(user1Result.agent?.apiKey).toBeUndefined();
expect(user2Result.agent?.apiKey).toBeUndefined();
expect(user3Result.agent?.apiKey).toBeUndefined();
```
you'll get an error like this:
```
 FAIL   e2e  e2e/tests/admin.test.ts > Admin API > should search for users and agents based on various criteria
AssertionError: expected '96ddd5d45554ce40_fa9bcb9c521514e3' to be undefined
```

dtbuchholz: ooooh that's my bad. those variables are from the `registerUser` response, and not the admin search.

plz ignore me

dtbuchholz (approved):

PR #1269, opened 2025-09-24
chore: Encapsulate logic for resetting balance based on competition type within BalanceService

stbrody:


apps/api/src/controllers/admin.controller.ts
```diff
@@ -1657,7 +1657,10 @@ export function makeAdminController(services: ServiceRegistry) {
             `Resetting agent balance as part of applying sandbox mode logic for admin adding agent ${agentId} to competition ${competitionId}`,
           );
 
-          await services.balanceService.resetAgentBalances(agentId);
+          await services.balanceService.resetAgentBalances(
```
stbrody, line 1660: this is a behavior change.  As-is before this change, whenever an agent is added to the competition their balance gets set to the initial balances that are used for paper trading comps, even for perpetual futures comps.  I assume this is an improvement and the previous behavior could be considered a bug, but want to call this out in case it was intentional

mzkrasner: Great catch - yes, this is a bug fix.

apps/api/src/services/balance.service.ts
```diff
@@ -235,9 +241,18 @@ export class BalanceService {
       });
       this.balanceCache.set(agentId, balanceMap);
 
-      serviceLogger.debug(
-        `[BalanceManager] Successfully reset balances for agent ${agentId}`,
-      );
+      switch (competitionType) {
```
mzkrasner, line 248: Minor suggestion, but this switch statement could be set up to handle unexpected competition types so if a new competition type is added in the future, TS will catch at compile time.

apps/api/src/services/balance.service.ts
```diff
@@ -208,10 +208,14 @@ export class BalanceService {
   }
 
   /**
```
mzkrasner, line 211: Also minor, but tsdoc could be more explicit about behavior

```
/**
 * Reset an agent's balances based on competition type
 * @param agentId The agent ID
 * @param competitionType The competition type
 * 
 * Behavior by competition type:
 * - "trading": Resets to standard paper trading balances (5k USDC per chain)
 * - "perpetual_futures": Clears all balances (empty balance map)
 */
 ```

apps/api/src/controllers/admin.controller.ts
mzkrasner: Also - since this fixes a bug in the admin flow, we could add a test case to prevent regression

stbrody: done

apps/api/src/services/balance.service.ts
stbrody: done

apps/api/src/services/balance.service.ts
stbrody: done

stbrody (commented):
good suggestions!  Updated

PR #1271, opened 2025-09-24
chore: Move business logic from AdminController registerUser and registerAgent functions down into Service layer

stbrody:
Builds on https://github.com/recallnet/js-recall/pull/1265

### DIFF
--- a/apps/api/src/types/index.ts
+++ b/apps/api/src/types/index.ts
@@ -1080,7 +1080,7 @@ export const AdminCreateAgentSchema = z.object({
     walletAddress: z.string().optional(),
     description: z.string().optional(),
     imageUrl: z.string().optional(),
-    metadata: z.nullish(AgentMetadataSchema),
+    metadata: AgentMetadataSchema.optional(),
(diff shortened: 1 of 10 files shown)


apps/api/src/types/index.ts
```diff
@@ -1080,7 +1080,7 @@ export const AdminCreateAgentSchema = z.object({
     walletAddress: z.string().optional(),
     description: z.string().optional(),
     imageUrl: z.string().optional(),
-    metadata: z.nullish(AgentMetadataSchema),
+    metadata: AgentMetadataSchema.optional(),
```
stbrody, line 1083: this is technically a minor behavior change in the API, but the AI tells me there are no callers in the frontend that explicitly pass null here, and making this optional instead of nullish makes it much more consistent with the rest of the API and the internal types and avoids some extra null->undefined type conversion that is necessary otherwise.

asutula: Nullish seems weird in general. This is better.

PR #1272, opened 2025-09-25
Claim boost + boost polish and edge cases

asutula:


apps/api/src/services/competition.service.ts
```diff
@@ -247,6 +247,7 @@ type CompetitionDetailsData = {
       agentId: string | null;
     }>;
     votingEnabled?: boolean;
+    boostEnabled: boolean;
```
stbrody: nitpick but calling this "boostEnabled" doesn't feel quite right to me.  When I first saw this I thought it was going to access some competition-specific configuration property to see if the comp was configured to use boost or not.  Really this is more like `boostingWindowOpen` or `withinBoostingWindow` or something like that

stbrody (approved):
changes related to the new framework LGTM, though it would probably make sense to have other people more familiar review the frontend changes and the changes to the boost logic

apps/api/src/services/competition.service.ts
asutula: Will update.

PR #1277, opened 2025-09-25
Update trophy badge to be thinner 

timothyylim:
https://linear.app/recall-labs/issue/APP-476/trophies-are-too-wide


timothyylim:
Deployment may fail because I might not have the correct access rights to my Vercel user.

carsonfarmer:
> Deployment may fail because I might not have the correct access rights to my Vercel user.

Ya sorry about that. I had moved you to "viewer", but I've reverted that and re-run the actions. So you should be good to go now...

apps/comps/components/trophy-badge/index.tsx
```diff
@@ -22,22 +22,56 @@ interface TrophyBadgeProps {
   size?: number;
 }
 
+// =================================================================
+// ONE PLACE TO SCALE & CONFIGURE
```
stephhuynh18: Minor nit: this is just a style nit!!!! I'm  not a big fan of these comments. Also not sure why this has to be  in a config object. These can just be standalone constants.

carsonfarmer: +1

PR #1285, opened 2025-09-26
fix(ci): build services package

dtbuchholz:
Fixes docker build issue, which is missing the `@recallnet/services` and `@recallnet/conversions` packages


apps/api/trade-simulator-docker/Dockerfile
```diff
@@ -20,6 +20,7 @@ COPY . .
 RUN pnpm --filter=@recallnet/db build && \
```
dtbuchholz, line 20: do we even need to filter these out anymore? we've gotten rid of the legacy blockchain deps as well as the problematic/autogen'd `api-sdk` & `api-mcp`

edit: i suppose its useful still because we dont want the comps frontend app to install/build. or, maybe we should just change it to ignore those?

asutula: Yea i don't think filtering is needed. Build it all.

PR #1292, opened 2025-09-26
chore: Move logic for AdminController.endCompetition down into the CompetitionService

stbrody:


brunocalza (approved):
Looks good! Looks like you even fixed a bug in the end competitions script. It was not assigning the winners. It could mean the `competition_rewards` is missing some data

stbrody:
> Looks like you even fixed a bug in the end competitions script. It was not assigning the winners. It could mean the competition_rewards is missing some data

hmm... is that something we need to be worried about?  FYI @KGmajor @dtbuchholz

asutula (approved):

dtbuchholz:
> But should be an easy fix with a simple update.

created a ticket to update the db: https://linear.app/recall-labs/issue/APP-513/update-prod-db-competition-rewards-table-with-missing-values

PR #1300, opened 2025-09-28
Update leaderboard button styling for better visibility and differentiation

carsonfarmer:
Change previously discussed with Jovi


carsonfarmer:
They are almost identical, so probably doesn't matter. I'll stop picking these up so I don't trample your PR! I *personally* like the other one because it keeps the white border, but it's @stephhuynh18's call!

PR #1303, opened 2025-09-29
Move all services to shared package

asutula:


stbrody:
looks like everything perps related is failing, so there's probably a single root cause for all of that, possibly related to the issues @mzkrasner points out above.

Then it also looks like the format of the getCompetitionAgents endpoint changed - this PR should not be changing any API request/response structures.  With a change this big we really want to separate code motion from functional changes, so we should revert any functional changes to the service method implementations

apps/api/src/services/competition.service.ts
```diff
@@ -349,12 +310,7 @@ type CompetitionTradesData = {
   success: boolean;
   trades: TradeWithAgent[];
   competition: SelectCompetition;
-  pagination: {
```
stbrody, line 352: potential behavior change

packages/services/src/competition.service.ts
```diff
@@ -349,12 +310,7 @@ type CompetitionTradesData = {
   success: boolean;
   trades: TradeWithAgent[];
   competition: SelectCompetition;
-  pagination: {
-    total: number;
-    limit: number;
-    offset: number;
-    hasMore: boolean;
-  };
+  total: number;
```
stbrody, line 313: potential behavior change

packages/services/src/competition.service.ts
```diff
@@ -363,12 +319,7 @@ type CompetitionTradesData = {
 type AgentCompetitionTradesData = {
   success: boolean;
   trades: SelectTrade[]; // Agent trades use basic SelectTrade format
-  pagination: {
-    total: number;
-    limit: number;
-    offset: number;
-    hasMore: boolean;
-  };
+  total: number;
```
stbrody, line 322: potential behavior change

packages/services/src/competition.service.ts
```diff
@@ -3046,16 +3061,12 @@ export class CompetitionService {
         registeredParticipants: competition.registeredParticipants,
         maxParticipants: competition.maxParticipants,
         agents,
-        pagination: buildPaginationResponse(
-          total,
-          params.queryParams.limit,
-          params.queryParams.offset,
-        ),
+        total,
```
stbrody, line 3061: potential behavior change

packages/services/src/competition.service.ts
```diff
@@ -3273,16 +3284,12 @@ export class CompetitionService {
         success: true,
         trades,
         competition,
-        pagination: buildPaginationResponse(
-          total,
-          params.pagingParams.limit,
-          params.pagingParams.offset,
-        ),
+        total,
```
stbrody, line 3284: potential behavior change

packages/services/src/competition.service.ts
```diff
@@ -3325,16 +3332,12 @@ export class CompetitionService {
       const result = {
         success: true,
         trades,
-        pagination: buildPaginationResponse(
-          total,
-          params.pagingParams.limit,
-          params.pagingParams.offset,
-        ),
+        total,
```
stbrody, line 3332: potential behavior change

packages/services/src/competition.service.ts
```diff
@@ -295,12 +261,7 @@ type CompetitionAgentsData = {
     score: number | null;
     voteCount: number;
   }>;
-  pagination: {
-    total: number;
-    limit: number;
-    offset: number;
-    hasMore: boolean;
-  };
+  total: number;
```
stbrody, line 264: potential behavior change

apps/api/src/controllers/admin.controller.ts
```diff
@@ -323,6 +324,10 @@ export function makeAdminController(services: ServiceRegistry) {
                 },
           });
 
+        const activeCompetition =
```
stbrody, line 327: can we just use the `competition` that was returned above rather than re-querying the database here?

apps/api/src/controllers/competition.controller.ts
```diff
@@ -394,7 +394,14 @@ export function makeCompetitionController(services: ServiceRegistry) {
           queryParams,
         });
 
-        res.status(200).json(result);
+        res.status(200).json({
```
stbrody, line 397: Looking at the changes to the competition service, it looks like there are other endpoints that need the pagination response structure preserved: getCompetitionTrades and getAgentTradesInCompetition

apps/api/src/services/__tests__/perps-data-processor-orchestration.test.ts
```diff
@@ -1,1003 +0,0 @@
-import { beforeEach, describe, expect, it, vi } from "vitest";
```
stbrody, line 1: what happened to this test and the other tests in this directory?  They seem to just be deleted?

stbrody: oh I think they were moved but git got confused.  If you use `git mv` instead of regular `mv`, then git can track files across moves much better and it makes reviewing changes like this a lot easier

apps/api/src/types/index.ts
```diff
@@ -1,492 +1,4 @@
 import { Request } from "express";
```
stbrody, line 1: is this line the only line left in this file?  Should this file just be deleted?

packages/services/src/__tests__/boost-award.service.test.ts
```diff
@@ -75,6 +75,393 @@ describe("BoostAwardService", () => {
     mockReset(mockUserService);
   });
 
+  describe("awardForStake", () => {
```
stbrody: I'm surprised to see new tests showing up in this PR?  I guess that's fine though

apps/api/src/services/__tests__/rewards.service.test.ts
```diff
@@ -258,217 +211,20 @@ describe("RewardsService", () => {
...
 
       await expect(
         service.allocate(competitionId, tokenAddress, startTimestamp),
-      ).rejects.toThrow("Blockchain transaction failed");
-
-      // Verify transaction rollback - database operations should not have been called
-      expect(mockTransaction.insert).not.toHaveBeenCalled();
-    });
-
-    it("should maintain transaction integrity when database operations fail", async () => {
```
stbrody, line 280: why are these test cases deleted?

stbrody, line 280: why are these test cases deleted?

packages/services/src/__tests__/rewards.service.test.ts
```diff
@@ -258,217 +211,20 @@ describe("RewardsService", () => {
...
-          createdAt: new Date(),
-        },
-      ];
-
-      vi.mocked(
-        rewardsRepository.getRewardsTreeByCompetition,
-      ).mockResolvedValue(mockTree);
-
-      const service = new RewardsService();
+      ).rejects.toThrow("Blockchain allocation failed");
```
stbrody: blegh the new github code review tool isn't letting me comment on deleted lines, but it looks like a bunch of test cases in this file got deleted - was that intentional?

packages/services/src/providers/perps-provider.factory.ts
```diff
@@ -1,17 +1,18 @@
...
 
 /**
  * Factory for creating perps data providers based on configuration
  */
 export class PerpsProviderFactory {
+  static logger = pino.default();
   /**
    * Create a perps data provider based on the configuration
    */
   static createProvider(config: PerpsProviderConfig): IPerpsDataProvider {
```
stbrody: nitpick: the logger could be passed as an arg instead of statically initialized to the default.  Not sure if we configure out process-wide logger instances with anything other than the default, but if we did that would ensure that we don't lose that customization in the instances created by this factory

packages/services/src/types/index.ts
```diff
@@ -0,0 +1,1200 @@
+import { z } from "zod/v4";
```
stbrody, line 1: didn't actually review this file, I'm just assuming the types are strict code movement from the old types file in the `api` package.  If anything here actually changed please lmk

packages/services/src/configuration.service.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  constructor(competitionRepo: CompetitionRepository, logger: Logger) {
+    this.competitionRepo = competitionRepo;
+    this.logger = logger;
+  }
+  /**
+   * Load competition-specific settings and update global configuration
+   * This method updates the global features object with settings from the active competition
+   */
+  async loadCompetitionSettings(
```
stbrody: is this used anywhere?  I think this whole file might be dead code, in which case lets delete it

packages/services/src/lib/trade-utils.ts
```diff
@@ -74,11 +74,12 @@ export function calculateSlippage(fromValueUSD: number): SlippageResult {
...
-export const EXEMPT_TOKENS = new Set([
-  // Collect all token addresses from all chains in config
-  ...Object.values(config.specificChainTokens).flatMap((chainTokens) =>
-    Object.values(chainTokens),
-  ),
-  // Add the zero address for native EVM tokens
-  "0x0000000000000000000000000000000000000000", // Ethereum, etc...
-]);
+export const EXEMPT_TOKENS = (specificChainTokens: SpecificChainTokens) =>
+  new Set([
```
stbrody, line 78: it looks like this is now constructing the set every time we want to check inclusion in the set.  Given this is used a bunch in `SimulatedTradeExecutionService`  in the code path of executing a trade, our hottest code path, it seems like we should try to make it so that this set only gets constructed once at startup rather than on every trade

packages/services/src/__tests__/rewards.service.test.ts
stbrody: lol I guess those comments actually did get recorded the UI just wouldn't show them to me 😅 

How is the github code review tool actually getting worse?!

apps/api/src/controllers/admin.controller.ts
asutula: Yea it's kinda annoying because `services.competitionService.startOrCreateCompetition()` returns a competition object that doesn't contain the `crossChainTradingType` property which is needed. `services.competitionService.getActiveCompetition()` does return that property. This is the result of a (necessary and good) abstraction we made in the db where the core competitions table is general for any type of competition. We have another table, `trading_competitions` which extends that to add trading-related things to competitions. The issue is that we haven't built a clean API on top of those table that allows you to properly and efficiently interact with those different types of competitions. Things are all over the place at the moment.

asutula: I want to ask @mzkrasner if this behavior (updating the global config object depending on information from the active competition) is even still necessary, because it's pretty confusing and will certainly break if we have concurrent competitions.

stbrody: okay, that's fine, we can leave it for now and revisit later

apps/api/src/controllers/competition.controller.ts
asutula: Nice catch. Fixing.

apps/api/src/types/index.ts
asutula: No, the whole contents is:
```
import { Request } from "express";

/**
 * Extended Request interface for authenticated requests
 */
export interface AuthenticatedRequest extends Request {
  privyToken?: string;
  agentId?: string;
  userId?: string;
  adminId?: string;
  wallet?: string;
  isAdmin?: boolean;
  admin?: {
    id: string;
    name: string;
  };
}```

File: packages/services/src/__tests__/boost-award.service.test.ts (hunk 12 above)
Comment (asutula) [reply packages/services/src/__tests__/boost-award.service.test.ts]: Ah yea, I asked the AI to update the tests for the new service API, and it created new and more exhaustive tests. Decided to let it ride. Let me know if it's a problem.

File: packages/services/src/providers/perps-provider.factory.ts (hunk 15 above)
Comment (asutula) [reply packages/services/src/providers/perps-provider.factory.ts]: Yea static apis like this are weird. Pass a logger into the static function? Can do.

File: packages/services/src/types/index.ts (hunk 16 above)
Comment (asutula) [reply packages/services/src/types/index.ts:1]: Only additions were the following:

```
/**
 * Custom error class with HTTP status code
 */
export class ApiError extends Error {
  statusCode: number;

  constructor(statusCode: number, message: string) {
    super(message);
    this.statusCode = statusCode;
    this.name = this.constructor.name;
    Error.captureStackTrace(this, this.constructor);
  }
}

export type SpecificChainBalances = Partial<
  Record<SpecificChain, Record<string, number>>
>;

export type SpecificChainTokens = Partial<
  Record<SpecificChain, Record<string, string>>
>;
```
This moved the ApiError type from some middleware error handler file in the api app to this shared location, and adds two types to represent some data structures used in the global config object.

packages/services/src/lib/trade-utils.ts
asutula: Good call. Fixing.

packages/services/src/configuration.service.ts
asutula: Done.

PR #1306, opened 2025-09-29
feat(api): auto-start comps

dtbuchholz:
## Summary

adds an "auto start" competition cron job, allowing competitions to both start and end without manual requiring admin API calls. closes [APP-200](https://linear.app/recall-labs/issue/APP-200/autostart-competition)

### DIFF
--- a/apps/api/src/services/competition.service.ts
+++ b/apps/api/src/services/competition.service.ts
@@ -22,6 +22,7 @@ import {
   findById,
   findByStatus,
   findCompetitionsNeedingEnding,
+  findCompetitionsNeedingStarting,
   findLeaderboardByTradingComp,
   get24hSnapshots,
   getAgentCompetitionRecord,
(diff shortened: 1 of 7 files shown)


dtbuchholz:
note: this is mostly ai-gen'd code, with some tweaks after a rebase. i tested it in local dev, and it works.

apps/api/src/services/competition.service.ts
```diff
@@ -2265,6 +2266,107 @@ export class CompetitionService {
...
+        serviceLogger.debug(
+          "[CompetitionManager] No competitions ready to start",
+        );
+        return;
+      }
+
+      // We only support running one competition at a time, so we will not start any competitions
+      // if we find more than one. Note: This should not happen if competitions are created with
+      // the correct start dates; it's defensive.
+      if (competitionsToStart.length > 1) {
```
dtbuchholz: i figure we want to be extra careful, and if there is more than one "ready" competition (in a `pending` state), that's probably due to human error with dates. thus, we just skip all comps and don't start any of them.

packages/db/src/repositories/competition.ts
```diff
@@ -2213,6 +2213,46 @@ export class CompetitionRepository {
...
+        .from(tradingCompetitions)
+        .innerJoin(
+          competitions,
+          eq(tradingCompetitions.competitionId, competitions.id),
+        )
+        .where(
+          and(
+            eq(competitions.status, "pending"),
+            eq(competitions.sandboxMode, false),
+            isNotNull(competitions.startDate),
```
dtbuchholz, line 2238: and the query itself only grabs comps that have a start date

apps/api/src/services/competition.service.ts
```diff
@@ -2265,6 +2266,100 @@ export class CompetitionService {
...
+        serviceLogger.warn(
+          `[CompetitionManager] Multiple competitions ready to start (${competitionsToStart.length}). Skipping auto-start checks`,
+        );
+        return;
+      }
+      serviceLogger.debug(
+        `[CompetitionManager] Found ${competitionsToStart.length} competitions ready to start`,
+      );
+
+      for (const competition of competitionsToStart) {
```
brunocalza: There's no need to be a for-loop. `competitionsToStart` only holds one item

apps/api/src/services/competition.service.ts
```diff
@@ -2265,6 +2266,100 @@ export class CompetitionService {
...
+        // Re-check no active competition right before attempting to start (race-safety)
+        const currentlyActive = await findActive();
```
brunocalza: Hum, this technically does not prevent race conditions. An admin could start a competition, right after this query. I am not even sure how much this helps, given we already did a `findActive` at the start of this method. If we truly want to avoid race conditions, we'd need some kind of lock mechanism. Or we could add some constraints in the db to avoid two active competitions (although this would be temporary because we want to support that in the future)

also, `startCompetition` is already doing the `findActive`. I think removing all `findActive` in this method and adding a lock on `startCompetition` would solve the potential race condition problems

apps/api/src/services/competition.service.ts
```diff
@@ -2265,6 +2266,100 @@ export class CompetitionService {
...
+
+        // Skip sandbox competitions (defensive - repository already filters)
+        if (competition.sandboxMode) {
+          serviceLogger.debug(
+            `[CompetitionManager] Skipping sandbox competition: ${competition.name} (${competition.id})`,
+          );
+          continue;
+        }
+
+        try {
```
brunocalza: nit: nested try-catch

apps/api/src/services/competition.service.ts
dtbuchholz: fair points, implemented these changes and your other comments; i've simplified things a bit.

> adding a lock on startCompetition would solve the potential race condition problems

i can certainly do this, but i think this would add a bit of complexity to `startCompetition`, which makes a bunch of different queries via function calls. given we call this method carefully and only once per week, i'm not sure if it's worth implementing in this PR?

apps/api/src/services/competition.service.ts
```diff
@@ -611,17 +612,18 @@ export class CompetitionService {
...
    */
   private async validateAgentIds(agentIds: string[]): Promise<string[]> {
     const validAgentIds: string[] = [];
     const invalidAgentIds: string[] = [];
 
-    for (const agentId of agentIds) {
-      const agent = await this.agentService.getAgent(agentId);
-      if (!agent || agent.status !== "active") {
-        invalidAgentIds.push(agentId);
+    const agents = await this.agentService.getAgentsByIds(agentIds);
```
dtbuchholz, line 621: i noticed we had an existing method `getAgentsByIds` which would effectively do the same thing instead of calling `getAgent` for every registered agent

apps/api/src/services/competition.service.ts
```diff
@@ -664,7 +666,7 @@ export class CompetitionService {
    */
   async startCompetition(
     competitionId: string,
-    agentIds: string[],
+    agentIds?: string[],
```
dtbuchholz, line 669: and this has bugged me for a while: there's no reason `startCompetition` should require `agentId` values. the steps below already fetch the active agent IDs via `getPreRegisteredAgentIds`., so, all we need to do is validate the IDs beforehand (that method will throw an error if any are invalid agent UUIDs) and then merge the lists together.

apps/api/src/services/competition.service.ts
```diff
@@ -2265,6 +2267,84 @@ export class CompetitionService {
...
+      await this.startCompetition(competition.id);
+      serviceLogger.debug(
+        {
+          competitionId: competition.id,
+          name: competition.name,
+        },
+        `[CompetitionManager] Successfully auto-started competition`,
+      );
+    } catch (error) {
+      // Continue silently if the competition has no registered nor provided agents
```
dtbuchholz, line 2335: i _think_ the silent failure is how we want to handle it here? we outsource the "check if there are any agents in this comp" to the `startCompetition` method, which throws a 400 if there are no registered agent.

it saves us a query.

apps/api/src/services/competition.service.ts
brunocalza: yeah agree, probably fine

brunocalza (approved):

sam701:
> cc @sam701 we'll want to set up a container in both prod & staging "main" envs that run exactly like the existing "auto end competitions" cron. this can use the pnpm run cron:auto-start-competitions script. what's the process for doing setting this up?

@dtbuchholz it would be a PR in `competitions-app-deployments`. I'll create one and add you to the reviewers.

PR #1309, opened 2025-09-29
feat: add admin endpoint to trigger rewards

brunocalza:
# Context

Implements the following admin call where you can trigger the rewards calculation and allocation. The rewards become available after `startTimestamp`
```bash
curl -X POST "http://localhost:3000/api/admin/rewards/allocate"  \
 -H "Content-Type: application/json"   \
 -H "Authorization: Bearer your-admin-token"   \ 
 -d '{
    "competitionId": "9e0140df-06b6-4fe1-a5e8-5bf523c09808",
    "startTimestamp": 1704067200
  }'
```

[APP-379](https://linear.app/recall-labs/issue/APP-379/rewards-allocation-trigger)

This was tested with production database ( > 600k boosts). This call takes 22 seconds to execute. 

# Details
- For the above to work, we added a new table `competition_prize_pools` that stores the prize pools 
- We add a new route `/admin/rewards/allocate`
- We add a new controller method `allocateRewards`
- We add a new service method `calculateAndAllocate`


apps/api/drizzle/0048_lush_overlord.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE "competition_prize_pools" (
```
brunocalza, line 1: new table stores the two prize pools in WEI. we still don't have a way to update these value. It will be done in future PR

apps/api/drizzle/0048_lush_overlord.sql
```diff
@@ -0,0 +1,15 @@
...
+	"id" uuid PRIMARY KEY NOT NULL,
+	"competition_id" uuid NOT NULL,
+	"agent_pool" numeric(78, 0) NOT NULL,
+	"user_pool" numeric(78, 0) NOT NULL,
+	CONSTRAINT "competition_prize_pools_competition_id_key" UNIQUE("competition_id")
+);
+--> statement-breakpoint
+ALTER TABLE "rewards" ADD COLUMN "user_id" uuid NOT NULL;--> statement-breakpoint
```
brunocalza, line 9: added `user_id` and `agent_id` to the `rewards` table so we can track

apps/api/e2e/tests/event-processor.test.ts
```diff
@@ -1,7 +1,7 @@
 import { eq } from "drizzle-orm";
```
brunocalza, line 1: nothing relevant in this file. just adjusting the tests to new rewards columns

apps/api/e2e/tests/rewards.test.ts
```diff
@@ -6,7 +6,7 @@ import { beforeEach, describe, expect, test, vi } from "vitest";
 import { BoostRepository } from "@recallnet/db/repositories/boost";
```
brunocalza, line 6: nothing relevant in this file. just adjusting the tests to new rewards columns

packages/staking-contracts/src/rewards-allocator.ts
```diff
@@ -38,11 +38,13 @@ class RewardsAllocator {
   private walletClient: WalletClient;
   private publicClient: PublicClient;
   private contractAddress: Hex;
+  private tokenAddress: Hex;
 
   constructor(
```
brunocalza, line 50: minor change in this class to accept `tokenAddress` in the constructor and not in `allocate` method. this changes avoid the admin passing `tokenAddress`  in the http call

packages/services/src/rewards.service.ts
```diff
@@ -94,7 +122,8 @@ export class RewardsService {
       const boostAllocations: BoostAllocation[] = boostSpendingData.map(
         (entry) => {
           return {
-            user: bytesToHex(entry.wallet) as string,
+            user_id: entry.userId,
```
brunocalza, line 133: here's where we add the new columns

packages/services/src/rewards.service.ts
```diff
@@ -110,18 +139,24 @@ export class RewardsService {
...
+        agentId: reward.competitor,
+        competitionId: competitionId,
+        address: reward.address,
+        amount: reward.amount,
+        leafHash: hexToBytes(
+          createLeafNode(reward.address as Hex, reward.amount),
+        ),
+        id: crypto.randomUUID(),
+      }));
+      await runWithConcurrencyLimit(
```
brunocalza, line 161: the number of rewards can get pretty big, made a change to run the inserts concurrently using `runWithConcurrencyLimit` helper

packages/services/src/rewards.service.ts
```diff
@@ -383,3 +433,45 @@ function areUint8ArraysEqual(a: Uint8Array, b: Uint8Array): boolean {
   }
   return true;
 }
+
+function* chunkGenerator<T>(array: T[], size: number): Generator<T[]> {
```
brunocalza, line 498: it generates chunks on the fly so we don't have to keep all chunks in memory

packages/services/src/rewards.service.ts
```diff
@@ -383,3 +433,45 @@ function areUint8ArraysEqual(a: Uint8Array, b: Uint8Array): boolean {
...
+  for (let i = 0; i < array.length; i += size) {
+    yield array.slice(i, i + size);
+  }
+}
+
+async function runWithConcurrencyLimit<T>(
```
brunocalza, line 504: This function launches a configurable (`concurrency`) number of concurrent tasks. When one finishes, another one starts, keeping the concurrency fixed.

packages/services/src/rewards.service.ts
```diff
@@ -328,19 +360,37 @@ export class RewardsService {
     );
 
     // in case an address is both a voter and a competitor, we need to sum the amounts
+    // while keeping the references to owner and competitor
     const rewards = [...userRewards, ...competitorRewards];
     const rewardsByAddress = rewards.reduce(
       (acc, reward) => {
-        acc[reward.address] = (acc[reward.address] || 0n) + reward.amount;
+        if (!acc[reward.address]) {
```
brunocalza, line 428: this is a change hard to follow. but it's just keeping track of additional data (user id, agent id) as we sum the rewards

apps/api/src/routes/admin.routes.ts
```diff
@@ -2337,5 +2337,65 @@ export function configureAdminRoutes(
...
+   *                   description: The competition ID that rewards were allocated for
+   *                   example: "12345678-1234-1234-1234-123456789012"
+   *       400:
+   *         description: Bad Request - Invalid request format or missing required parameters
+   *       401:
+   *         description: Unauthorized - Admin authentication required
+   *       500:
+   *         description: Server error
+   */
+  router.post("/rewards/allocate", controller.allocateRewards);
```
brunocalza, line 2398: new route

packages/services/src/rewards.service.ts
```diff
@@ -51,15 +45,46 @@ export class RewardsService {
...
+   * @param competitionId The competition ID (UUID) to calculate and allocate rewards for
+   * @param startTimestamp The timestamp from which rewards can be claimed
+   * @throws Error if reward calculation fails or no rewards exist to allocate
+   */
+  public async calculateAndAllocate(
+    competitionId: string,
+    startTimestamp: number,
+  ): Promise<void> {
+    const prizePool =
+      await this.competitionRepository.getCompetitionPrizePools(competitionId);
```
brunocalza, line 64: We get the prize pools from the database

packages/services/src/rewards.service.ts
```diff
@@ -51,15 +45,46 @@ export class RewardsService {
...
+    if (!prizePool) {
+      throw new Error(`No prize pool found for competition ${competitionId}`);
+    }
+
+    await this.calculateRewards(
```
brunocalza: Here we calculate the rewards using the `rewards` package and store them in the database table `rewards` (this method already existed from previous work)

packages/services/src/rewards.service.ts
```diff
@@ -51,15 +45,46 @@ export class RewardsService {
...
+      competitionId,
+      prizePool.userPool,
+      prizePool.agentPool,
+    );
+    await this.allocate(competitionId, startTimestamp);
```
brunocalza: Here we build the merkle tree and make the contract call (this method already existed from previous work)

packages/services/src/rewards.service.ts
```diff
@@ -328,19 +360,37 @@ export class RewardsService {
...
+            reward.amount,
+            reward.owner,
+            reward.competitor,
+          ];
+          return acc;
+        }
+
+        acc[reward.address] = [
+          acc[reward.address]![0] + reward.amount,
+          reward.owner,
```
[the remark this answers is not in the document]
brunocalza: not a problem

packages/rewards/src/index.ts
```diff
@@ -79,30 +79,41 @@
...
+      userIds[user_wallet] = user_id;
     }
 
     const effectiveBoost = boostDecayFn(timestamp).times(
       new Decimal(boost.toString()),
     );
 
-    userTotals[user]![competitor] = (
-      userTotals[user]![competitor] || new Decimal(0)
+    userTotals[user_wallet]![competitor] = (
```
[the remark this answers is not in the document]
dtbuchholz: seems like this _could_ be useful? will defer to you here, idk if it's necessary

apps/api/drizzle/0048_lush_overlord.sql
dtbuchholz: just curious: why are agent IDs nullabe, but users are not?

is the idea that users will always be the ones who actually claim rewards because they're the wallet holders? and agent are more like optional claimable constraints (like user + agent ID) / metadata?

brunocalza: yes. `user_id` and `agent_id` are just metadata. What matters is the wallet, and that wallet comes from the user that owns the agent. the `agent_id` is `null` for the boosters/voters rewards and `not null` for agent/competitors rewards

brunocalza: I added `user_id` and `agent_id` to keep track of whom that rewards belongs to. if we eventually support changing wallets, without `user_id` we wouldn't know to whom that reward went

packages/rewards/src/index.ts
```diff
@@ -177,6 +194,8 @@ export function calculateRewardsForCompetitors(
     rewards.push({
       address: walletAddress,
       amount: BigInt(split.toFixed(0, Decimal.ROUND_DOWN)),
+      owner: competitorToOwnerMap[competitor]!,
```
dtbuchholz: [nit] we should avoid the `!`, and we can assign `competitor` directly instead of `competitor: competitor`

packages/services/src/rewards.service.ts
```diff
@@ -110,18 +147,24 @@ export class RewardsService {
...
-          address: reward.address,
-          amount: reward.amount,
-          leafHash: hexToBytes(
-            createLeafNode(reward.address as Hex, reward.amount),
-          ),
-          id: crypto.randomUUID(),
-        })),
+      const rewardsToInsert = rewards.map((reward) => ({
+        userId: reward.owner,
+        agentId: reward.competitor,
```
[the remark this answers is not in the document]
dtbuchholz: (maybe relevant, idk)

packages/services/src/rewards.service.ts
```diff
@@ -203,14 +248,13 @@ export class RewardsService {
 
     const rootHash = merkleTree.getHexRoot();
 
-    await this.db.transaction(async (tx) => {
+    const executeWithTransaction = async (transaction: Transaction) => {
       this.logger.info(
         `[RewardsService] Publishing root hash ${rootHash} to blockchain for competition ${competitionId}`,
       );
 
       const result = await this.rewardsAllocator!.allocate(
```
dtbuchholz: unrelated to these changes—but we can remove the `!`. it doesn't have any effect here

dtbuchholz (approved):
lgtm

PR #1311, opened 2025-09-29
feat: Handle global leaderboard sort and pagination in database

stbrody:
Removed support for global leaderboard sort options other than "rank" since they weren't exposed in the UI anyway, then pushed the sort on rank, alongside the pagination options (limit + offset), down into the SQL query so they no longer happen in memory.

There are also some small renames and code cleanups - you can view the first commit in the stack separately from the others to separate the functional changes from the code cleanups and make the PR a bit less noisy to review


apps/api/e2e/tests/leaderboard.test.ts
```diff
@@ -366,241 +366,4 @@ describe("Leaderboard API", () => {
     expect(agents[0]?.voteCount).toBe(3);
     expect(agents[1]?.voteCount).toBe(1);
   });
-
-  test("should sort global leaderboard by different fields", async () => {
```
dtbuchholz, line 370: why did we remove this test? we need to be able to sort the `/leaderboard` endpoint with query params, and it looks like you removed the `sort` feature altogether? at the very least, sort by rank/score

dtbuchholz: for context: our original leaderboards had all columns sortable. imo the "new" leaderboard designs regressed and removed that sorting feature. we should definitely still allow the existing sorting params.

stbrody: yeah, there was a bit of a discussion about this on slack.  I saw that no custom sorting was exposed at all in the UI so it seemed safe to remove it.  Supporting the different sorts while also preserving "rank" calculation and pushing the pagination into the DB was going to be complicated so it didn't seem worth it if it isn't even being used.  But I can put it back if we think its really important, but its going to add some complexity to the code and extra load to the db.

The trickiest ones are if the sorts on other properties other than rank/score, since then calculating the rank to display in the UI is tricky since rank is based on a global sort by score.  If we don't need to display rank when sorting by other properties then that would be easier (although still somewhat tricky to sort on the computed properties like number of votes), but if we need to preserve "rank" calculation when sorting by things like agent name, number of competitions, or number of votes, then we'll need a db query that runs over *all* registered agents in the whole db, not just the ones in the query result set, to calculate the rank, which will add load to the db, especially if the number of registered agents grows large.

Might need to loop in @KGmajor and @jovirecall here on what sorting options we actually want to expose

dtbuchholz: for reference, other sites with leaderboards:
- allows sorting: https://lmarena.ai/leaderboard
- only shows tops: https://openrouter.ai/rankings

stbrody:
okay, given discussion on Slack, okay to merge this @dtbuchholz ?

PR #1318, opened 2025-09-30
fix: remove annualization, fix rounding

mzkrasner:
# Summary

- Removes annualization altogether in calmar service (api)
- Ensure rounding for risk metrics in frontend


dtbuchholz:
if we run a 2 month comp, would it make sense to keep it? a 6 week comp? a 31 day comp?

i dont know the financial specifics here, but imo yeah, it feels like we just dont need to annualize it ever.

PR #1328, opened 2025-10-01
feat: Get user profile via serverless function

stbrody:


apps/comps/providers/session-provider.tsx
```diff
@@ -145,21 +146,23 @@ export function SessionProvider({ children }: { children: React.ReactNode }) {
...
     enabled: authenticated && ready && isLoginToBackendSuccess,
     staleTime: 5 * 60 * 1000, // 5 minutes
     gcTime: 10 * 60 * 1000, // 10 minutes (formerly cacheTime)
-    // TODO: Use a client that has better error types that include status codes.
-    // retry: (failureCount, error) => {
-    //   // Don't retry on 401/403 errors
-    //   if (error?.status === 401 || error?.status === 403) return false;
-    //   return failureCount < 3;
-    // },
+    retry: (failureCount, error) => {
```
stbrody: I don't fully understand why this code was commented out in the first place and I'm not totally sure it makes sense to uncomment it now, but my AI seemed to think so 😅.

@dtbuchholz what do you think?

apps/comps/rpc/router/user/get-profile.ts
```diff
@@ -0,0 +1,56 @@
...
+      const user = await context.userService.getUser(context.user.id);
+
+      if (!user) {
+        throw errors.NOT_FOUND();
+      }
+
+      return serializeUser(user);
+    } catch (error) {
+      // Re-throw if it's already one of our defined errors
+      if (error && typeof error === "object" && "code" in error) {
```
stbrody: not sure if there might be a better way to do this check for expected errors?

apps/comps/providers/session-provider.tsx
```diff
@@ -145,21 +146,23 @@ export function SessionProvider({ children }: { children: React.ReactNode }) {
   } = useQuery({
     queryKey: ["user"],
     queryFn: async () => {
-      const response = await apiClient.current.getProfile();
-      if (!response.success) {
-        throw new Error("Failed to fetch user");
-      }
-      return response.user;
+      return await rpcClient.user.getProfile();
```
asutula: Cool so you're using the vanilla client here. That is fine, but you could also leverage the tanstack client and that would replace this argument to `useQuery`. Just a little more simple.

apps/comps/app/rpc/[[...rest]]/route.ts
```diff
@@ -15,7 +14,7 @@ async function handleRequest(request: Request) {
       cookies: await cookies(),
       privyClient,
       boostService,
-      db,
```
Comment (asutula) [inline apps/comps/app/rpc/[[...rest]]/route.ts:18]: Nice to not pass the db in any more.

apps/comps/rpc/router/user/get-profile.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export const getProfile = base
+  .use(authMiddleware)
+  .handler(async ({ context, errors }) => {
+    try {
+      const user = await context.userService.getUser(context.user.id);
+
+      if (!user) {
+        throw errors.NOT_FOUND();
+      }
```
asutula: Ok, so cool thing is the user you are querying already exists at `context.user`, so you don't have to query for it again. Just return it. And it will always be found because you used the auth middleware on this procedure, so for the handler to even run implies the user must exist.

apps/comps/rpc/router/user/get-profile.ts
```diff
@@ -0,0 +1,56 @@
...
+
+      return serializeUser(user);
```
asutula: One cool feature about oRPC is that it takes care of all serialization for you (and deserialization on the client side). So sending things like date instances is totally fine. `null` is also fine, so really you don't need your serialize function at all.

apps/comps/rpc/router/user/get-profile.ts
```diff
@@ -0,0 +1,56 @@
...
+      // Re-throw if it's already one of our defined errors
+      if (error && typeof error === "object" && "code" in error) {
+        throw error;
+      }
+
+      // Wrap unexpected errors as INTERNAL
+      throw errors.INTERNAL({
+        message:
+          error instanceof Error ? error.message : "Failed to get user profile",
+      });
```
asutula: I think these oRPC errors have a `cause` property where you can pass along the inner error. Could be wrong but may be worth checking.

apps/comps/rpc/router/user/get-profile.ts
asutula: Will have to play around with this a bit.

apps/comps/lib/services.ts
asutula, line 1: I know wiring up these dependency trees can get tedious and hard to follow once they get large (although there are tools to make that better if we ever need), but this is so nice to see how everything clicks together and to know there are no hidden dependencies sneaking into the implementation.

apps/comps/rpc/router/user/get-profile.ts
asutula: https://orpc.unnoq.com/docs/rpc-handler#supported-data-types

apps/comps/rpc/router/user/get-profile.ts
stbrody: oh yeah.  I totally knew that and did this the right way in an earlier version of this PR, but then forgot about it this time around 🤦.  Good call

apps/comps/rpc/router/user/get-profile.ts
stbrody: So it looks like if I remove this then we wind up changing the types that the frontend receives.  Currently dates come back as ISO strings and null fields get converted to undefined and omitted entirely from the response.  Using oRPC to fully serialize the `SelectUser` type means dates get to the frontend as Date objects and nulls are preserved.  Is the idea to update the frontend to use the `SelectUser` type directly?  We'll have a mix for while as we migrate, since some places still need the old `User` type where they are still using the old API.

stbrody: Hmm, I tried making the change to use `SelectUser` in the frontend directly, but now there's a bunch of places littered throughout the frontend code that needs to map the `null` fields back to `undefined`.  TBH I think its cleaner to keep the explicit serialization.  Maybe once *everything* is switched over to oRPC then we can remove the serialization in one fell swoop for everything once there won't be type mismatches between the oRPC endpoints and the old API REST endpoints.

apps/comps/providers/session-provider.tsx
```diff
@@ -142,25 +143,26 @@ export function SessionProvider({ children }: { children: React.ReactNode }) {
...
-    // TODO: Use a client that has better error types that include status codes.
-    // retry: (failureCount, error) => {
-    //   // Don't retry on 401/403 errors
-    //   if (error?.status === 401 || error?.status === 403) return false;
-    //   return failureCount < 3;
-    // },
-  });
+  } = useQuery(
+    tanstackClient.user.getProfile.queryOptions({
+      queryKey: ["user"],
```
asutula: You can delete this. The tanstackClient handles query keys internally by default.

apps/comps/providers/session-provider.tsx
dtbuchholz: (looks like the diff no longer has that commented out?). edit: ignore. i looked at the wrong file/commit

hm i cant remember, seems fine to uncomment

PR #1330, opened 2025-10-01
Don't fetch leaderboard data if it won't be displayed

asutula:
We currently don't display the leaderboard carousel (it's controlled by an env var), but the fetch of the leaderboard happens anyway. This is on the main landing page, so this happens every time someone visits the site. The change will only load the data if it will be displayed.

### DIFF
--- a/apps/comps/app/competitions/page.tsx
+++ b/apps/comps/app/competitions/page.tsx
@@ -35,6 +35,7 @@ export default function CompetitionsPage() {
   const { data: leaderboard, isLoading: isLoadingLeaderboard } =
     useLeaderboards({
       limit: 25,
+      enabled: !DISABLE_LEADERBOARD,
     });
   const session = useSession();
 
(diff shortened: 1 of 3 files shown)


dtbuchholz (approved):
lgtm

some background: we originally set this up because we had to temporarily hide the agent rank global leaderboard. we released the agent rank leaderboard, but then we realized the landing page called the `/leaderboard` endpoint (a costly query).

the only reason we call `/leaderboard` on the landing is for the carousel at the top, which is designed to show _some_ sort of "cool" information (like live trades or something). we didn't have any API designed for the carousel, so we decided to use temporarily use the response from `/leaderboard` and simply display the agent ranks. 

all of that's to say....the actual fix is probably:
1. get rid of the carousel. in a world of multiple competitions, it's probably confusing to have this. and we have never displayed it in prod, so there's no feature loss here.
2. on the landing / comps hub page, use some other data that isn't the `/leaderboard` response, and then we can remove the `useLeaderboards` hook.

PR #1331, opened 2025-10-01
feat: updateProfile via serverless function

stbrody:


apps/comps/components/update-profile.tsx
```diff
@@ -42,36 +40,18 @@ export const UpdateProfile: React.FC<UpdateProfileProps> = ({ onSubmit }) => {
...
-        form.setError("email", {
-          type: "manual",
-          message: error.message,
-        });
-      } else {
-        // Re-throw other errors to be handled by parent
-        throw error;
-      }
-    }
+    const transformedData: UpdateProfileRequest = {
```
stbrody, line 49: the frontend doesn't actually let you update email, and the backend was ignoring the email field if provided to update user.  So this is just deleting effectively-dead code in support of a feature we don't actually have (editing the email of an existing user).

apps/comps/providers/session-provider.tsx
```diff
@@ -214,48 +214,40 @@ export function SessionProvider({ children }: { children: React.ReactNode }) {
...
-      if (!authenticated) throw new Error("Not authenticated");
-      const response = await apiClient.current.updateProfile(updates);
-      if (!response.success) {
-        throw new Error("Failed to update user");
-      }
+  } = useMutation(
+    tanstackClient.user.updateProfile.mutationOptions({
+      onMutate: async (updates) => {
+        // Cancel outgoing refetches so they don't overwrite our optimistic update
+        await queryClient.cancelQueries({ queryKey: ["user"] });
```
stbrody: not sure if the tanstackClient has any magic way to detect the queryKey here too?

apps/comps/rpc/router/user/update-profile.ts
```diff
@@ -0,0 +1,53 @@
...
+  })
+  .handler(async ({ input, context, errors }) => {
+    try {
+      const updatedUser = await context.userService.updateUser({
+        id: context.user.id,
+        ...input,
+      });
+
+      return serializeUser(updatedUser);
+    } catch (error) {
```
stbrody, line 28: this is as clean as I could make the error handling without changing the UserService to use `neverthrow`.  Its still more string matching than I'd like, but hopefully we can clean this up later with a `neverthrow`-focused refactoring

stbrody:
pretty sure the "coverage" failure is unrelated to my changes.

apps/comps/providers/session-provider.tsx
asutula: Yea you can use `wait queryClient.cancelQueries({ queryKey: tanstackClient.user.profile.key() });`

asutula (approved):
Nice!

Issue #317, opened 2025-05-14
Explore "catalogs" feature


dtbuchholz:
Many packages/apps have the same external dependencies (e.g., `viem`). `pnpm` has a "[catalogs](https://pnpm.io/catalogs)" feature. It's sort of like the `workspace:*` feature, but a root pnpm yaml defines external dep versions and allows the repo packages to use the dep like `"<package>": "catalog:".

This could help ensure we avoid ambiguous monorepo build issues, which happen whenever there's a set of overlapping deps with different versions.

PR #1337, opened 2025-10-01
chore: Move privy verification functions into shared services package

stbrody:
Builds on https://github.com/recallnet/js-recall/pull/1331

Moving these in preparation for moving more functionality that requires them into serverless functions.

Also changes the `comps` package to use `@privyio/server-auth` instead of `@privyio/node`, for consistency with our other packages


apps/api/src/services/index.ts
```diff
@@ -87,8 +90,21 @@ class ServiceRegistry {
...
+      this._privyClient = new MockPrivyClient(
+        config.privy.appId,
+        config.privy.appSecret,
+      ) as unknown as PrivyClient;
+    } else {
+      this._privyClient = new PrivyClient(
+        config.privy.appId,
+        config.privy.appSecret,
+      );
+    }
```
asutula, line 107: This is pretty strange. I'm no vitest expert, but isn't the idea that, in your test file, you mock the import for `import { PrivyClient } from "@privy-io/server-auth";` and then this service registry is using the mocked object?

stbrody: yeah that's probably better, I was just copying what was already there

apps/comps/lib/privy-client.ts
```diff
@@ -1,6 +1,6 @@
-import { PrivyClient } from "@privy-io/node";
+import { PrivyClient } from "@privy-io/server-auth";
 
-export const privyClient = new PrivyClient({
-  appId: process.env.NEXT_PUBLIC_PRIVY_APP_ID || "",
-  appSecret: process.env.PRIVY_APP_SECRET || "",
-});
+export const privyClient = new PrivyClient(
+  process.env.NEXT_PUBLIC_PRIVY_APP_ID || "",
```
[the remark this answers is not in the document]
stbrody: hmm... I was just copying what was there.  @dtbuchholz what do you think about this comments?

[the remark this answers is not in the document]
dtbuchholz: @stbrody irrelevant / incorrect. the env var is set in our vercel build env, so it'll be included in the bundle & accessible server-side. the `NEXT_PUBLIC...` simply makes sure the var is part of that bundle, which is what we want (it's ok that it's publicly exposed).

we could potentially add some helper, if we wanted to make the fallback to empty string logic more robust:
```ts
function requireEnv(name: string): string {
  const v = process.env[name];
  if (!v) throw new Error(`Missing env var: ${name}`);
  return v;
}
```
but not strictly needed.

apps/api/package.json
```diff
@@ -114,7 +114,7 @@
     "swagger-jsdoc": "^6.2.8",
     "swagger-ui-express": "^5.0.1",
     "uuid": "^9.0.0",
-    "viem": "^2.28.1",
+    "viem": "^2.37.8",
```
dtbuchholz, line 117: note: we had some issues in the past when there's a mismatch between viem package versions. 

it doesn't appear we're having that problem, but imo, it'd be nice if we keep [all viem versions](https://github.com/search?q=repo%3Arecallnet%2Fjs-recall%20%22viem%22%3A&type=code) the same. either in this PR, or as a follow up, or we can test out [pnpm catalogs](https://github.com/recallnet/js-recall/issues/317)

apps/comps/lib/privy-client.ts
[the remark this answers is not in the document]
asutula: I'm working on a better env/config setup now that uses zod fyi.

asutula (approved):

apps/api/package.json
stbrody: yeah, that's actually exactly what I'm fixing here.  I'm making the viem package versions match

dtbuchholz: *partially. my point was that there are a couple more that have a [different version](https://github.com/search?q=repo%3Arecallnet%2Fjs-recall+%22viem%22%3A&type=code) within our monorepo.

PR #1338, opened 2025-10-01
feat: linkWallet via serverless function

stbrody:
Builds on https://github.com/recallnet/js-recall/pull/1337

### DIFF
--- a/apps/comps/rpc/router/user/link-wallet.ts
+++ b/apps/comps/rpc/router/user/link-wallet.ts
@@ -1,0 +1,71 @@
+import { ORPCError } from "@orpc/server";
+import { z } from "zod";
+
+import { verifyPrivyUserHasLinkedWallet } from "@recallnet/services/lib";
+
+import { base } from "@/rpc/context/base";
+import { authMiddleware } from "@/rpc/middleware/auth";
+import { serializeUser } from "@/rpc/router/utils/serialize-user";
+
+export const linkWallet = base
+  .use(authMiddleware)
+  .input(z.object({ walletAddress: z.string() }))
+  .handler(async ({ input, context, errors }) => {
+    try {
+      const idToken = context.privyIdToken;
+      if (!idToken) {
+        throw errors.UNAUTHORIZED();
+      }
+
+      const isLinked = await verifyPrivyUserHasLinkedWallet(
+        idToken,
+        context.privyClient,
+        input.walletAddress,
+      );
+
+      if (!isLinked) {
+        throw errors.BAD_REQUEST({ message: "Wallet not linked to user" });
+      }
+
+      // Update user with the new wallet address and verification timestamp
+      const now = new Date();
+      const updatedUser = await context.userService.updateUser({
+        id: context.user.id,
+        walletAddress: input.walletAddress,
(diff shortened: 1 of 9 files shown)


apps/comps/rpc/router/user/link-wallet.ts
```diff
@@ -0,0 +1,73 @@
...
+import { authMiddleware } from "@/rpc/middleware/auth";
+import { serializeUser } from "@/rpc/router/utils/serialize-user";
+
+export const linkWallet = base
+  .use(authMiddleware)
+  .input(z.object({ walletAddress: z.string() }))
+  .handler(async ({ input, context, errors }) => {
+    try {
+      // Get Privy identity token from cookies
+      const idToken = context.cookies.get("privy-id-token")?.value;
```
asutula: This is totally fine, but also could be nice to add the privy id token to the context at in the privy user middleware. Either way is fine, though.

apps/comps/rpc/router/user/link-wallet.ts
```diff
@@ -0,0 +1,73 @@
...
+        walletAddress: input.walletAddress,
+        walletLastVerifiedAt: now,
+        updatedAt: now,
+      });
+
+      // Grant initial boost if in stake mode
+      const isStakeMode = !process.env.NEXT_PUBLIC_NON_STAKE_BOOST_AMOUNT;
+      if (isStakeMode) {
+        await context.boostAwardService.initForStake(updatedUser.walletAddress);
+      }
```
asutula, line 43: Leave this for now, but I really like the idea of just keeping the "start boosting" button on competition pages, then there is no need for this side effect.

asutula: Let's talk about that as a team and decide.

asutula (approved):
✨

apps/comps/rpc/router/user/link-wallet.ts
stbrody: done. Wasn't totally sure if it made sense to make this function depend on both authMiddleware and privyUserMiddleware or pass the token through all the middleware layers so it was accessible just on the authMiddleware.  I wound up going with making linkWallet use both middlewares, lmk if you think it should work the other way

apps/comps/rpc/router/user/link-wallet.ts
```diff
@@ -5,20 +5,20 @@ import { verifyPrivyUserHasLinkedWallet } from "@recallnet/services/lib";
 
 import { base } from "@/rpc/context/base";
 import { authMiddleware } from "@/rpc/middleware/auth";
+import { privyUserMiddleware } from "@/rpc/middleware/privy-user";
 import { serializeUser } from "@/rpc/router/utils/serialize-user";
 
 export const linkWallet = base
   .use(authMiddleware)
+  .use(privyUserMiddleware)
```
asutula: You can delete this because it's already in the chain of middleware used by authMiddleware.

stbrody: huh, it wasn't compiling for me like that before, but now it is 🤷 
Anyway done

asutula (approved):

PR #1339, opened 2025-10-02
feat: migrate website to monorepo

derrekcoleman:
- Copy all `recall-website` files to apps/website/
- Create monorepo-compatible package.json with turbo scripts
- Use shared @recallnet/eslint-config and @recallnet/typescript-config
- Fix TypeScript strict null checks for monorepo standards:
  - Add undefined checks in Showcase component (d.tsx)
  - Add nullish coalescing for split() operations in fetch-rss-news.ts
- Fix security warnings: add rel="noreferrer" to external links
- Format all code with Prettier to match monorepo style
- Dev server runs on port 3002 (avoids conflicts with api:3000, comps:3001)

TODO
- [ ] [Fix this security vulnerability](https://github.com/recallnet/js-recall/security/code-scanning/54)
- [ ] Deploy Vercel preview

### DIFF
--- a/apps/website/package.json
+++ b/apps/website/package.json
@@ -23,6 +23,7 @@
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
     "rss-parser": "^3.13.0",
+    "sanitize-html": "^2.17.0",
     "tailwind-merge": "^2.6.0"
   },
   "devDependencies": {
@@ -31,6 +32,7 @@
     "@types/node": "^20",
     "@types/react": "^19",
     "@types/react-dom": "^19",
+    "@types/sanitize-html": "^2.16.0",
     "eslint": "^9",
     "eslint-config-next": "15.1.6",
     "postcss": "^8",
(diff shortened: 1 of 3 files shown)


dtbuchholz:
passerby comment: why are we migrating it to the monorepo, and not keeping it separate from our app codebase?

derrekcoleman:
^ @carsonfarmer?

dtbuchholz:
yeh I won't die on the hill. seems fine. or we just publish packages to npm so the site can use it. 

I just sorta like dev-only stuff being built vs (like before) where the build / local dev process starts to get noisy.

derrekcoleman:
agreed that js local dev is already a bit onerous w/ things like pre-push hooks taking a while<br><br>the more intertwined the static site and the app become, the more my instinct is we want to be checking for build failures locally and in CI. <br><br>if, e.g, landing the perps category in leaderboards triggers a failure in static site build because the shape of the data changed - I think we'd want to catch that and update static site accordingly

dtbuchholz:
@derrekcoleman that can be accomplished without moving it into the monorepo, though; it can just call APIs or use published packages (whenever we get to that point). 

let's keep it separate and as its own repo—we dont want to mess with our dev setup.

derrekcoleman:
Great points. Closing this and will reopen as a new, separate repo.

PR #1342, opened 2025-10-02
feat: global agent ranks by competition type

dtbuchholz:
## Summary

Segments global agent scores by `trading` and `perpetual_futures`. Closes [APP-532](https://linear.app/recall-labs/issue/APP-532/create-two-leaderboard-types-for-trading-and-perpetual-futures)

## Details

- Allows `agent_score` to have _more than_ 1 row per agent, but it's unique per the agent ID and type. E.g., if an agent participates in paper trading and perps, there will be two distinct rows with different ranks.
- Includes the `type` (`trading` or `perpetual_futures`) in both the `agent_score` and `agent_score_history` tables to properly track this new field.
- Tests to account for "segregating" the global ranks. This mainly impacts the `/leaderboard` endpoint. There's 1 scenario where we technically leverage the global ranks for a `pending` competition's ordering; a test covers this, too.
- UI integration for the new perps type. This also fixes a bug with `trading` leaderboard, which was reading the average/top score from hardcoded JSON data and not the actual query.
Screenshot 2025-10-01 at 11 31 24 PM
Screenshot 2025-10-01 at 11 31 41 PM

- note: the "Evaluation Process" and whatnot probably need tweaks, but we can handle that later if needed


packages/db/src/schema/ranking/defs.ts
```diff
@@ -31,7 +32,8 @@ export const agentScore = pgTable(
   (table) => [
     index("idx_agent_score_agent_id").on(table.agentId),
     index("idx_agent_score_ordinal").on(table.ordinal),
-    unique("unique_agent_score_agent_id").on(table.agentId),
+    index("idx_agent_score_type").on(table.type),
```
dtbuchholz, line 35: callout re: the `agent_score` and its new constraints

packages/services/src/agentrank.service.ts
```diff
@@ -39,19 +39,36 @@ export class AgentRankService {
     );
 
     try {
+      // Get competition type using transaction if available
+      const type = await this.competitionRepo.getCompetitionType(
```
dtbuchholz, line 43: before updating ranks, we fetch the comp type (used in subsequent queries below)

packages/services/src/competition.service.ts
```diff
@@ -1322,18 +1322,21 @@ export class CompetitionService {
...
   private async calculatePendingCompetitionLeaderboard(
     competitionId: string,
+    type: CompetitionType,
   ): Promise<LeaderboardEntry[]> {
-    const agents =
+    const agentIds =
       await this.competitionRepo.getCompetitionAgents(competitionId);
-    const globalLeaderboard =
-      await this.agentScoreRepo.getAllAgentRanks(agents);
+    const globalLeaderboard = await this.agentScoreRepo.getAllAgentRanks({
```
dtbuchholz, line 1329: example of how the global rank is, technically, used when we calculate the pending comps order. this'll make it relative to the comp. 

e.g., if agent ABC is #1 in trading, but never participated in perps, they'll be treated like the rest of the "never participated" agents. we could get fancy here, but this is easier.

packages/services/src/leaderboard.service.ts
```diff
@@ -71,14 +71,12 @@ export class LeaderboardService {
    * @param params Pagination parameters
    * @returns Object with paginated agent metrics and total count
    */
-  private async getGlobalAgentMetrics(params: {
-    limit: number;
-    offset: number;
-  }): Promise<{
+  private async getGlobalAgentMetrics(params: LeaderboardParams): Promise<{
```
dtbuchholz, line 74: the method that calls `getGlobalAgentMetrics` is directly passing `LeaderboardParams`, so we can just reuse it

packages/db/src/schema/ranking/defs.ts
```diff
@@ -18,6 +18,7 @@ export const agentScore = pgTable(
   {
     id: uuid().primaryKey().notNull(),
     agentId: uuid("agent_id").notNull(),
+    type: competitionType("type").default("trading").notNull(),
```
dtbuchholz, line 21: note: i defaulted to `trading` since that'll ensure we don't have to do any special db migration logic.

apps/api/e2e/tests/leaderboard.test.ts
```diff
@@ -366,4 +370,291 @@ describe("Leaderboard API", () => {
     expect(agents[0]?.voteCount).toBe(3);
     expect(agents[1]?.voteCount).toBe(1);
   });
+
```
dtbuchholz, line 373: these tests make sure that there's no spillage b/w rankings. we want to make sure that `trading` and `perpetual_futures` are kept separate and never interact with one another.

apps/comps/components/unified-leaderboard/hub/skill-overview-card.tsx
```diff
@@ -104,7 +104,7 @@ export const SkillOverviewCard: React.FC<SkillOverviewCardProps> = ({
               <span className="text-sm text-gray-300">
                 Top:{" "}
                 {typeof stats.topScore === "number"
-                  ? stats.topScore.toFixed(1)
+                  ? stats.topScore.toFixed(0)
```
dtbuchholz, line 108: note: the `toFixed(1)` was annoying me. below is what it looks like in prod, so i got rid of the decimal (this is pretty standard)

Image

apps/comps/hooks/useUnifiedLeaderboard.ts
```diff
@@ -40,6 +49,36 @@ export const useUnifiedLeaderboard = () => {
...
+              totalParticipants: agentsResponse.pagination?.total || 0, // Use total from API
+              modelCount: 0,
+              agentCount: agentsResponse.pagination?.total || 0, // Use total from API
+              avgScore,
+              topScore,
+            },
+            // Store pagination info for potential use
+            pagination: agentsResponse.pagination,
+          };
+        } else if (skillId === "perpetual_futures") {
```
dtbuchholz, line 91: here's where we include the perps in the global leaderboard. the helpers fix the bug where avg and top scores didnt reflect the queried data.

apps/comps/hooks/useUnifiedLeaderboard.ts
```diff
@@ -40,6 +58,38 @@ export const useUnifiedLeaderboard = () => {
             offset: 0,
           });
 
+          const { avgScore, topScore } = calculateSummaryScores(
```
brunocalza, line 71: Looks like the current prod top score is broken, and this function seems to fix it. 
Screenshot from 2025-10-02 11-10-02

I want to point out that if we eventually paginate the results, this function will not work because it's calculating the avg and max from the response, and with a paginated response you would get the wrong avg and max.

dtbuchholz: I'm guessing we'll keep the current leaderboard limit to only show the top 100 agents for the foreseeable future, but forsure, makes sense. 

I'll add a TODO comment to keep this in mind.

PR #1345, opened 2025-10-02
Zod-based config

asutula:
When wiring up all the services and repositories as part of an upcoming PR, dealing with env vars and config objects to pass into services was getting unwieldy. I've also always wanted to have typesafe env vars to use, so went ahead an build something. It covers all the configuration I currently need for wiring up services, and I added a couple more comps app specific things to it just to see how it works. If we like this solution, we can add more to it.

### DIFF
--- a/apps/api/src/config/index.ts
+++ b/apps/api/src/config/index.ts
@@ -6,10 +6,7 @@ import {
   parseEvmChains,
   specificChainTokens,
 } from "@recallnet/services/lib";
-import {
-  CrossChainTradingType,
-  SpecificChain,
-} from "@recallnet/services/types";
+import { CrossChainTradingType } from "@recallnet/services/types";
 
 import { createSentryConfig } from "@/lib/sentry-config.js";
 


apps/comps/config/private.ts
asutula, line 1: The idea is private config is this is what you would use on the server. It may contain secrets and things that are not exposed to the client through NextJS env var processing. It extends the public config.

apps/comps/config/private.ts
```diff
@@ -0,0 +1,105 @@
+import z from "zod/v4";
+
+import { SpecificChainSchema } from "@recallnet/services/types";
+
+import {
+  config as publicConfig,
+  configSchema as publicConfigSchema,
+} from "./public";
+
+const configSchema = publicConfigSchema
```
asutula, line 15: This is a zod schema that defines the config object. It provides default values where appropriate. I based these default (or lack of defaults) on the config in the api app.

asutula: Since this extends the public schema, you can access any public schema properties when using the private config.

apps/comps/config/private.ts
```diff
@@ -0,0 +1,105 @@
...
+    tradingConstraints: z.object({
+      defaultMinimum24hVolumeUsd: z.coerce.number().default(100000),
+      defaultMinimumFdvUsd: z.coerce.number().default(1000000),
+      defaultMinimumLiquidityUsd: z.coerce.number().default(100000),
+      defaultMinimumPairAgeHours: z.coerce.number().default(168),
+    }),
+    maxTradePercentage: z.coerce.number().min(1).max(100).default(25),
+    rateLimiting: z.object({
+      maxRequests: z.coerce.number().default(100),
+      windowMs: z.coerce.number().default(60000),
```
asutula, line 71: Nice to be able to coerce env var values into the type we want.

apps/comps/config/private.ts
```diff
@@ -0,0 +1,105 @@
...
+    }),
+    tradingApi: z.object({
+      baseUrl: z.url(),
+      sandboxApiUrl: z.url().optional(),
+      sandboxAdminApiKey: z.string().optional(),
+    }),
+  })
+  .strict();
+
+export const config = {
```
asutula: This is where we create the config object, mapping properties to whatever env var they would be set from.

apps/comps/config/private.ts
```diff
@@ -0,0 +1,105 @@
...
+    windowMs: process.env.RATE_LIMITING_WINDOW_MS,
+  },
+  tradingApi: {
+    baseUrl: process.env.NEXT_PUBLIC_API_BASE_URL,
+    sandboxApiUrl: process.env.NEXT_PUBLIC_SANDBOX_API_URL,
+    sandboxAdminApiKey: process.env.SANDBOX_ADMIN_API_KEY,
+  },
+};
+
+export const parsedConfig = configSchema.parse(config);
```
asutula: We parse our config and export the result. This will crash the app if the parse fails. The caller will use this parsed config object.

apps/comps/config/public.ts
```diff
@@ -0,0 +1,25 @@
...
+    noStakeBoostAmount: z.coerce.bigint().optional(),
+  }),
+  clientFlags: z.object({
+    enableSandbox: z.coerce.boolean().default(false),
+    disableLeaderboard: z.coerce.boolean().default(false),
+  }),
+});
+
+export const config = {
+  frontendUrl: process.env.NEXT_PUBLIC_FRONTEND_URL,
```
asutula, line 15: The public config will always contain env vars prefixed with `NEXT_PUBLIC_`. These are available on the client side.

apps/comps/config/private.ts
```diff
@@ -0,0 +1,105 @@
...
+      .partialRecord(
+        SpecificChainSchema,
+        z.record(z.string().min(1), z.string().min(1)),
+      )
+      .default({}),
+    specificChainBalances: z
+      .partialRecord(
+        SpecificChainSchema,
+        z.record(z.string().min(1), z.number()),
+      )
```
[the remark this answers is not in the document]
dtbuchholz: (not sure if this is true, but worth checking?)

dtbuchholz (approved):
i'm not sure where the CI zod error is coming from, but aside from that, lgtm.

nit: for `parsedConfig`, is there any reason we don't just call it `config` (shorter word).

dtbuchholz:
what's the var? just curious e.g. if it'll impact local dev

apps/comps/config/private.ts
stbrody: I think this is missing actually setting up `specificChainTokens` and `specificChainBalances`

asutula: Yea I left them default (empty) for now since we're not doing trading apis yet. We can set them for sure.

apps/comps/config/private.ts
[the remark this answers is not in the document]
asutula: Nonsense I say.

PR #1346, opened 2025-10-02
chore: expose AgentService to rpc context

stbrody:


apps/api/src/config/index.ts
```diff
@@ -32,129 +34,6 @@ dotenv.config({ path: path.resolve(process.cwd(), envFile) });
 // Log which environment file was loaded (helpful for debugging)
 configLogger.info(`Config loaded environment variables from: ${envFile}`);
 
-// Helper function to parse specific chain initial balance environment variables
```
stbrody, line 35: just code motion - moving these helpers to a common config-utils.ts file in the services package so it can be shared by both the api and the comps packages

PR #1350, opened 2025-10-02
RPC getCompetitions and getCompetition

asutula:


apps/api/src/controllers/competition.controller.ts
```diff
@@ -287,23 +287,15 @@ export function makeCompetitionController(services: ServiceRegistry) {
       next: NextFunction,
     ) {
       try {
-        const agentId = req.agentId;
         const userId = req.userId;
-        const isAdmin = req.isAdmin === true;
```
asutula, line 292: These were unused.

apps/comps/components/agents-table/index.tsx
```diff
@@ -49,7 +46,7 @@ import { RankBadge } from "./rank-badge";
 export interface AgentsTableProps {
   agents: AgentCompetition[];
   totalVotes?: number;
-  competition: Competition;
```
asutula, line 52: This type that is defined in the comps app was actually not always accurate. Maybe in some situations it was representative of the data returned from the api, but in the cases I was working on, it was inaccurate.

apps/comps/components/agents-table/index.tsx
```diff
@@ -49,7 +46,7 @@ import { RankBadge } from "./rank-badge";
...
+  competition: RouterOutputs["competitions"]["getById"];
```
asutula, line 49: Better to just use the type returned by the rpc router. I'm using this syntax of accessing nested types, but we could always clean that up if we want.

apps/comps/components/timeline-chart/timeline-chart.tsx
```diff
@@ -123,16 +123,16 @@ export const TimelineChart: React.FC<PortfolioChartProps> = ({
...
   useEffect(() => {
     if (!parsedData.length) return;
 
     const now = new Date();
-    const endDate = competition.endDate ? new Date(competition.endDate) : null;
-
     let targetIndex = parsedData.length - 1; // Default to most recent week
 
-    if (competition.status === CompetitionStatus.Active) {
+    if (status === "active") {
```
asutula, line 135: No need to use these client defined types when we can just use typesafe data from the server directly.

apps/comps/components/competition-state-config.tsx
```diff
@@ -15,50 +15,40 @@ type VotingStatusConfig = {
...
-    : null;
+  const {
+    status,
+    startDate,
+    endDate,
+    votingStartDate,
+    votingEndDate,
+    joinStartDate,
+    joinEndDate,
+  } = competition;
```
asutula, line 29: I did go ahead and embrace the data as it comes from the competitions service (via the rpc router). I made this choice because the typescript types used on the server and client before were mismatched and definitely causing bugs. And look how much nicer this is ;)

apps/comps/lib/open-for-boosting.ts
```diff
@@ -0,0 +1,19 @@
+import { RouterOutputs } from "@/rpc/router";
+
+export function openForBoosting(competition: {
```
asutula, line 3: One of the things that was making the typescript types between server and client before was the derivation of a `openForBoosting` property on the server side that was sometimes included in the response and sometimes not. This was making a messy situation even messier, so I just got rid of that and added this helper on the client side to easily calculate it.

apps/comps/rpc/router/competitions/get-by-id.ts
```diff
@@ -0,0 +1,28 @@
+import { z } from "zod/v4";
+
+import { base } from "@/rpc/context/base";
+import { userMiddleware } from "@/rpc/middleware/user";
+
+export const getById = base
+  .use(userMiddleware)
```
asutula, line 10: Using the user middleware here. This doesn't mean they are authenticated (there may not be a user), but the user is resolved, if possible, and it is undefined otherwise. This is because the service function to get a competition by id also accepts an optional user id, and will "enrich" the competition with user-specific data. I hate this, but that is the way it is.

apps/comps/rpc/router/competitions/index.ts
```diff
@@ -0,0 +1,10 @@
+import { GetByIdType, getById } from "./get-by-id";
+import { ListEnrichedType, listEnriched } from "./list-enriched";
+
+export const router: {
+  listEnriched: ListEnrichedType;
+  getById: GetByIdType;
```
asutula, line 6: Typescript compiler was complaining about non portable code because of some nested usage of zod types. This was the easy fix, although this is usually not necessary.

apps/comps/rpc/router/competitions/list-enriched.ts
```diff
@@ -0,0 +1,35 @@
...
+import {
+  CompetitionStatusSchema,
+  PagingParamsSchema,
+} from "@recallnet/services/types";
+
+import { base } from "@/rpc/context/base";
+import { userMiddleware } from "@/rpc/middleware/user";
+
+export const listEnriched = base
+  .use(userMiddleware)
```
asutula, line 14: Same use of user middleware for "enriched" data.

apps/comps/rpc/router/index.ts
```diff
@@ -1,7 +1,13 @@
...
+import { router as competitions } from "./competitions";
 import { router as user } from "./user";
 
 export const router = {
   boost,
+  competitions,
   user,
 } as const;
+
+export type RouterOutputs = InferRouterOutputs<typeof router>;
```
asutula, line 13: Here is that inferred type that is very useful on the client side if you need a rpc procedure return type. There is also a `InferRouterInputs` if we ever need it.

packages/services/src/types/index.ts
```diff
@@ -28,13 +28,13 @@ export class ApiError extends Error {
...
+export type SpecificChainBalances = {
+  [K in SpecificChain]?: Record<string, number>;
+};
 
-export type SpecificChainTokens = Partial<
-  Record<SpecificChain, Record<string, string>>
->;
+export type SpecificChainTokens = {
+  [K in SpecificChain]?: Record<string, string>;
+};
```
asutula: Had to adjust these to get the behavior I believe we want, which is "zero or more keys of SpecificChain, each referencing a defined Record<string, string>"

packages/services/src/competition.service.ts
```diff
@@ -166,7 +166,9 @@ type CompetitionRulesData = {
 /**
  * Enriched competition data structure
  */
-type EnrichedCompetition = SelectCompetition & {
```
asutula, line 169: This was just wrong.

packages/services/src/competition.service.ts
```diff
@@ -231,7 +233,6 @@ type CompetitionDetailsData = {
       agentId: string | null;
     }>;
     votingEnabled?: boolean;
-    openForBoosting: boolean;
```
asutula, line 234: That simplification I was talking about to relieve some of the typing craziness we have going on.

packages/services/src/competition.service.ts
```diff
@@ -2864,8 +2865,6 @@ export class CompetitionService {
     status?: CompetitionStatus;
     pagingParams: PagingParams;
     userId?: string;
-    agentId?: string;
-    isAdmin?: boolean;
```
asutula, line 2868: Unused parameters.

apps/comps/components/timeline-chart/timeline-chart.tsx
```diff
@@ -160,29 +160,15 @@ export const TimelineChart: React.FC<PortfolioChartProps> = ({
     }
 
     setDateRangeIndex(targetIndex);
-  }, [
-    parsedData,
-    competition.status,
-    competition.startDate,
-    competition.endDate,
-  ]);
+  }, [parsedData, status, startDate, endDate]);
```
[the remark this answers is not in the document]
asutula: No, you are wrong. `startDate` and `endDate` are `Date`s. I'm not sure why you can't understand that.

apps/comps/components/competition-state-config.tsx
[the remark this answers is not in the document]
asutula: No, you are wrong. All those date are `Date`s. I'm not sure why you can't understand that.

apps/comps/lib/open-for-boosting.ts
```diff
@@ -0,0 +1,19 @@
...
+  status: RouterOutputs["competitions"]["listEnriched"]["competitions"][number]["status"];
+  votingStartDate: Date | null;
+  votingEndDate: Date | null;
```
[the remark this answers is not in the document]
asutula: No, the API returns `Date`s.

apps/comps/components/competition-status-badge.tsx
```diff
@@ -1,21 +1,21 @@
...
   text: string;
   variant: "green" | "blue" | "gray";
 };
 
 export function getCompetitionStatusConfig(
-  status: CompetitionStatus,
+  status: RouterOutputs["competitions"]["getById"]["status"],
 ): StatusConfig {
-  return status === CompetitionStatus.Active
+  return status === "active"
```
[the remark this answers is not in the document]
asutula: That is fine.

packages/services/src/competition.service.ts
stbrody: 💯

packages/services/src/balance.service.ts
```diff
@@ -7,11 +7,12 @@ import { assertUnreachable } from "./lib/typescript-utils.js";
 import {
   CompetitionType,
   SpecificChain,
+  SpecificChainBalances,
   SpecificChainTokens,
 } from "./types/index.js";
 
 export interface BalanceServiceConfig {
-  specificChainBalances: Record<SpecificChain, Record<string, number>>;
+  specificChainBalances: SpecificChainBalances;
```
stbrody, line 15: I have the exact same change in my PR 😅

stbrody: oh wait, looks like you also change the definition of what this type means, which I don't do 🤔

apps/comps/utils/competition-utils.ts
```diff
@@ -1,14 +1,19 @@
...
 import { Competition, CompetitionStatus, UserCompetition } from "@/types";
 import { CompetitionType } from "@/types/competition";
 
 import { formatDate } from "./format";
 
-export const STATUS_ICONS = {
-  [CompetitionStatus.Active]: Play,
-  [CompetitionStatus.Pending]: ClockIcon,
-  [CompetitionStatus.Ended]: CheckIcon,
+export const STATUS_ICONS: Record<
```
stbrody: hmm... this seems a lot more complicated than the types that were there before?

apps/comps/utils/competition-utils.ts
```diff
@@ -1,14 +1,19 @@
...
+  RouterOutputs["competitions"]["getById"]["status"],
+  React.ComponentType<LucideProps>
+> = {
+  active: Play,
+  pending: ClockIcon,
+  ending: Play,
```
stbrody: I think the icon for `ending` should probably be the same as for `ended`?

apps/comps/rpc/router/competitions/index.ts
stbrody: my PR upgrades the version of Zod used, which caused some weird zod errors to go away for me (I think they were maybe also about non-portal types across package boundaries).  I wonder if it will help with this one

apps/comps/rpc/router/competitions/get-by-id.ts
```diff
@@ -0,0 +1,28 @@
...
+  )
+  .handler(async ({ context, input, errors }) => {
+    try {
+      const res = await context.competitionService.getCompetitionById({
+        competitionId: input.id,
+        userId: context.user?.id,
+      });
+      return res.competition;
+    } catch (error) {
+      throw errors.INTERNAL({
```
stbrody: you should be able to preserve some more specific error type information.  Check out how I do this in my PR.  At the very least I can see that `competitionService.getCompetitionById` sometimes throws a 404 error so we could at least turn that one into a `NOT_FOUND` error, and there may be other codes it can catch and return more specific errors for

apps/comps/rpc/router/competitions/list-enriched.ts
```diff
@@ -0,0 +1,35 @@
...
+      status: CompetitionStatusSchema,
+      paging: PagingParamsSchema.optional(),
+    }),
+  )
+  .handler(async ({ context, input, errors }) => {
+    try {
+      const res = await context.competitionService.getEnrichedCompetitions({
+        status: input.status,
+        userId: context.user?.id,
+        pagingParams: input.paging || PagingParamsSchema.parse({}),
```
stbrody, line 26: hmm... in the existing endpoint, the PagingParams are required, but the fields inside it are optional and have defaults: https://github.com/recallnet/js-recall/blob/5ca5940c40c376f8ddf405a43456aaffeb2e94f9/packages/services/src/types/index.ts#L696.

I'm worried about potentially losing the default `limit` of 10 here?

apps/comps/app/competitions/page.tsx
```diff
@@ -45,24 +46,27 @@ export default function CompetitionsPage() {
   }, [trackEvent]);
 
   const { data: activeCompetitions, isLoading: isLoadingActiveCompetitions } =
-    useCompetitions({
-      status: CompetitionStatus.Active,
-      sort: "startDate",
-    });
+    useQuery(
+      tanstackClient.competitions.listEnriched.queryOptions({
+        input: { status: "active", paging: { sort: "startDate" } },
```
stbrody, line 51: nice, I like adding the nested structure for the paging args

apps/comps/rpc/router/competitions/index.ts
asutula: I hope so.

apps/comps/rpc/router/competitions/list-enriched.ts
asutula: I just thought it was weird to pass an empty object `{}` when you just want to defaults. `pagingParams: input.paging || PagingParamsSchema.parse({})` results in the default values if you don't pass anything, so `limit` of `10`.

stbrody: oh right duh.  Brain is fried

apps/comps/utils/competition-utils.ts
asutula: Yes. Done.

apps/comps/utils/competition-utils.ts
asutula: The map is kinda not a great way with the union type for status. It's much nicer with a little function. If you don't like `RouterOutputs["competitions"]["getById"]["status"]`, we can always alias that to `type Statys = RouterOutputs["competitions"]["getById"]["status"]`, but I won't do that now.

apps/comps/rpc/router/competitions/get-by-id.ts
asutula: Done.

PR #1351, opened 2025-10-03
feat: Migrate several endpoints from UserController over to serverless functions

stbrody:
Adds new serverless functions for createAgent, updateAgent, getUserAgent, getUserAgents, getAgentApiKey and migrates their callers in the frontend to use the new rpc endpoints.

Adds new serverless functions for subscribeToMailingList and unsubscribeFromMailingList, but those aren't used in the frontend anywhere, so deletes their unused frontend hooks.

Adds a new serverless function for getUserCompetitions, but doesn't yet update the frontend to use it, because the type mismatches between the backend and frontend are a mess here.  Going to wait until https://github.com/recallnet/js-recall/pull/1350 goes in and then use its new approach to managing types between the frontend and backend.

Apologies this wound up being a bigger PR than I expected.  It is broken up into independent commits so take a look at the commit list - it might make it easier to review in chunks.

Fixes https://linear.app/recall-labs/issue/APP-358/createagent, https://linear.app/recall-labs/issue/APP-366/updateagent, https://linear.app/recall-labs/issue/APP-355/getuseragent, https://linear.app/recall-labs/issue/APP-365/getuseragents, https://linear.app/recall-labs/issue/APP-367/getagentapikey, https://linear.app/recall-labs/issue/APP-546/subscribetomailinglist, https://linear.app/recall-labs/issue/APP-547/unsubscribefrommailinglist


apps/api/src/controllers/admin.controller.ts
```diff
@@ -1466,20 +1466,12 @@ export function makeAdminController(services: ServiceRegistry) {
         const result =
           await services.agentService.getDecryptedApiKeyById(agentId);
 
-        if (!result.success) {
```
stbrody, line 1469: changed getDecryptedApiKeyById to throw on error instead of returning `success:false`.  I know eventually we want to move away from throwing, but this was a pretty ugly way to do that that was actually less type safe then throwing with a fixed error code.

apps/api/src/controllers/user.controller.ts
```diff
@@ -338,19 +337,6 @@ export function makeUserController(
...
-        if (!result.success) {
-          // If there was an error, use the error code and message from the service
-          throw new ApiError(
-            result.errorCode || 500,
-            result.errorMessage || "Unknown error",
-          );
-        }
-
-        // Audit log for security tracking
-        userLogger.debug(
```
stbrody, line 350: this log message was moved into the service function

apps/api/src/controllers/user.controller.ts
```diff
@@ -515,9 +486,6 @@ export function makeUserController(
           id: userId,
           isSubscribed: true,
         });
-        if (!updatedUser) {
```
stbrody, line 518: updateUser always returns the user or throws, it never returns a null or undefined return value, so this check was completely unnecessary

apps/comps/hooks/useEmailSubscription.ts
```diff
@@ -1,39 +0,0 @@
-import { useMutation, useQueryClient } from "@tanstack/react-query";
-
-import { apiClient } from "@/lib/api-client";
-
-/**
- * Hook to subscribe to the email list
- * @returns Mutation for subscribing to the email list
- */
-export const useSubscribeToMailingList = () => {
```
stbrody, line 9: these hooks are completely unused anywhere in the app frontend code

apps/api/package.json
```diff
@@ -115,7 +115,7 @@
     "swagger-ui-express": "^5.0.1",
     "uuid": "^9.0.0",
     "viem": "^2.37.8",
-    "zod": "3.25.17"
+    "zod": "3.25.76"
```
asutula, line 118: Post tge, we should upgrade to zod 4 offically. We're already using  `zod/v4` in most places.

apps/comps/config/private.ts
```diff
@@ -72,7 +72,10 @@ const configSchema = publicConfigSchema
     }),
     tradingApi: z.object({
       baseUrl: z.url().default("https://api.competitions.recall.network/api"),
-      sandboxApiUrl: z.url().optional(),
+      sandboxApiUrl: z.preprocess(
+        (val) => (val ? val : undefined), // convert empty string to undefined
+        z.url().optional(),
+      ),
```
asutula, line 72: Good idea.

apps/comps/hooks/useAgent.ts
```diff
@@ -1,26 +1,21 @@
...
-    },
-    enabled: !!id,
-  });
+export const useUserAgent = (id?: string): UseQueryResult<Agent, Error> => {
+  return useQuery(
+    tanstackClient.agent.getAgent.queryOptions({
+      input: { agentId: id! },
+      enabled: !!id,
+    }),
+  );
```
asutula, line 18: I was having those type portability errors when wrapping the tanstackclient stuff in a custom hook like this. Cool if this worked ok!

apps/comps/hooks/useAgents.ts
```diff
@@ -20,38 +26,37 @@ export const useAgents = (params: GetAgentsParams = {}) => {
...
-      return apiClient.getUserAgents(params);
-    },
-    enabled: isAuthenticated,
-    placeholderData: (prev) => prev,
-  });
+  return useQuery(
+    tanstackClient.agent.getAgents.queryOptions({
+      input: params,
+      enabled: isAuthenticated,
+      placeholderData: (prev) => prev,
```
asutula, line 43: I'm not positive what this does, but if it does what I'm guessing (use the provided data while the query is refetching), that would be cool to prevent flashing of the component.

apps/comps/hooks/useCreateAgent.ts
```diff
@@ -13,21 +12,20 @@ export const useCreateAgent = () => {
...
-      // Invalidate relevant queries
-      queryClient.invalidateQueries({ queryKey: ["profile"] });
-      queryClient.invalidateQueries({ queryKey: ["agents"] });
-      queryClient.invalidateQueries({ queryKey: ["leaderboard"] });
-    },
-  });
+        // Invalidate relevant queries
+        queryClient.invalidateQueries({ queryKey: ["profile"] });
+        queryClient.invalidateQueries({ queryKey: ["agents"] });
+        queryClient.invalidateQueries({ queryKey: ["leaderboard"] });
```
asutula: I think we really have to use the provided query key helper from the tanstackclient just for consistency.

apps/comps/rpc/middleware/user-agent.ts
```diff
@@ -0,0 +1,34 @@
...
+import { base } from "@/rpc/context/base";
+
+import { authMiddleware } from "./auth";
+
+/**
+ * Middleware that verifies a user owns a specific agent
+ * Requires authenticated user in context
+ * Takes agentId as input and adds the verified agent to context
+ */
+export const userAgentMiddleware = base
```
asutula, line 12: Nice! Great use of middleware and keeps the logic in the handler more simple.

apps/comps/rpc/router/agent/create-agent.ts
```diff
@@ -0,0 +1,63 @@
...
+      };
+
+      return {
+        agent: serializeAgentPublic(sanitizedAgent),
+      };
+    } catch (error) {
+      // Re-throw if already an oRPC error
+      if (error instanceof ORPCError) {
+        throw error;
+      }
```
asutula, line 39: I don't understand the need for this. I can see in the above try block there is no ORPCError thrown. Is this just future proofing?

apps/comps/rpc/router/agent/get-agent-api-key.ts
```diff
@@ -0,0 +1,55 @@
...
+import { z } from "zod";
+
+import { ApiError } from "@recallnet/services/types";
+
+import { base } from "@/rpc/context/base";
+import { userAgentMiddleware } from "@/rpc/middleware/user-agent";
+
+export const getAgentApiKey = base
+  .input(z.object({ agentId: z.string().uuid() }))
+  .use(userAgentMiddleware, (input) => ({ agentId: input.agentId }))
```
asutula, line 11: Ah cool I hadn't done this before. This is how you supply input to the middleware. Makes sense.

apps/comps/rpc/router/agent/get-agent-api-key.ts
```diff
@@ -0,0 +1,55 @@
...
+        }
+      }
+
+      // Handle generic Error instances
+      if (error instanceof Error) {
+        throw errors.INTERNAL({ message: error.message });
+      }
+
+      // Unknown error type
+      throw errors.INTERNAL({ message: "Failed to get agent API key" });
```
asutula, line 53: Do you feel how you want to move this to a reusable function but it's werid/hard because the `errors` object is passed into this function and is specific to this function? We gotta figure that out.

apps/comps/types/agent.ts
```diff
@@ -148,8 +153,7 @@ export interface CreateAgentRequest {
 }
 
 export interface CreateAgentResponse {
-  agent: Agent & { apiKey: string };
-  success: boolean;
+  agent: Omit<Agent, "stats" | "skills" | "trophies" | "hasUnclaimedRewards">;
```
asutula, line 156: We'll have to move to just using the types returned by the router asap. It sucks to maintain these types and not mess something up.

packages/services/src/index.ts
```diff
@@ -19,4 +19,5 @@ export * from "./simulated-trade-execution.service.js";
 export * from "./trade-simulator.service.js";
 export * from "./trading-constraints.service.js";
 export * from "./user.service.js";
+export * from "./utils/pagination.js";
```
asutula: I kept this top level index.ts just for exporting services. There are other entrypoints in subdirectories, including `lib`. Just add this there then callers can access it as `import { buildPaginationResponse } from "@recallnet/services/lib`, for example.

asutula (commented):
Really good stuff. I left one or two comments for things that really need to be updated, so address them, then all good.

apps/comps/hooks/useCreateAgent.ts
asutula: I think this must be fixed to approve.

packages/services/src/index.ts
asutula: And also this needs to change before approval.

apps/comps/hooks/useAgents.ts
stbrody: yeah not sure, this is just how the AI translated what was already there

apps/comps/hooks/useCreateAgent.ts
stbrody: yeah so this is the part I wasn't sure about.  Can we update the keys for functions that haven't yet been migrated to the new rpc?  Like, the `leaderboard` query definitely hasn't been switched over yet.  I figured we'd go back and do a pass to update these once all the endpoints were migrated.

But looking again I guess the `profile` and `agents` ones at least could be updated, and I can leave leaderboard for now.  I just need to figure out if "profile" means user profile or agent profile...

apps/comps/rpc/router/agent/create-agent.ts
stbrody: pretty sure there was at least one of these that did throw an ORPCError internally which is how the first one got this.  After that it probably just became a pattern that got copied.  Probably not really necessary for this function, but I do kind of think its nice future-proofing.

apps/comps/rpc/router/agent/get-agent-api-key.ts
stbrody: yeah the AI could NOT figure this out on its own, but I was able to read the docs and figure it out

apps/comps/rpc/router/agent/get-agent-api-key.ts
stbrody: yeah 💯

apps/comps/types/agent.ts
stbrody: yeah, I do like how things looked in your PR for the most part.  I agree that moving towards the router functions being the source of truth for the return types is going to be the way to go

packages/services/src/index.ts
stbrody: done.  Just moved that file and also error-to-message.ts into the `lib` dir and got rid of the `utils` dir entirely

apps/comps/hooks/useCreateAgent.ts
stbrody: Okay updated, even caught a few previous rpc that had been migrated before my PR but still needed to be updated for invalidating their queries

apps/comps/hooks/useBoost.ts
```diff
@@ -60,13 +61,19 @@ export const useBoostAgent = (
       options?.onSuccess?.(data, variables, context);
       // Invalidate relevant queries
       queryClient.invalidateQueries({
-        queryKey: ["boostTotals", variables.competitionId],
+        queryKey: tanstackClient.boost.agentBoostTotals.key({
```
stbrody, line 64: these endpoints aren't updated in my PR, but they were updated previously so fixing this invalidation logic while we're here

apps/comps/hooks/useCreateAgent.ts
```diff
@@ -22,8 +22,9 @@ export const useCreateAgent = () => {
         });
 
         // Invalidate relevant queries
-        queryClient.invalidateQueries({ queryKey: ["profile"] });
```
stbrody: As far as I can tell this was just a mistake.  There is no endpoint with "profile" and the queryKey, and invaliding the user profile (which uses key "user") doesn't really make sense here since the getUserProfile rpc doesn't include any agent information

asutula (approved):
Looks great!

PR #1353, opened 2025-10-03
feat: add navbar hover states

stephhuynh18:
https://linear.app/recall-labs/issue/APP-469/missing-hover-states

Note: I am not good with front-end design stuff!!! I tried to follow the design as close as possible (with the help of ai) 

  - Added hover states to navbar navigation items with color-specific glow effects (yellow for COMPETITIONS, green for LEADERBOARDS)
  - Added hover background (`gray-900/50`) to logo, boost display, and profile button
  - Conditional hover effects on profile button (disabled when on `/profile` page)
  - Conditional left border on profile section (only shows when boost is visible)
  - Refactored nav items to use data-driven color configuration

### DIFF
--- a/apps/comps/components/navbar/index.tsx
+++ b/apps/comps/components/navbar/index.tsx
@@ -23,8 +23,18 @@ const formattedNumber = new Intl.NumberFormat();
 export const Navbar: React.FunctionComponent = () => {
   const pathname = usePathname();
   const navItems = [
-    { label: "COMPETITIONS", href: "/competitions" },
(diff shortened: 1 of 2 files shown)


dtbuchholz (commented):
this is a regression from the previous PR's code. now, it has a weird hover highlight effect on the button's full border instead of the text
Screenshot 2025-10-06 at 2 56 32 PM

Screenshot 2025-10-06 at 2 56 23 PM

PR #1355, opened 2025-10-03
feat: Update frontend to use new serverless rpc function for getUserCompetitions

stbrody:


apps/comps/components/competitions-table.tsx
```diff
@@ -132,9 +132,11 @@ export const CompetitionsTable: React.FC<CompetitionsTableProps> = ({
         id: "rank",
         header: () => "Rank",
         cell: ({ row }) => {
-          const agent = row.original.agents[0];
-          return agent ? (
-            <RankBadge rank={agent.rank} />
+          const agent = row.original.agents?.[0];
```
stbrody: this is trying to handle what if the rank is missing, which can happen in some edge cases (like a new agent that hasn't competed yet).  But I'm very much not a frontend eng so please double-check that this does something reasonable in this case

apps/comps/app/competitions/page.tsx
```diff
@@ -69,7 +69,17 @@ export default function CompetitionsPage() {
     );
 
   const { data: userCompetitions, isLoading: isLoadingUserCompetitions } =
-    useUserCompetitions();
+    useQuery(
+      tanstackClient.user.getCompetitions.queryOptions({
+        input: {
+          limit: 10,
+          offset: 0,
+          sort: "",
```
asutula: You can leave out properties that are just default values.

apps/comps/components/competitions-table.tsx
asutula: I think this more simple change should work:
```
const agent = row.original.agents[0];
          return agent?.rank ? (
            <RankBadge rank={agent.rank} />
          ) : (
            <span className="text-xs text-slate-400">-</span>
          );
```

PR #1356, opened 2025-10-03
feat: getGlobalLeaderboard via serverless rpc function

stbrody:


apps/comps/components/unified-leaderboard/skill-detail/skill-detail-page.tsx
```diff
@@ -52,7 +50,7 @@ export const SkillDetailPage: React.FC<SkillDetailPageProps> = ({
   const { refetch: loadMoreAgents, isLoading: isLoadingMore } = useQuery({
     queryKey: ["load-more-agents", currentOffset],
     queryFn: async () => {
-      const response = await apiClient.getGlobalLeaderboard({
+      const response = await client.leaderboard.getGlobal({
```
stbrody: the AI told me it couldn't use the tanstack client for this because its part of a more complicated query logic than the other cases we've migrated so far.  Please let me know if you disagree and think we should still be using the tanstackClient here (and how)

apps/comps/hooks/useUnifiedLeaderboard.ts
```diff
@@ -62,7 +60,7 @@ export const useUnifiedLeaderboard = () => {
           // If we ever paginate these results, we'll need the API itself to return the avg &
           // max score. These are probably useful, too.
           // See: https://linear.app/recall-labs/issue/APP-553/add-new-stats-to-global-leaderboard-endpoint-average-and-max-score
-          const agentsResponse = await apiClient.getGlobalLeaderboard({
+          const agentsResponse = await client.leaderboard.getGlobal({
```
stbrody, line 63: ditto about not being able to use tanstack client

apps/comps/rpc/router/leaderboard/get-global.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Get global leaderboard across all relevant competitions
+ *
+ * Note: LeaderboardService.getGlobalLeaderboard never throws errors.
+ * It returns an empty response as fallback on any failure.
+ */
+export const getGlobal = base
+  .input(LeaderboardParamsSchema)
+  .handler(async ({ input, context }) => {
+    return await context.leaderboardService.getGlobalLeaderboard(input);
+  });
```
asutula, line 32: No error handling like you've previously done?

apps/comps/components/unified-leaderboard/skill-detail/skill-detail-page.tsx
asutula: That is surprising here because the getGlobal function is maybe the simplest RPC we have. No args, just call a function and get data back. I wonder if you would have more luck if you used the tanstackClient directly instead of wrapping it in that custom hook?

apps/comps/hooks/useUnifiedLeaderboard.ts
asutula: That makes sense in this context that you'd want to just use the vanilla async client. This query function is doing a bunch of async work and data processing to build up the final response. This really shouldn't happen client side like this. We should create a service function to create the unified leaderboard so that we can cache the result and keep the client code simple. We can do that later, though.

asutula (commented):
Really just one thing to address re: error handing in the rpc procedure. Other comments and responses to your comments.

apps/comps/rpc/router/leaderboard/get-global.ts
stbrody: no, because `leaderboardService.getGlobalLeaderboard` explicitly swallows all errors and has an `emptyLeaderboardResponse` it returns in the case of any error

stbrody: anyway I put some basic boiler-plate error handling just for future proofing even though its not really needed atm

apps/comps/components/unified-leaderboard/skill-detail/skill-detail-page.tsx
stbrody: hmm yeah that was actually really easy to update this time, not sure what tripped it up last time.  Anyway switched this one over to the tanstackClient, left the other one below using the lower level client

apps/comps/components/unified-leaderboard/skill-detail/skill-detail-page.tsx
```diff
@@ -47,18 +47,16 @@ export const SkillDetailPage: React.FC<SkillDetailPageProps> = ({
...
         type: "trading",
         limit: 100,
         offset: currentOffset,
-      });
-      return response;
-    },
-    enabled: false, // Manual trigger
-  });
+      },
+      enabled: false, // Manual trigger
```
asutula: I'm surprised this works. I would think if the query is disabled, even calling refetch wouldn't run the query.

asutula: Let's verify this then I can approve.

stbrody: yeah no idea on that one 🤷

asutula: Doesn't look promising. AI is wild. All the AIs say it should work (at the top of the Google search results, for example), but looks like the docs say no: https://tanstack.com/query/v4/docs/framework/react/guides/disabling-queries

> When enabled is false: ... The query will ignore query client invalidateQueries and refetchQueries calls that would normally result in the query refetching.

stbrody: here's what my AI analysis said: 

After analyzing the code, I can confirm that **this will NOT work correctly as intended**. Here's why:

## The Problem

When `currentOffset` changes (line 65), the `useQuery` options change, which means TanStack Query creates a **new query instance** with a different internal query key. Since `enabled: false`, it won't automatically run, but here's the critical issue:

1. **First click**: `refetch()` executes with `offset: 100`, updates `currentOffset` to `200`
2. **Component re-renders** with `currentOffset: 200`
3. **New query instance** is created with `offset: 200` and a **new `refetch` function**
4. **The `useCallback` dependency `loadMoreAgents` changes**, causing the callback to be recreated
5. **Second click**: Uses the NEW `refetch` with `offset: 200` ✅

So it *might* work by accident because:
- Each render creates a fresh query with the current offset
- The new `refetch` reference updates the callback

However, this is **fragile and not the intended pattern**. Here's a better approach:

[6 tools called]

Perfect! The code now works correctly. Let me explain what I changed and why:

## Summary

**Your concern was correct** - the original pattern with `enabled: false` and `refetch()` was problematic. Here's what was wrong and how I fixed it:

### The Original Problem

```typescript
// ❌ Problematic pattern
const { refetch: loadMoreAgents, isLoading: isLoadingMore } = useQuery(
  tanstackClient.leaderboard.getGlobal.queryOptions({
    input: {
      type: "trading",
      limit: 100,
      offset: currentOffset,  // This changes!
    },
    enabled: false,
  }),
);
```

**Issues:**
1. Every time `currentOffset` changed, TanStack Query created a **new query instance** with a different query key
2. The `refetch` function reference changed on every render
3. This created unnecessary query instances in the cache
4. While it *might* work by accident, it's fragile and not the intended pattern

### The Fix

```typescript
// ✅ Better pattern - direct API call
const handleLoadMore = useCallback(async () => {
  setIsLoadingMore(true);
  try {
    const result = await client.leaderboard.getGlobal({
      type: "trading",
      limit: 100,
      offset: currentOffset,
    });
    // ... handle result
  } finally {
    setIsLoadingMore(false);
  }
}, [currentOffset]);
```

**Benefits:**
1. **Clear intent**: This is clearly a manual action, not a query
2. **No cache pollution**: Doesn't create multiple query instances
3. **Predictable**: The callback uses the current offset value when called
4. **Maintainable**: Easy to understand what's happening
5. **Follows existing patterns**: The `useUnifiedLeaderboard` hook uses the same `client.leaderboard.getGlobal()` direct call

This is the correct pattern for "load more" functionality where you're manually fetching additional data rather than managing query state.

stbrody: okay, pushed this change into the PR, PTAL

PR #1358, opened 2025-10-03
feat: CoinGecko price provider

dtbuchholz:
## Summary

Adds CoinGecko as a price provider. Closes [APP-445](https://linear.app/recall-labs/issue/APP-445/move-paper-trading-api-from-dexscreener-to-coingecko)

## Details

- Allows you to configure which paper trading price provider you want to use via an environment variable `PRICE_PROVIDER`. This defaults to DexScreener because it's a free, API-keyless service—which is easier for dev mode or non-prod settings.
- Keeps APIs and whatnot backward compatible without overhauling too much.
- Adds tests and refactors existing ones to use the new configurable `MultiChainProvider` setup.
- Removes some dead code for unused price providers (Noves, Jupiter, Raydium).

### DIFF
--- a/apps/comps/next-env.d.ts
+++ b/apps/comps/next-env.d.ts
@@ -1,6 +1,5 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-/// <reference path="./.next/types/routes.d.ts" />
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/app/api-reference/config/typescript for more information.


apps/api/src/config/index.ts
```diff
@@ -128,15 +132,24 @@ export const config = {
...
-      apiKey: process.env.NOVES_API_KEY || "",
-      enabled: !!process.env.NOVES_API_KEY,
-    },
     // Domain for API authentication and verification purposes
     domain:
       process.env.API_DOMAIN || "https://api.competitions.recall.network/",
   },
-
+  // Multichain price provider configuration (DexScreener or CoinGecko) for paper trading
+  priceProvider: {
```
dtbuchholz, line 140: new config for the price provider. it defaults to dexscreener (no API key needed) so that you dont need to set up a coingecko account. 

if you do want to use coingecko, you have the option between a "demo" (free) or "pro" (paid) plan. the endpoints are different, which is why we need the "mode" variable.

in prod, we'll always use the paid pro mode; hence, the fallback default logic when the mode isnt set.

apps/api/.env.sandbox
```diff
@@ -63,9 +63,6 @@ PRICE_CACHE_TTL_MS=60000
 # Trading Configuration
 MAX_TRADE_PERCENTAGE=25          # Maximum trade size as percentage of portfolio value (default: 25)
 
-# Optional: Price fetching using our primary provider
-NOVES_API_KEY=your_noves_api_key_here
```
dtbuchholz, line 67: i got rid of some dead code for price providers that we've never used: noves, jupiter, and raydium.

apps/api/.env.test
```diff
@@ -82,3 +82,7 @@ LOOPS_API_KEY=faux_loops_api_key
 LOOPS_MAILING_LIST_ID=faux_loops_mailing_list_id
 LOOPS_BASE_URL=http://127.0.0.1:4010
 
+# If you'd like to test "live" CoinGecko APIs in e2e tests, uncomment the following
```
dtbuchholz, line 85: since the demo API key is heavily rate limited, i figure we should keep our e2e tests as-is and use dexscreener. but, you can run e2e tests with coingecko locally if you set this up.

longer term: we probably need a simple "live" API set of tests that simply ensure the responses align to our expectations. if we tried to use CG in CI—since every PR triggers these tests, the 30 req/min limit with coingecko would get 429ed too easily.

packages/services/src/providers/__tests__/mocks/coingecko.ts
```diff
@@ -0,0 +1,260 @@
+import { Coingecko } from "@coingecko/coingecko-typescript";
```
dtbuchholz, line 1: some mocks for our unit tests. if we ever want to mock the response in the `apps/api` tests, we might want to think about creating a shared "tests" package for reusable helpers

packages/services/src/providers/__tests__/multi-chain.provider.test.ts
```diff
@@ -160,4 +163,141 @@ describe("MultiChainProvider", () => {
       }
     });
   });
+
+  describe("Configurable price providers", () => {
```
dtbuchholz, line 167: just a callout that with this feature, we can easily switch between price providers; all we need to do is update an env var and restart the server

packages/services/src/providers/price/coingecko.provider.ts
```diff
@@ -0,0 +1,432 @@
...
+    this.specificChainTokens = config.specificChainTokens;
+    this.logger = logger;
+    this.coingeckoLogger = logger.child({ context: "CoinGeckoSDK" });
+    const { apiKey, mode } = config;
+    const opts: ClientOptions = {
+      environment: mode,
+      maxRetries: this.MAX_RETRIES,
+      timeout: this.MAX_TIMEOUT,
+      logger: this.coingeckoLogger,
+    };
```
dtbuchholz, line 146: we use the coingecko SDK's native retry and timeout logic. 

however, this doesnt account for 429s. you'll see below that i wrap method calls in `withRetry` to account for this. alternatively, we could disable the coingecko SDK's logic by setting `maxRetries` and `timeout`, and rely purely on our helper method.

packages/services/src/providers/price/coingecko.provider.ts
```diff
@@ -0,0 +1,432 @@
...
+    );
+  }
+
+  /**
+   * Converts a Solana address to its canonical base58 encoded representation
+   * @param addr - The Solana address to convert
+   * @returns The canonical base58 representation of the address
+   * @throws If the address is invalid (wrong length, alphabet, or incorrect case)
+   */
+  private toCanonicalSolanaAddress(addr: string): string {
```
dtbuchholz, line 217: one callout: solana addresses are base58 encoded. we try to ensure the string value that gets passed it properly encoded before we call the API. this is important because API calls for solana-only are case sensitive, so we have to make sure the string value is correct (e.g., not lowercased).

this is a bit different than dexscreener, which is case insensitive for solana.

both providers dont care if EVM addresses are uppercase or lowercase. the CG API response always lowercases the EVM address, so we do that, too (internally within the method calls).

packages/services/src/providers/price/coingecko.provider.ts
```diff
@@ -0,0 +1,432 @@
...
+  /**
+   * Normalizes a token address for proper API calls and comparisons
+   * @param tokenAddress - The token address to normalize
+   * @returns Normalized token address
+   * @throws If the token address is invalid
+   */
+  private normalizeAddress(tokenAddress: string): string {
+    try {
+      return tokenAddress.startsWith("0x")
+        ? checksumAddress(tokenAddress as `0x${string}`)
```
dtbuchholz: but, i figured that if we're formatting solana addresses, we might as well checksum EVM addresses. 

alternatively, we can just do `toLowercase` if we dont want to checksum the EVM address. the coingecko API will still work.

packages/services/src/providers/price/coingecko.provider.ts
```diff
@@ -0,0 +1,432 @@
...
+
+  /**
+   * Get the created at timestamp from the token's oldest pool:
+   * - There are multiple pools for a token, and we want to use the oldest one since this is part
+   *   of how we enforce trading constraints.
+   * - The timestamp is natively in ISO 8601 format and must be converted to unix.
+   * @param pools - The token pools for the token
+   * @returns The created at timestamp
+   */
+  private getCreatedAtTimestampFromPools(
```
dtbuchholz, line 263: our trading constraints have a "minimum age" requirement. this is a bit nuanced. the API response will return a `data` and `included` array. the `data` has everything we need about price, FDV, volume etc.

however, it does not include any age-related information. you have to parse that from one or more token pool objects in the `included` array. the best way to do this is by looking at each pool (typically 1-3) and finding the oldest one. that effectively gives us the actual time the token created.

the response looks something like this:
```js
{
    "id": "solana_DezXAZ8z7PnrnRJjz3wXBoRgixCa6xjnB7YaB1pPB263",
    "type": "token",
    "attributes": {
      "address": "DezXAZ8z7PnrnRJjz3wXBoRgixCa6xjnB7YaB1pPB263",
      "name": "Bonk",
      ...
    },
    "relationships": {
      "top_pools": {
        "data": [
          {
            "id": "solana_5zpyutJu9ee6jFymDGoK7F6S5Kczqtc9FomP3ueKuyA9",
            "type": "pool"
          },
         ...
        ]
      }
    }
  },
  "included": [
    {
      "id": "solana_5zpyutJu9ee6jFymDGoK7F6S5Kczqtc9FomP3ueKuyA9",
      "type": "pool",
      "attributes": {
        "address": "5zpyutJu9ee6jFymDGoK7F6S5Kczqtc9FomP3ueKuyA9",
        "pool_created_at": "2023-07-05T14:41:09Z", // Use this
       ...
      },
      ...
  ]
}
```

packages/services/src/providers/price/coingecko.provider.ts
```diff
@@ -0,0 +1,432 @@
...
+   * @param data - The token data
+   * @param pools - The token pools
+   * @returns The token info
+   */
+  private parseOnchainResponse(response: TokenGetAddressResponse): TokenInfo {
+    const {
+      success,
+      error,
+      data: parsedData,
+    } = OnchainResponseSchema.safeParse(response);
```
dtbuchholz, line 302: use zod to parse the API response into the values we need

packages/services/src/providers/price/coingecko.provider.ts
```diff
@@ -0,0 +1,432 @@
...
+    try {
+      // Note: the CoinGecko SDK natively handles retries upon internal network failures. However,
+      // in case of 429 errors, `withRetry` will catch and retry manually with exponential backoff.
+      const getTokenInfo = () =>
+        this.client.onchain.networks.tokens.getAddress(tokenAddress, {
+          network,
+          include_composition: true,
+          include: "top_pools",
+        });
+      const response = await withRetry(getTokenInfo, {
```
dtbuchholz, line 388: here's that `withRetry` helper and the _somewhat_ double retry logic...only for 429s, though, since the CG SDK should first do retries on 500s etc

(i'm debating if we should only use our helper, or if this dual-setup is better)

packages/services/src/providers/price/coingecko.provider.ts
```diff
@@ -0,0 +1,432 @@
...
+  }
+
+  /**
+   * Get prices for multiple tokens in batches
+   * @param tokenAddresses - Array of token addresses to fetch prices for
+   * @param chain - Blockchain type (EVM or SVM)
+   * @param specificChain - Specific chain identifier
+   * @returns Map of token addresses to their price information (null if not found)
+   */
+  async getBatchPrices(
```
dtbuchholz, line 547: ~note: i realized there _is_ a "get batch addresses" API from CG. originally, i didnt see this.~

~i'm in the process of refactoring and including that since iiuc we do use the batch method in the `apps/api`. for now, we just do some promises and fetch on token at a time~

edit: this was refactored to use CG's batch onchain token API. the only intracacies:
- the API has a max of either 30 or 50 token addresses, depending on if you use the free or paid API.
- we do some weird address tracking (original vs normalized) to make sure the API requests dont fail with ill-formed solana addresses, and the response contains the original address that was passed to the method.

packages/services/src/providers/multi-chain.provider.ts
```diff
@@ -65,12 +93,9 @@ export class MultiChainProvider implements PriceSource {
     specificChain?: SpecificChain,
   ): Promise<PriceReport | null> {
     try {
-      // Normalize token address to lowercase
-      const normalizedAddress = tokenAddress.toLowerCase();
```
dtbuchholz, line 69: note: i removed this because it caused issues with CG. that's why we implemented the base58 logic above for solana addresses

packages/services/src/types/index.ts
```diff
@@ -34,7 +34,16 @@ export type SpecificChainBalances = {
...
 };
 
-export type SpecificChainTokens = typeof specificChainTokens;
+type TokenDict = Record<string, string>;
+type KnownChains = keyof typeof specificChainTokens;
+
+/**
+ * Standard chain tokens with optional additional tokens
+ */
+export type SpecificChainTokens = Partial<{
```
dtbuchholz, line 43: we had some existing tests that try to add new token keys or new chains. that variable is derived from this object:
```ts
declare const specificChainTokens: {
    readonly eth: {
        readonly eth: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2";
        readonly usdc: "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48";
        readonly usdt: "0xdAC17F958D2ee523a2206206994597C13D831ec7";
    };
...
```

with this change, it just makes it more flexible so that you can add chains/tokens but still access the default list

apps/api/e2e/tests/multi-agent-competition.test.ts
```diff
@@ -308,47 +296,7 @@ describe("Multi-Agent Competition", () => {
     // Wait for all trades to settle
     await wait(500);
 
-    // Step 5: Verify each agent has a unique token composition
-    for (let i = 0; i < NUM_AGENTS; i++) {
```
dtbuchholz, line 313: i refactored the setup to just make trades to the `0x00...dead` address because that eliminates an API call in CI. i dont think the BASE_TOKENS are really worth testing here.

apps/api/e2e/tests/multi-agent-competition.test.ts
```diff
@@ -366,179 +314,4 @@ describe("Multi-Agent Competition", () => {
       }
     }
   });
-
-  // Test that portfolio values change over time due to price fluctuations
-  test("portfolio values should change differently for agents holding different tokens", async () => {
```
dtbuchholz, line 372: this seemed like an unnecessary test since we do so many similar checking elsewhere...and it takes forever

apps/api/e2e/tests/trading.test.ts
```diff
@@ -554,7 +554,11 @@ describe("Trading API", () => {
     );
 
     // 1. Fetch the price for the arbitrary token
-    const priceResponse = await agentClient.getPrice(wethTokenAddress);
+    const priceResponse = await agentClient.getPrice(
+      wethTokenAddress,
+      BlockchainType.SVM,
```
dtbuchholz, line 559: as part of [APP-614](https://linear.app/recall-labs/issue/APP-614/refactor-multichainprovider-and-dexscreenerprovider-logic), we shouldnt "guess" the chain anymore.

in practice, we only call `getPrice` [here](https://github.com/recallnet/js-recall/blob/07445e8bd4179e8ac7d07af9adcbaca968891765/apps/api/src/controllers/price.controller.ts#L75), and it'll have the chain info.

packages/services/src/providers/price/coingecko.provider.ts
```diff
@@ -0,0 +1,616 @@
...
+  ): number | undefined {
+    const poolsWithTimestamps = pools.filter(
+      (p) => p.attributes?.pool_created_at,
+    );
+    if (poolsWithTimestamps.length === 0) {
+      return undefined;
+    }
+    // Note: the timestamps are guaranteed to be non-null because we filtered above
+    const sorted = poolsWithTimestamps.sort((a, b) => {
+      const aTime = new Date(a.attributes!.pool_created_at!).getTime();
```
dtbuchholz, line 274: note: i hate using `!` but opted for it here for simplicity reason

PR #1359, opened 2025-10-03
feat: display two prize pools

brunocalza:


apps/comps/components/competition-card.tsx
```diff
@@ -104,7 +105,15 @@ export const CompetitionCard: React.FC<CompetitionCardProps> = ({
               <h3 className="text-secondary-foreground mb-1 text-xs font-semibold uppercase">
                 Reward
               </h3>
-              {competition.rewards ? (
+              {competition.rewardsTge ? (
```
brunocalza, line 113: we first check if `rewardsTge` is available, if not we check `rewards`, if not we display `TBA`

apps/comps/components/Recall.tsx
```diff
@@ -0,0 +1,80 @@
+import Image from "next/image";
```
brunocalza, line 1: this is the Recal Token Icon component

packages/db/src/repositories/competition-rewards.ts
```diff
@@ -56,37 +56,6 @@ export class CompetitionRewardsRepository {
     }
   }
 
-  /**
```
brunocalza, line 59: removed some things that are not being used

packages/services/src/competition-reward.service.ts
```diff
@@ -60,22 +60,6 @@ export class CompetitionRewardService {
     return this.competitionRewardsRepo.createRewards(rewardsData, tx);
   }
 
-  /**
```
brunocalza, line 63: removed methods that are not being used

packages/services/src/competition.service.ts
```diff
@@ -562,6 +572,15 @@ export class CompetitionService {
         `[CompetitionManager] Created trading constraints for competition ${id}`,
       );
 
+      // Create prize pools if provided
+      if (prizePools) {
+        const attoPrizePools = {
+          agent: valueToAttoBigInt(prizePools.agent),
```
brunocalza, line 542: we are storing values in atto units

apps/api/e2e/tests/admin.test.ts
```diff
@@ -2358,4 +2359,173 @@ describe("Admin API", () => {
...
+
+  // ===== Prize Pool Tests =====
+
+  test("should create competition with prizePools.agent and prizePools.users", async () => {
+    const client = createTestClient(getBaseUrl());
+    await client.loginAsAdmin(adminApiKey);
+
+    // Create a competition with prize pools
+    const createResponse = await axios.post(
+      `${getBaseUrl()}/api/admin/competition/create`,
```
dtbuchholz: note: we have a test `ApiClient` that we should update to take these new params into account: [`createCompetition`](https://github.com/recallnet/js-recall/blob/f630a5eeb242f3fb9766c691978a962a1095980a/apps/api/e2e/utils/api-client.ts#L513) & [`updateCompetition`](https://github.com/recallnet/js-recall/blob/f630a5eeb242f3fb9766c691978a962a1095980a/apps/api/e2e/utils/api-client.ts#L589).

and then we can do:
```ts
const adminClient = createTestClient();
const createCompetitionResult = await adminClient.createCompetition({
  ...
});
```

packages/db/src/repositories/competition.ts
```diff
@@ -2638,4 +2655,47 @@ export class CompetitionRepository {
...
+        .onConflictDoUpdate({
+          target: competitionPrizePools.competitionId,
+          set: {
+            agentPool: pools.agent,
+            userPool: pools.users,
+          },
+        })
+        .returning();
+
+      return result!;
```
dtbuchholz: we remove the bang operator and check if the result exists

apps/comps/components/modals/confirm-agent-entry.tsx
```diff
@@ -103,7 +105,28 @@ export const ConfirmAgentEntryModal: React.FC<ChooseAgentModalProps> = ({
...
+                          compact
+                        />,
+                      ]
+                    : [
+                        <Rewards
+                          key="rewards"
+                          rewards={competition?.rewards || []}
+                          compact
+                        />,
+                      ]}
```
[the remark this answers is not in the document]
dtbuchholz: ^ah, yes this is true

packages/db/src/repositories/competition.ts
```diff
@@ -51,7 +51,11 @@ import {
   InsertPortfolioSnapshot,
   SelectPortfolioSnapshot,
 } from "../schema/trading/types.js";
-import { Database, Transaction } from "../types.js";
+import {
+  Database,
+  Transaction as DatabaseTransaction,
+  Transaction,
```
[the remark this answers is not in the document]
dtbuchholz: seems valid

PR #1361, opened 2025-10-03
RPC for getting competition agents + ui integration

asutula:


apps/api/src/controllers/competition.controller.ts
```diff
@@ -384,14 +384,7 @@ export function makeCompetitionController(services: ServiceRegistry) {
...
         });
 
-        res.status(200).json({
-          ...result,
-          pagination: buildPaginationResponse(
-            result.total,
-            queryParams.limit,
-            queryParams.offset,
-          ),
-        });
```
asutula, line 394: Moved the building of the pagination object into the service

apps/comps/components/user-vote.tsx
```diff
@@ -30,10 +30,12 @@ export const UserVote: React.FC<UserVoteProps> = ({
...
-  const { data: agentData } = useAgent(agentId);
-  const { data: competitionAgentsData } = useCompetitionAgents(competitionId, {
-    filter: agentData?.agent.name,
-  });
+  // TODO: Filter by agent name once the api supports it.
+  const { data: competitionAgentsData } = useQuery(
+    tanstackClient.competitions.getAgents.queryOptions({
+      input: { competitionId },
+    }),
+  );
```
asutula, line 38: The filter param provided by the old code wasn't actually used by the service, so deleted it. We're going to delete this file anyways since we no longer do voting.

apps/comps/hooks/useLeaveCompetition.ts
```diff
@@ -26,13 +26,17 @@ export const useLeaveCompetition = () => {
...
+        }),
       });
       queryClient.invalidateQueries({
         queryKey: ["agent-competitions", agentId],
       });
       queryClient.invalidateQueries({ queryKey: ["leaderboard"] });
-      queryClient.invalidateQueries({ queryKey: ["competitions"] });
+      queryClient.invalidateQueries({
+        queryKey: tanstackClient.competitions.listEnriched.key(),
+      });
```
asutula, line 39: Found some query invalidations for apis we've already moved to rpc.

apps/api/src/controllers/competition.controller.ts
stbrody: interesting, I've mostly been going the other direction in my migrations, since response format structure feels like more the concern of the handler than the service.  But its pretty minor, I don't feel strongly either way

apps/comps/app/competitions/[id]/page.tsx
```diff
@@ -64,12 +63,20 @@ export default function CompetitionPage({
...
+      placeholderData: (prev) => prev,
+      input: {
+        competitionId: id,
+        paging: {
+          sort: agentsSort,
+          offset: agentsOffset,
+          limit: LIMIT_AGENTS_PER_PAGE,
+        },
+      },
+    }),
```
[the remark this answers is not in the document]
Comment (stbrody) [reply apps/comps/app/competitions/[id]/page.tsx:77]: this looks potentially like a real issue?

apps/comps/components/agents-table/index.tsx
```diff
@@ -76,9 +76,9 @@ export const AgentsTable: React.FC<AgentsTableProps> = ({
   const [columnVisibility, setColumnVisibility] = useState<VisibilityState>({
     yourShare: session.ready && session.isAuthenticated,
   });
-  const [selectedAgent, setSelectedAgent] = useState<AgentCompetition | null>(
-    null,
-  );
+  const [selectedAgent, setSelectedAgent] = useState<
+    RouterOutputs["competitions"]["getAgents"]["agents"][number] | null
```
stbrody, line 80: I think at some point it might be nice to give these types name aliases so things are a bit easier to read, but not important right now.  Once everything in the frontend is using this method of inferring the types from the backend, we can delete all the existing types defined in the frontend's types file and add aliases for these RouterOutputs-derived types there instead as a nice readability cleanup

apps/comps/components/modals/choose-agent.tsx
```diff
@@ -35,7 +36,12 @@ export const ChooseAgentModal: React.FC<ChooseAgentModalProps> = ({
   onContinue,
   competition,
 }) => {
-  const { data: compAgents } = useCompetitionAgents(competition?.id);
+  const { data: compAgents } = useQuery({
+    ...tanstackClient.competitions.getAgents.queryOptions({
```
stbrody: my AI tried to do this a few times.  I'm pretty sure its wrong, or at least not the way the rest of the code does it and seems uglier.  `enabled` should be an argument to queryOptions (at the same level as "input"), not useQuery, and we don't need the spread operator.

I'm not 100% sure thats right, but it's definitely how a lot of the other calls are set up now and we should probably figure out which is right, or if they both are still probably a good idea to standardize on one approach

apps/comps/app/competitions/[id]/chart/page.tsx
```diff
@@ -37,10 +36,12 @@ export default function CompetitionChartPage({
     data: agentsData,
     isLoading: agentsLoading,
     error: agentsError,
-  } = useCompetitionAgents(id, {
-    limit: 10, // Default page size
-    offset: 0,
-  });
+  } = useQuery(
+    tanstackClient.competitions.getAgents.queryOptions({
```
Comment (stbrody) [inline apps/comps/app/competitions/[id]/chart/page.tsx:40]: `useCompetitionAgents` previously had an "enabled" check based on if the competitionId was defined or not.  Not all the places where you inlined the query carried over that "enabled" arg.  I think there are a few others throughout this PR besides just this one

apps/comps/app/competitions/[id]/page.tsx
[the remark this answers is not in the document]
Comment (stbrody) [reply apps/comps/app/competitions/[id]/page.tsx:77]: nevermind, I see your comment below that the filter wasn't used.  Does that mean the `debouncedFilterTerm` defined above is unused now and can be deleted?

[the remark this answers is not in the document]
Comment (asutula) [reply apps/comps/app/competitions/[id]/page.tsx:77]: I left it assuming we'll want to actually support it on the server side.

apps/comps/components/agents-table/index.tsx
asutula: Sounds like a plan.

apps/comps/components/modals/choose-agent.tsx
asutula: Oh weird. Yea fixing.

apps/comps/app/competitions/[id]/chart/page.tsx
Comment (asutula) [reply apps/comps/app/competitions/[id]/chart/page.tsx:40]: In some contexts, the competition id is defined `string` and in other contexts it is `string | undefined`. I think in cases where it is possibly `undefined`, I do set `enabled`, but when I know it is defined there is no reason to.

Comment (asutula) [reply apps/comps/app/competitions/[id]/chart/page.tsx:40]: In this particular case, `id` is always defined.

apps/api/src/controllers/competition.controller.ts
asutula: My logic was: The service function receives pagination parameters as an argument and queries the database accordingly, so it makes sense that the service function would return pagination parameters as well. I also don't feel strongly.

PR #1362, opened 2025-10-03
feat: Use serverless functions for public getAgent and getAgents calls

stbrody:


apps/comps/hooks/useAgent.ts
```diff
@@ -19,21 +19,16 @@ export const useUserAgent = (id?: string): UseQueryResult<Agent, Error> => {
...
-    enabled: !!id,
-  });
-};
+export const useAgent = (
+  id?: string,
+): UseQueryResult<RouterOutputs["agent"]["getAgent"], Error> =>
+  useQuery(
+    tanstackClient.agent.getAgent.queryOptions({
+      input: { agentId: id! },
+      enabled: !!id,
```
asutula: I learned the hard way this doesn't work. The query options are evaluated immediately, so `input` is calculated right away with no consideration of the `enabled` property, so `id!` will throw an exception. You just need to set `agentId` conditionally like 
```
input: { agentId: id || "" },
```

stbrody: fixed.  Also fixed 2 others that were already in

asutula (approved):

PR #1364, opened 2025-10-05
Cache middleware

asutula:


apps/comps/rpc/context/base.ts
```diff
@@ -52,6 +53,7 @@ export const base = os
     competitionService: CompetitionService;
     agentService: AgentService;
     emailService: EmailService;
+    logger: Logger;
```
asutula, line 60: Added this so we can log anywhere in the rpc tree easily.

apps/comps/rpc/middleware/cache.ts
```diff
@@ -0,0 +1,70 @@
...
+export function cacheMiddleware<
+  TInContext extends Context & { logger: Logger },
+  TInput,
+  TOutput,
+  TErrorMap extends ErrorMap,
+  TMeta extends Meta,
+>(options?: {
+  key?: string[];
+  revalidate?: number;
+  tags?: string[];
```
asutula, line 32: These args mirror the args to the NextJS `unstable_cache` function and will be passed through to it.

apps/comps/rpc/middleware/cache.ts
```diff
@@ -0,0 +1,70 @@
...
+  includeContext?: (context: TInContext) => Record<string, any>;
```
asutula, line 33: This extra arg is a callback that allows the caller to select data from the context that should be included in the cache key. This is useful when the procedure input doesn't fully encompass what the cache key should be. For example, if a procedure uses auth middleware and the user id from the context, and passes the user id along with procedure input into a service function, you'd want to select the user id from the context so it is included in the cache key.

apps/comps/rpc/middleware/cache.ts
```diff
@@ -0,0 +1,70 @@
...
+  Record<never, never>,
+  TInput,
+  TOutput,
+  ORPCErrorConstructorMap<TErrorMap>,
+  TMeta
+> {
+  return os
+    .$context<{ logger: Logger }>()
+    .middleware(async ({ context, next, path }, input) => {
+      const contextValues = options?.includeContext?.(context as TInContext);
```
asutula, line 46: Get our optional context data from the callers provided callback.

apps/comps/rpc/middleware/cache.ts
```diff
@@ -0,0 +1,70 @@
...
+
+      const serializer = new StandardRPCJsonSerializer();
+      const [serialized] = serializer.serialize({
+        input,
+        contextValues,
+      });
```
asutula, line 53: Our input and context may contain non json-compatible properties like `Date` and `bigint`, so we use oRPC's built in serializer to serialize them into json.

apps/comps/rpc/middleware/cache.ts
```diff
@@ -0,0 +1,70 @@
...
+
+      const cacheKey = [
+        ...(options?.key || []),
+        ...path,
+        JSON.stringify(serialized),
+      ];
```
asutula, line 60: Build our cache key array using items provided by the caller, the procedure path (this would be, for example, `[competitions, getAgents]`, and the json serialized input and context values.

apps/comps/rpc/middleware/cache.ts
```diff
@@ -0,0 +1,70 @@
...
+            "Cache middleware miss, invoking next()",
+          );
+          return await next();
+        },
+        cacheKey,
+        {
+          revalidate: options?.revalidate,
+          tags: options?.tags,
+        },
+      );
```
asutula, line 76: Wrap our call to `next` in `unatable_cache`, passing in our cache key and other parameters provided by the user.

apps/comps/rpc/middleware/cache.ts
```diff
@@ -0,0 +1,70 @@
...
+
+      context.logger.debug(
+        { cacheKey },
+        "Cache middleware checking cache for key",
+      );
+      return await cachedHandler();
```
asutula, line 82: Call it.

apps/comps/rpc/router/competitions/get-agents.ts
```diff
@@ -4,8 +4,14 @@ import { z } from "zod/v4";
...
 
 import { base } from "@/rpc/context/base";
+import { cacheMiddleware } from "@/rpc/middleware/cache";
 
 export const getAgents = base
+  .use(
+    cacheMiddleware({
+      revalidate: 20,
+    }),
+  )
```
asutula, line 14: Example of using the middleware when getting competition agents. This is a simple case where we just cache the result for 20 seconds.

.npmrc
```diff
@@ -1 +1,2 @@
-shamefully-hoist=false
\ No newline at end of file
+public-hoist-pattern[]=*import-in-the-middle*
+public-hoist-pattern[]=*require-in-the-middle*
```
asutula, line 2: The sentry-related errors in the console were driving me crazy so fixed it according to their docs https://docs.sentry.io/platforms/javascript/guides/node/troubleshooting/#pnpm-resolving-import-in-the-middle-external-package-errors

apps/comps/rpc/middleware/cache.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export function cacheMiddleware<
+  TInContext extends Context & { logger: Logger },
+  TInput,
+  TOutput,
+  TErrorMap extends ErrorMap,
+  TMeta extends Meta,
+>(options?: {
+  key?: string[];
+  revalidate?: number;
```
stbrody: can we call this `revalidateSecs` so its clear what the unit is?

Really I'd love for us to start using a real Duration type throughout the case base for stuff like this to really eliminate the risk of bugs related to misused units for durations

apps/comps/rpc/middleware/cache.ts
```diff
@@ -0,0 +1,70 @@
...
+  ErrorMap,
+  Meta,
+  Middleware,
+  ORPCErrorConstructorMap,
+  os,
+} from "@orpc/server";
+import { unstable_cache } from "next/cache";
+import type { Logger } from "pino";
+
+export function cacheMiddleware<
```
stbrody, line 23: this could benefit from some doc comments, its relatively complex and non necessary obvious what all the args and generics represent

apps/comps/rpc/middleware/cache.ts
stbrody: lets document this behavior please!

apps/comps/rpc/middleware/cache.ts
```diff
@@ -0,0 +1,70 @@
...
+
+      const serializer = new StandardRPCJsonSerializer();
+      const [serialized] = serializer.serialize({
+        input,
+        contextValues,
+      });
+
+      const cacheKey = [
+        ...(options?.key || []),
+        ...path,
```
stbrody, line 57: extreme nitpick and probably makes zero difference, but in my head it would makes sense for the rpc path to be the first part of the cache key, then have the explicit key and inputs after it.  Maybe could be relevant in debugging issues?

apps/comps/rpc/router/competitions/get-agents.ts
stbrody: wow this is so nice and clean, love it

stbrody (approved):
lgtm, mostly just want more comments explaining how this works in the code

PR #1365, opened 2025-10-05
feat: rm Agent serialization and make frontend use types derived from the backend function signatures

stbrody:
Builds on https://github.com/recallnet/js-recall/pull/1362


apps/comps/components/create-agent/agent-created.tsx
```diff
@@ -11,10 +11,10 @@ import { CopyButton } from "@/components/copy-button";
 import { socialLinks } from "@/data/social";
 import { useSession } from "@/hooks/useSession";
 import { useUnlockKeys } from "@/hooks/useUnlockKeys";
-import { Agent } from "@/types";
+import type { RouterOutputs } from "@/rpc/router";
```
asutula, line 14: It's never been clear to me if it is necessary to use `import type` when the think you are importing is already a type. Just thinking out loud here.

asutula (approved):
Nice, that was a clean migration, no real changes to any react components!

PR #1366, opened 2025-10-05
feat: Serverless function for getCompetitionTrades

stbrody:


packages/db/src/repositories/trade.ts
```diff
@@ -222,7 +222,7 @@ export class TradeRepository {
           },
         })
         .from(trades)
-        .leftJoin(agents, eq(trades.agentId, agents.id))
+        .innerJoin(agents, eq(trades.agentId, agents.id))
```
asutula, line 225: Nice catch.

apps/comps/components/trades-table/index.tsx
```diff
@@ -104,16 +105,26 @@ export const TradesTable: React.FC<TradesTableProps> = ({
...
+            return (
+              <div className="text-secondary-foreground flex flex-col items-end">
+                Unknown
+              </div>
+            );
+          }
+          return (
+            <div className="flex flex-col items-end">
+              <span className="text-primary-foreground text-sm">
+                {format(new Date(timestamp), "MMM d, yyyy")}
```
asutula: I believe timestamp is already a `Date`.

PR #1368, opened 2025-10-06
feat: Use serverless function for getCompetitionPerpsPositions

stbrody:


apps/comps/components/positions-table/index.tsx
```diff
@@ -86,11 +86,13 @@ export const PositionsTable: React.FC<PositionsTableProps> = ({
...
               {row.original.isLong ? "LONG" : "SHORT"}{" "}
-              {formatAmount(row.original.size)} {row.original.asset}
+              {formatAmount(Number(row.original.positionSize))}{" "}
+              {row.original.asset}
             </div>
             <div className="text-secondary-foreground text-xs">
-              {row.original.leverage}x leverage •{" "}
-              {formatAmount(row.original.collateral)} collateral
+              {row.original.leverage ? Number(row.original.leverage) : 0}x
+              leverage • {formatAmount(Number(row.original.collateralAmount))}{" "}
```
asutula, line 94: Not a problem with this pr, but it sucks to return all these properties as strings from the db. Could probably use bigint mode.  That's another conversation.

asutula (approved):
Looks good.

apps/comps/components/positions-table/index.tsx
stbrody: yeah I meant to point that out on the PR.  Its the first time since moving towards having the frontend just use the backend database types without serialization that I'm wondering if doing some intermediate serialization might make sense after all

asutula: I think Drizzle types are typically pretty ideal. Those columns are `numeric` type, and Drizzle let's you configure the `mode` that data is returned in. By default it is `string`, but you can also choose `bigint` or `number` depending on your requirements related to precision etc. Anything token amount related can be `bigint` and other types of values can probably be `number`.

PR #1369, opened 2025-10-06
fix(config): rewards config

brunocalza:
if rewards config are not set, do not try to initialize the rewards allocator

### DIFF
--- a/apps/api/src/services/index.ts
+++ b/apps/api/src/services/index.ts
@@ -53,6 +53,7 @@ import { EventProcessor } from "@/indexing/event-processor.js";
 import { EventsRepository } from "@/indexing/events.repository.js";
 import { IndexingService } from "@/indexing/indexing.service.js";
 import {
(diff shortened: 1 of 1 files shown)


apps/api/src/services/index.ts
```diff
@@ -121,6 +121,13 @@ class ServiceRegistry {
     if (config.server.nodeEnv === "test") {
       this._rewardsAllocator =
         new MockRewardsAllocator() as unknown as RewardsAllocator;
+    } else if (
+      !config.rewards.allocatorPrivateKey ||
+      !config.rewards.contractAddress ||
+      !config.rewards.rpcProvider
+    ) {
+      this._rewardsAllocator =
```
sam701, line 131: Should we add a log statement here, in case it's helpful for troubleshooting?

brunocalza: sounds good!

PR #1371, opened 2025-10-06
fix(api|services): competition feature flags & cross chain trading

dtbuchholz:
## Summary

Fixes problem where competition-specific feature flags only read from the static defaults, and do not use the dynamically updated values.Closes [APP-587](https://linear.app/recall-labs/issue/APP-587/fix-competition-feature-flag-always-using-hardcoded-values).

## Details

`SimulatedTradeExecutionService` was using a global `this.features.CROSS_CHAIN_TRADING_TYPE` flag to validate cross-chain trading instead of reading the competition-specific `crossChainTradingType` setting from the database. This caused all comps to use the default "disallow all" setting instead of the db `trading_competitions` table.


apps/api/src/services/index.ts
```diff
@@ -47,13 +47,12 @@ import {
 } from "@recallnet/services/providers";
 import RewardsAllocator from "@recallnet/staking-contracts/rewards-allocator";
 
-import config, { features } from "@/config/index.js";
+import config from "@/config/index.js";
```
dtbuchholz, line 50: this is the root of the problem. we load `features` into the constructor. features _always_ has `CROSS_CHAIN_TRADING_TYPE: "disallowAll"`, but the database stores these features within table values.

packages/services/src/__tests__/trade-simulator.constraints.test.ts
```diff
@@ -534,6 +533,119 @@ describe("SimulatedTradeExecutionService - Trading Constraints", () => {
       });
     });
 
+    describe("validateCrossChainTrading", () => {
```
dtbuchholz, line 536: this is the main test that validates the bug exists.

if you check out [`f9514fc`](https://github.com/recallnet/js-recall/pull/1371/commits/f9514fc8939c17ff3ecc5609a8f42f0be4fc3845) and run the following test files, you'll see they fail:
- `e2e/tests/trading.test.ts`
- `src/__tests__/trade-simulator.constraints.test.ts`

then, check out the final commit in this branch, build the services package, and re-run the tests. they pass.

packages/services/src/simulated-trade-execution.service.ts
```diff
@@ -170,8 +165,11 @@ export class SimulatedTradeExecutionService {
         chainOptions,
       );
 
-      // Validate cross-chain trading rules
-      this.validateCrossChainTrading(chainInfo);
+      // Validate cross-chain trading rules using competition-specific settings
+      this.validateCrossChainTrading(
+        chainInfo,
+        competition.crossChainTradingType,
```
dtbuchholz, line 171: since the comp stores the feature, we can just use the db value instead of relying on the constructor value. 

note: "features" is probably a bit of a misnomer since it's not really a feature flag, but an explicit and overridable stateless value. it should probably be a zod default in the admin "create competition" endpoint (iirc it already works this way), and that's it.

created a follow up: https://linear.app/recall-labs/issue/APP-588/remove-competition-app-feature-flags

PR #1374, opened 2025-10-06
feat: Remove authenticated rules endpoint in favor of public one

stbrody:
Before this change, there were 2 endpoints for getting the rules for a competition:
-getRules returns the rules for the currently active competition and requires you to be logged in as either an admin or an agent that is registered in the competition.
- getCompetitionRules returns the rules for any competition by competitionId

The two endpoints return identical data.  This PR removes the authenticated version, and updates the public one to make competitionId optional and default to the current competition (also updates the URL/params schema slightly)

### DIFF
--- a/apps/api/e2e/tests/competition.test.ts
+++ b/apps/api/e2e/tests/competition.test.ts
@@ -4070,7 +4070,7 @@ describe("Competition API", () => {
       expect(Array.isArray(response.data.agents)).toBe(true);
     });
 
-    test("should allow unauthenticated access to GET /competitions/rules?competitionId={id}", async () => {
+    test("should allow unauthenticated access to GET /competitions/{id}/rules", async () => {
       // Setup: Create competition with trading constraints via admin
       const adminClient = createTestClient();
       await adminClient.loginAsAdmin(adminApiKey);
@@ -4094,7 +4094,7 @@ describe("Competition API", () => {
 
       // Test: Direct axios call without authentication
       const response = await axios.get(
-        `${getBaseUrl()}/api/competitions/rules?competitionId=${competition.id}`,
+        `${getBaseUrl()}/api/competitions/${competition.id}/rules`,
       );
 
       expect(response.status).toBe(200);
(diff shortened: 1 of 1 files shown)


packages/services/src/competition.service.ts
```diff
@@ -2666,151 +2666,172 @@ export class CompetitionService {
...
-    try {
-      // Get active competition first, as rules are always for the active one
-      const activeCompetition = await this.getActiveCompetition();
-
-      if (!activeCompetition) {
-        throw new ApiError(
-          404,
-          "No active competition found to get rules for.",
-        );
+  private buildInitialBalanceDescriptions(): string[] {
```
stbrody, line 2672: this is just breaking up the large getCompetitionRules function into multiple smaller helper functions

apps/api/e2e/tests/competition.test.ts
```diff
@@ -4338,6 +4339,39 @@ describe("Competition API", () => {
...
+          minimumLiquidityUsd: 100000,
+          minimumFdvUsd: 1000000,
+        },
+      });
+
+      expect(competitionResponse.success).toBe(true);
+      const competition = competitionResponse as CreateCompetitionResponse;
+
+      // Test with specific competition ID via query parameter (public access)
+      const specificResponse = await axios.get(
```
dtbuchholz: we should just use the `adminClient.getCompetitionRules` method instead of axios

apps/api/src/routes/competitions.routes.ts
```diff
@@ -662,12 +583,12 @@ export function configureCompetitionsRoutes(
...
    *                           description: Minimum number of trades required per day (null if no requirement)
    *       404:
-   *         description: Competition not found
+   *         description: Competition not found (or no active competition when competitionId is omitted)
    *       500:
    *         description: Server error
    */
   router.get(
-    "/:competitionId/rules",
+    "/rules",
```
dtbuchholz: this still needs to be `/:competitionId/rules`. it should _not_ default to the active comp—e.g., whenever we have parallel comps, we wont have any "default to active" behavior.

also, the frontend uses this endpoint here, so this change would break it: https://github.com/recallnet/js-recall/blob/446c3fb7e05dbaf33270283616a716e0ca5a835e/apps/comps/lib/api-client.ts#L323

apps/comps/lib/api-client.ts
```diff
@@ -317,15 +317,16 @@ export class ApiClient {
...
   /**
    * Get competition rules
-   * @param competitionId - Competition ID
-   * @returns Competition rules for the specified competition
+   * @param competitionId - Competition ID (optional - if not provided, returns rules for active competition)
+   * @returns Competition rules for the specified competition or active competition
    */
   async getCompetitionRules(
-    competitionId: string,
+    competitionId?: string,
```
dtbuchholz: i.e., i think this should be a required path param, not a query param

apps/api/src/routes/competitions.routes.ts
stbrody: api-client function is removed

asutula (approved):

PR #1376, opened 2025-10-06
feat: Use serverless function for getCompetitionRules

stbrody:
Builds on https://github.com/recallnet/js-recall/pull/1374


apps/comps/hooks/useCompetitionRules.ts
```diff
@@ -1,23 +1,21 @@
...
-    retry: 1, // Only retry once if it fails
-  });
+export const useCompetitionRules = (
+  competitionId: string,
+): UseQueryResult<RouterOutputs["competitions"]["getRules"], Error> =>
+  useQuery(
+    tanstackClient.competitions.getRules.queryOptions({
+      input: { competitionId },
+      enabled: !!competitionId,
+      staleTime: 5 * 60 * 1000, // Cache for 5 minutes
```
stbrody, line 18: I'm not quite sure how this caching logic works and how it compares to the other caching logic you've been working on @asutula.

Is this client-side caching that is local to just the individual user's browser?  In which case we'd still want to add server-side NextJs caching

dtbuchholz (approved):
🤘

apps/comps/hooks/useCompetitionRules.ts
asutula: Correct, this is just client side caching. These rules seem like a good candidate for server side caching.

asutula (approved):
Nice and simple.

PR #1377, opened 2025-10-06
chore: Clean up auth controller

stbrody:
Throw errors instead of returning `success: false`, and move key business logic for `login` into `userService`

### DIFF
--- a/packages/services/src/agent.service.ts
+++ b/packages/services/src/agent.service.ts
@@ -1785,35 +1785,19 @@ export class AgentService {
     agentId: string,
     message: string,
     signature: string,
-  ): Promise<{
-    success: boolean;
-    walletAddress?: string;
-    error?: string;
-  }> {
+  ): Promise<string> {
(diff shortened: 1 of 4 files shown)


asutula (approved):

packages/services/src/agent.service.ts
```diff
@@ -1848,17 +1820,17 @@ export class AgentService {
           "[AgentManager] Error recovering wallet address:",
           error,
         );
-        return { success: false, error: "Invalid signature" };
+        throw new ApiError(400, "Invalid signature");
```
dtbuchholz, line 1834: should the actual `ApiError` class include a `success: false` in the body so that the controllers still return that field?

stbrody: tbh I don't really think it's a good thing for controllers to return the "success" field, its unnecessary extra information.  If the endpoint didn't return an error, you know it was successful

PR #1382, opened 2025-10-06
feat: Use serverless function for login

stbrody:
Builds on https://github.com/recallnet/js-recall/pull/1377

### DIFF
--- a/apps/comps/providers/session-provider.tsx
+++ b/apps/comps/providers/session-provider.tsx
@@ -20,11 +20,9 @@ import {
   createContext,
   useCallback,
   useEffect,
-  useRef,
   useState,
 } from "react";
 
-import { ApiClient } from "@/lib/api-client";
 import { mergeWithoutUndefined } from "@/lib/merge-without-undefined";
 import { tanstackClient } from "@/rpc/clients/tanstack-query";
 import { User as BackendUser, UpdateProfileRequest } from "@/types";
@@ -80,7 +78,6 @@ export const SessionContext = createContext<Session | null>(null);
 
 export function SessionProvider({ children }: { children: React.ReactNode }) {
   const queryClient = useQueryClient();
-  const apiClient = useRef(new ApiClient());
 
   const { user, ready, authenticated, logout, isModalOpen, createWallet } =
     usePrivy();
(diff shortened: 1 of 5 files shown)


apps/comps/providers/session-provider.tsx
```diff
@@ -101,7 +117,7 @@ export function SessionProvider({ children }: { children: React.ReactNode }) {
         await createWallet();
       }
       setShouldLinkWallet(isNewUser);
-      loginToBackend();
+      loginToBackend(undefined);
```
asutula, line 173: Weird you need `undefined`, but ok.

apps/comps/rpc/router/user/login.ts
```diff
@@ -0,0 +1,70 @@
...
+        throw errors.UNAUTHORIZED({
+          message: "Invalid authentication token",
+        });
+      }
+
+      // Authenticate and create/update user record
+      const user = await context.userService.loginWithPrivyToken(
+        context.privyIdToken,
+        context.privyClient,
+      );
```
asutula: I have to go back to the`userService.loginWithPrivyToken` code to verify, but it seems like we're likely verifying the user twice now when login is called... Once in the `privyUser` middleware and once in `userService.loginWithPrivyToken`. Seems like it would be sufficient to just call a user service function to update the user with any new data in the privy user by passing the privy user as an arg. Maybe I'm forgetting about some addition work `userService.loginWithPrivyToken` does?

stbrody: Good call, I removed the privyUserMiddleware here and just get the privy token from the cookies directly.  The call to `userService.loginWithPrivyToken` is important because it has a lot of fallback handling of legacy pre-privy users in there.  But this ensures we only validate the privy token once.

Also I decided not to port over the manual privy token verification.  I think the amount of visibility we gain into malformed or otherwise invalid privy tokens is pretty small and unlikely to come up often, vs the cost of doing an extra duplicate set of cryptographic verification on every login.  I could be swayed to bring it back though if there's concern about this

stbrody:
okay @asutula @dtbuchholz, this is ready for another review

asutula (approved):

PR #1387, opened 2025-10-07
fix: Invalidate queries using key derived from tanstackClient

stbrody:
also fixes one frontend call that wasn't using the new RPC yet.

Just cleaning up the stuff that slipped through the cracks during the main refactorings


apps/comps/hooks/useBoost.ts
```diff
@@ -1,80 +0,0 @@
...
-  useMutation,
-  useQuery,
-  useQueryClient,
-} from "@tanstack/react-query";
-
-import { useSession } from "@/hooks/useSession";
-import { apiClient } from "@/lib/api-client";
-import { tanstackClient } from "@/rpc/clients/tanstack-query";
-
-export const useBoosts = (competitionId: string) => {
```
stbrody, line 12: all dead code not called anywhere

PR #1394, opened 2025-10-07
feat: updates activate boost logic

brunocalza:
# Context

Changes the logic of the "Start boosting" button that allocates some boosts to the user, with a feature flag. We now should consider the stakes to determine how much boost to allocate. 

Closes [APP-599](https://linear.app/recall-labs/issue/APP-599/update-activate-boost-logic-for-tge)

This was manually tested by inserting stakes in the database, and manually changing the code to activate one logic or the other.

### DIFF
--- a/apps/comps/components/agents-table/index.tsx
+++ b/apps/comps/components/agents-table/index.tsx
@@ -29,6 +29,7 @@ import {
 import { toast } from "@recallnet/ui2/components/toast";
 
 import { Pagination } from "@/components/pagination/index";
+import { config } from "@/config/public";
 import { useSession } from "@/hooks/useSession";
 import { useVote } from "@/hooks/useVote";
 import { openForBoosting } from "@/lib/open-for-boosting";
@@ -158,11 +159,26 @@ export const AgentsTable: React.FC<AgentsTableProps> = ({
     }),
   );
 
+  const { mutate: claimStakedBoost } = useMutation(
+    tanstackClient.boost.claimStakedBoost.mutationOptions({
+      onSuccess: () => {
+        toast.success("Successfully claimed competition boost!");
+        queryClient.invalidateQueries({
+          queryKey: tanstackClient.boost.balance.key(),
+        });
+      },
+      onError: (error) => {
+        toast.error(error.message);
+      },
+    }),
+  );
+
   const isOpenForBoosting = useMemo(
     () => openForBoosting(competition),
     [competition],
   );
 
+  // TODO: edit condition to check if user has stakes
   const showClaimBoost = useMemo(() => {
     return (
       boostBalance === 0 &&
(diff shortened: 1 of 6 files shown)


apps/comps/components/agents-table/index.tsx
```diff
@@ -230,7 +246,12 @@ export const AgentsTable: React.FC<AgentsTableProps> = ({
   };
 
   const handleClaimBoost = () => {
-    claimBoost({ competitionId: competition.id });
+    // TODO: maybe use a different feature flag
```
brunocalza, line 255: This is where the feature flag is added ( which one should we use? `noStakeBoostAmount` does not look appropriate). If "TGE" we call `claimStakedBoost`.

apps/comps/rpc/router/boost/claim-staked-boost.ts
```diff
@@ -0,0 +1,37 @@
+import { z } from "zod";
```
brunocalza, line 1: Implementation of `claimStakedBoost`, that calls [`claimStakedBoost` ](https://github.com/recallnet/js-recall/pull/1394/files#diff-8560885d52b54f9fc7a0d2654b70b701fce8015f26e657a0feba779e213068a4R93)from the service.

packages/services/src/boost.ts
```diff
@@ -75,6 +90,80 @@ export class BoostService {
     });
   }
 
+  claimStakedBoost(userId: string, wallet: string, competitionId: string) {
```
brunocalza, line 86: new logic for activating boost. We loop through all stakes of a wallet and call `awardForStake`.

packages/services/src/boost.ts
```diff
@@ -75,6 +90,80 @@ export class BoostService {
...
+            {
+              id: competition.id,
+              votingStartDate: competition.votingStartDate,
+              votingEndDate: competition.votingEndDate,
+            },
+            tx,
+          );
+
+          // We should get "applied" for all
+          if (result.type === "noop") {
```
brunocalza, line 131: `awardForStake` should return `noop` for all or `applied` for all. There's no in between state (at least it shouldn't be) we should handle. So if we get one `noop` (which should be the first stake), we consider that the user has already claimed.

apps/comps/components/agents-table/index.tsx
```diff
@@ -158,11 +159,26 @@ export const AgentsTable: React.FC<AgentsTableProps> = ({
...
+      },
+    }),
+  );
+
   const isOpenForBoosting = useMemo(
     () => openForBoosting(competition),
     [competition],
   );
 
+  // TODO: edit condition to check if user has stakes
```
brunocalza, line 181: note: another condition should be added here. Probably if the user has any stakes, or the balance in the navbar is positive. this can be done in another PR

apps/comps/components/agents-table/index.tsx
asutula: Cool. I am in the middle of some work to incorporate a `tge` feature flag everywhere it matters. Maybe let's just leave this as is and I'll update it in my PR?

apps/comps/components/agents-table/index.tsx
asutula: Ok, I'm also in the middle of updating all this state to take into account if the user has a wallet connected, etc. Will take this into consideration.

asutula (approved):
Nice work, thanks for jumping in here.

PR #1395, opened 2025-10-07
fix: standardize color palette across application

stephhuynh18:
Fixes https://linear.app/recall-labs/issue/APP-452/standardize-green-yellow-red-blue

### DIFF
--- a/apps/comps/components/timeline-chart/constants.ts
+++ b/apps/comps/components/timeline-chart/constants.ts
@@ -6,9 +6,9 @@ import { createContext } from "react";
 
 export const CHART_COLORS = [
   "#cc66ff", // Purple
-  "#14f195", // Green
-  "#ff4d4d", // Red
-  "#ffcc00", // Yellow
+  "#38a430", // Green
+  "#d94545", // Red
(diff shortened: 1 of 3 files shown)


apps/comps/components/timeline-chart/constants.ts
```diff
@@ -6,9 +6,9 @@ import { createContext } from "react";
 
 export const CHART_COLORS = [
```
dtbuchholz, line 7: if it's straightforward, we should swap these hardcoded values for the css vars you set. iirc the only place this gets used is here https://github.com/recallnet/js-recall/blob/main/apps/comps/components/timeline-chart/timeline-chart.tsx#L420

i.e., we use that array to somewhat randomly assign a chart color, while making sure it's consistent to the agent and also ensures colors aren't the same for sequential agents.

dtbuchholz (approved):
one minor comment but otherwise lgtm

note: if you used an LLM to codify the hex -> oklch colors, I noticed they hallucinate. I had to check them all myself via a site like https://oklch.com

PR #1397, opened 2025-10-07
feat: Limit users to one agent per competition

stbrody:


packages/db/src/repositories/competition.ts
```diff
@@ -450,6 +450,37 @@ export class CompetitionRepository {
           throw new Error(`Competition ${competitionId} not found`);
         }
 
+        // Check if the agent's owner already has ANY agent (any status) in this competition
+        // First, get the owner of the agent being added
```
stbrody, line 455: So I think ideally this would be a separate function that takes a txn object so the service can orchestrate the restriction logic but within the same transaction.  But this function already has business logic in it (to enforce the cap on number of agents per competition), so this seemed the quickest solution without trying to fully refactor all of this right now.

packages/db/src/repositories/competition.ts
```diff
@@ -450,6 +451,39 @@ export class CompetitionRepository {
           throw new Error(`Competition ${competitionId} not found`);
         }
 
+        // Check if the agent's owner already has ANY OTHER agent (any status) in this competition
```
brunocalza, line 454: This method is [also called in the admin call](https://github.com/recallnet/js-recall/blob/161899c0c433114b4bec9bfec242150d6568ba6b/apps/api/src/controllers/admin.controller.ts#L1398), I am not sure if we also want that restriction when admin is adding an agent. better confirm with @KGmajor

stbrody: Asked on slack, confirmed its fine for admins to be subject to the same restriction.

Good thinking to check though Bruno

brunocalza:
wait, do we need a feature flag for this?

KGmajor:
I can do a quick run through our docs to make sure we don't say you can register many agents at a time (we did have that at some point) - ok not to have that on the frontend. We'll just make it a norm moving forward.

PR #1398, opened 2025-10-07
Stephanie/app 515 fix failed to fetch error

stephhuynh18:
## Summary
* Add retry logic to login to backend. The fetch succeeds BUT we get the failed to fetch error. So added retry logic until we know more about the error

Screenshot 2025-10-07 at 11 55 35 AM

* Added logging for more information to see if its related to the error that sentry also shows before the failed to fetch
`Failed to add embedded wallet connector: Wallet proxy not initialized`

### DIFF
--- a/apps/comps/providers/session-provider.tsx
+++ b/apps/comps/providers/session-provider.tsx
@@ -126,6 +126,8 @@ export function SessionProvider({ children }: { children: React.ReactNode }) {
     mutationFn: async () => {
       await apiClient.current.login();
     },
+    retry: 3,
+    retryDelay: (attemptIndex) => Math.min(1000 * 2 ** attemptIndex, 10000), // Exponential backoff: 1s, 2s, 4s, max 10s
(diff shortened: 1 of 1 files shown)


apps/comps/providers/session-provider.tsx
```diff
@@ -84,6 +85,7 @@ export function SessionProvider({ children }: { children: React.ReactNode }) {
 
   const { user, ready, authenticated, logout, isModalOpen, createWallet } =
     usePrivy();
+  const { ready: walletsReady } = useWallets();
```
dtbuchholz, line 88: [nit] i'd prefer to _not_ add another hook into the privy equation, purely for logging purposes.

stephhuynh18: fair, its a long shot anyway because we don't interact with the wallets during this flow (except create wallet but these errors are not hitting that path). Will remove

PR #1399, opened 2025-10-07
feat: Remove unused endpoints: getCompetitionLeaderboard, getCompetitionStatus, getUpcomingCompetitions, and getPerpsCompetitionSummary

stbrody:
## Removed 4 Unused Competition API Endpoints

- `GET /api/competitions/status`: Removed endpoint and service method. E2E tests updated to use new `getActiveCompetition()` helper that calls `GET /api/competitions?status=active` and returns the first result.

- `GET /api/competitions/leaderboard`: Removed endpoint and `getLeaderboardWithAuthorization()` service method, but **kept** `CompetitionService.getLeaderboard()` as it's used by internal scripts and admin endpoints. E2E tests updated to use `GET /api/competitions/:id/agents` (note: response structure changed from `.leaderboard` array to `.agents` array; filter by `.active` field for active/inactive agents).

- `GET /api/competitions/upcoming`: Removed endpoint and service methods. Updated `award_boost.ts` script to use `CompetitionService.getCompetitions("pending", ...)`. E2E tests updated to use `GET /api/competitions?status=pending`.

- `GET /api/competitions/:id/perps/summary`: Fully removed endpoint, controller method, and all references. E2E tests updated to use `GET /api/competitions/:id` which includes stats in the response.

**NOTE TO REVIEWERS:**
The most controversial changes here are probably that:
1) Added an `includeInactive` arg to the existing getCompetitionAgents endpoint to show inactive agents.  This is exposed in the old API endpoint, though not used in the frontend anywhere.  Mostly this is used by the updated e2e tests since some of them relied on being able to see inactive agents in the leaderboard.  In the future, we may choose to expose this in the new corresponding serverless function as well if we ever want to use it in the frontend.
2) Changed meaning of the `isActive` field in getCompetitionAgents from the agent's global active status to its active status in this particular competition instead.  Since this endpoint is specifically for getting the agents in a specific competition, it makes sense to me that you'd care more about the status in that competition than the global status.


apps/api/e2e/tests/admin.test.ts
```diff
@@ -1308,19 +1309,23 @@ describe("Admin API", () => {
...
-      (await adminClient.getCompetitionLeaderboard()) as LeaderboardResponse;
-    expect(updatedLeaderboard.success).toBe(true);
-    const activeAgentIds = updatedLeaderboard.leaderboard.map(
-      (entry) => entry.agentId,
+    const updatedLeaderboard = await adminClient.getCompetitionAgents(
+      competition.id,
+      { sort: "rank", includeInactive: true },
     );
+    expect(updatedLeaderboard.success).toBe(true);
+    if (!updatedLeaderboard.success) throw new Error("Failed to get agents");
```
dtbuchholz: remove the thrown error and use `expect`

apps/api/e2e/tests/admin.test.ts
```diff
@@ -1359,11 +1364,17 @@ describe("Admin API", () => {
...
-    const leaderboardAfterRemoval =
-      (await adminClient.getCompetitionLeaderboard()) as LeaderboardResponse;
-    const activeAgentIds = leaderboardAfterRemoval.leaderboard.map(
-      (entry) => entry.agentId,
+    const leaderboardAfterRemoval = await adminClient.getCompetitionAgents(
+      competition.id,
+      { sort: "rank" },
+    );
+    if (!leaderboardAfterRemoval.success)
+      throw new Error("Failed to get agents");
```
dtbuchholz: same here

apps/api/e2e/tests/admin.test.ts
```diff
@@ -1379,16 +1390,25 @@ describe("Admin API", () => {
...
-    const leaderboardAfterReactivation =
-      (await adminClient.getCompetitionLeaderboard()) as LeaderboardResponse;
-    const reactivatedAgentIds = leaderboardAfterReactivation.leaderboard.map(
-      (entry) => entry.agentId,
+    const leaderboardAfterReactivation = await adminClient.getCompetitionAgents(
+      competition.id,
+      { sort: "rank" },
+    );
+    if (!leaderboardAfterReactivation.success)
+      throw new Error("Failed to get agents");
```
dtbuchholz: and here. some more instances below (i'll stop commenting)

packages/services/src/competition.service.ts
```diff
@@ -648,7 +609,7 @@ export class CompetitionService {
   ): Promise<string[]> {
     const competitionAgents = await this.agentService.getAgentsForCompetition(
       competitionId,
-      { sort: "", limit: 1000, offset: 0 },
+      { sort: "", limit: 1000, offset: 0, includeInactive: false },
```
dtbuchholz, line 612: created a follow up for the hardcoded "limit" because iirc there are a few places we do think. not a big deal rn.

https://linear.app/recall-labs/issue/APP-608/handle-hardcoded-agent-limit-in-services-params

packages/services/src/competition.service.ts
```diff
@@ -1222,7 +1186,9 @@ export class CompetitionService {
         imageUrl: agent.imageUrl,
         score,
         active: isActive,
-        deactivationReason: !isActive ? agent.deactivationReason : null,
+        deactivationReason: !isActive
+          ? (competitionDeactivationReason ?? null)
```
dtbuchholz: the `?? null` in `(competitionDeactivationReason ?? null)` seems redundant since it falls back to `null`?

dtbuchholz (approved):
@stbrody everything (mostly) lgtm. a few nits about tests that i think are worth fixing

apps/api/e2e/tests/admin.test.ts
stbrody: yeah, it actually already has the `expect`, but without this there's a build error cause the compiler thinks the type could be an error.  Anyway I changed it and added type casting instead, debatable which is cleaner though

PR #1406, opened 2025-10-08
feat(contracts): add timeout, polling interval to wait logic

dtbuchholz:
To [fix a unit test failure](https://github.com/recallnet/js-recall/actions/runs/18344656834/job/52248198376), this PR adds a polling interval, applies the timeout option to the wait for tx logic, and uses these in test (along w/ increased test timeout).

### DIFF
--- a/packages/staking-contracts/src/rewards-allocator.ts
+++ b/packages/staking-contracts/src/rewards-allocator.ts
@@ -37,12 +37,14 @@ interface ClaimResult {
 interface Options {
   timeout?: number;
   retryCount?: number;
+  pollingInterval?: number;
 }
 
 class RewardsAllocator {
   private walletClient: WalletClient;
   private publicClient: PublicClient;
   private contractAddress: Hex;
+  private timeout?: number;
 
   constructor(
     privateKey: Hex,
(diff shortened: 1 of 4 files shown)


packages/staking-contracts/src/rewards-allocator.ts
```diff
@@ -93,6 +98,7 @@ class RewardsAllocator {
 
     const receipt = await this.publicClient.waitForTransactionReceipt({
       hash,
+      timeout: this.timeout,
```
dtbuchholz, line 101: `viem`'s typedocs state the `pollingInterval` for `waitForTransactionReceipt` will default to whatever the `publicClient` or `walletClient` set. the `timeout`, however, needs to be explicitly set in this method.

packages/staking-contracts/test/rewards-allocator.test.ts
```diff
@@ -13,10 +13,10 @@ describe("Allocator Error Path", () => {
     // Mock a failed transaction to hit `allocate`
     const allocator = new RewardsAllocator(
       "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
-      "http://invalid-url", // Invalid RPC to trigger failure
+      "http://127.0.0.1:9999", // Invalid RPC to trigger failure
```
dtbuchholz, line 16: LLMs seems to think this'll ensure there isnt some weird DNS attempt, and it'll fail faster

packages/staking-contracts/test/rewards-claimer.test.ts
```diff
@@ -5,28 +5,35 @@ import { Network } from "../src/rewards-allocator.js";
...
-      "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
-      "http://invalid-url", // Invalid RPC to trigger failure
-      "0x0000000000000000000000000000000000000000",
-      Network.Hardhat,
-      { timeout: 100, retryCount: 0 },
-    );
+  it(
+    "should throw error on network failure",
+    {
+      timeout: 10_000,
```
dtbuchholz, line 11: and this is the test that's currently failing. it's the same logic, except we add the polling interval, change the URL, and increase the test timeout

PR #1407, opened 2025-10-08
Clean up blockchain config

asutula:


ramirotw (approved):
looks good.

as we discussed, two improvements to take care of next week are 1) have both base and base sepolia support, while making base the default for prod and 2) move contract addresses into code for transparency

PR #1719, opened 2026-03-19
Add cli flags to the claims indexer backfill script

joewagner:
This allows us to run the conviction claims backfill script with block ranges via a flag instead of being hard coded.


KGmajor (approved):
👍🏽

